PR #1, opened 2015-11-26
DCP RFC

bensigelman:
See the README.md... I'll add a few comments along the way, too.


opentracing/README.md
```diff
@@ -0,0 +1,109 @@
+# DCP RFC
```
bensigelman: be sure to read this README (preferably via the "View" button that honors the formatting).

opentracing/context.go
```diff
@@ -0,0 +1,39 @@
+package opentracing
```
bensigelman: ("DCP" is too acronymy... I don't have strong feelings about this particular naming possibility.)

opentracing/context.go
```diff
@@ -0,0 +1,35 @@
...
+
+// A ContextID is the smallest amount of state needed to describe a span's
+// context within a larger [potentially distributed] trace. The ContextID is
+// not intended to encode the span's operation name, timing, or log data, but
+// merely any unique identifiers (etc) needed to contextualize it within a
+// larger trace tree.
+//
+// ContextIDs are sufficient to propagate the, well, *context* of a particular
+// trace from process to process.
+type ContextID interface {
```
bensigelman: Note that this ContextID does *not* specify that the implementations must use a trace_id and span_id, etc, etc... that is a choice the tracing impl can make on its own.

opentracing/dapperish/dapper.go
```diff
@@ -0,0 +1,90 @@
+package main
```
bensigelman: this whole `dapperish` package is just an example implementation... obviously the code is not very useful, etc.

opentracing/dapperish/main.go
```diff
@@ -0,0 +1,68 @@
...
+
+func server() {
+	http.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
+		reqCtxID, err := opentracing.GetContextIDFromHttpHeader(
+			req.Header, opentracing.GlobalTracer())
+		if err != nil {
+			panic(err)
+		}
+
+		serverSpan, _ := opentracing.StartSpan("serverSpan", reqCtxID)
```
bensigelman: note that I'm implicitly proposing that the client and server side of an RPC get their own spans.

opentracing/global.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+var (
+	globalOpenTracer OpenTracer = noopOpenTracer{noopContextIDSource{}}
+)
+
+// Should be called as early as possible in main(), prior to calling the
+// `StartSpan*` (etc) global funcs below. Prior to calling `InitGlobal`, any
+// Spans started via the `StartSpan*` globals are noops.
+func InitGlobalTracer(rec Recorder, ctxIDSource ContextIDSource) {
```
bensigelman: everything in this file can be avoided by people who think singletons are horrible :)

opentracing/raw.go
```diff
@@ -0,0 +1,67 @@
+package opentracing
+
+import "time"
+
+type Tags map[string]interface{}
+
+type RawSpan struct {
```
bensigelman: this file is roughly the "raw" interface in the DCP proposal.

opentracing/raw.go
```diff
@@ -0,0 +1,67 @@
...
+	//
+	//   {transactions: [{amount: 10, currency: "USD"}, {amount: 11, currency: "USD"}]}
+	Message string
+
+	// `Payload` can be a POD type, a string, or nested maps and slices; i.e.,
+	// it is a base type or an anonymous struct.
+	Payload interface{}
+}
+
+type Recorder interface {
```
bensigelman: implementations must do something here.

opentracing/span.go
```diff
@@ -0,0 +1,91 @@
+package opentracing
+
+import "golang.org/x/net/context"
+
+type Span interface {
```
bensigelman: this is roughly the "construction" layer from the "DCP Layers" doc

opentracing/span.go
```diff
@@ -0,0 +1,91 @@
...
+	// undefined behavior.
+	Finish()
+
+	// Suitable for serializing over the wire, etc.
+	ContextID() ContextID
+}
+
+// A simple, thin interface for Span creation. Though other implementations are
+// possible and plausible, most users will be fine with `NewStandardTracer()`.
+type OpenTracer interface {
```
bensigelman: Implementations are also expected to do something here... again, this is because different tracing systems represent the `ContextID` in different ways.

In going through this exercise (of creating an actual API in an actual language), I realized that there are really two almost totally separate concerns: the actual span (and log) data, and the "graph structure" of the trace itself. Hence the ContextID stuff to represent the latter.

opentracing/span.go
```diff
@@ -0,0 +1,91 @@
...
+	// `context.Context` descends from the `context.Background()` and the
+	// returned `Span` descends from the provided `opentracing.ContextID`.
+	//
+	// If `parent` is omitted, the returned `Span` is a "root" span: i.e., it
+	// has no known parent.
+	StartSpan(operationName string, parent ...interface{}) (Span, context.Context)
+}
+
+////////////////////////////////////
+// begin context.Context boilerplate
```
bensigelman: FWIW, I have mixed feelings about `context.Context`, but I think that ship has sailed. Might as well get on board!

opentracing/span_standard_impl.go
```diff
@@ -0,0 +1,176 @@
...
+	"time"
+
+	"golang.org/x/net/context"
+)
+
+// Creates and returns a standard OpenTracer which defers to `rec` and
+// `ctxIdSource` as appropriate.
+//
+// See `SetGlobalTracer()`.
+func NewStandardTracer(rec Recorder, ctxIdSource ContextIDSource) OpenTracer {
```
bensigelman: The stuff in this file would change as we make this package more robust... It's included here to illustrate some of the issues that come up, but is not as important as the `interface` definitions elsewhere in this package.

opentracing/utils.go
```diff
@@ -0,0 +1,27 @@
...
+	"encoding/base64"
+	"fmt"
+	"net/http"
+)
+
+const (
+	OpenTracingContextIDHeader = "OpenTracing-Context-Id"
+)
+
+func AddContextIDToHttpHeader(ctxID ContextID, h http.Header) {
```
bensigelman: I'm not sure how many such util functions we would want to bake into the library, but I wanted to write up a few as examples.

opentracing/README.md
```diff
@@ -0,0 +1,121 @@
...
+            // request.
+            opentracing.AddContextIDToHttpHeader(span.ContextID(), httpReq.Header)
+
+            resp, err := httpClient.Do(httpReq)
+            ...
+        }
+        ...
+    }
+
+#### Deerializing from the wire
```
yurishkuro: typo

opentracing/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+	http.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
+		// Grab the ContextID from the HTTP header using the opentracing
+        // helper.
+		reqCtxID, err := opentracing.GetContextIDFromHttpHeader(
+			req.Header, opentracing.GlobalTracer())
+		if err != nil {
+			panic(err)  // obviously don't do this in real code!
```
yurishkuro: usually you want to start a root span in this case

opentracing/README.md
```diff
@@ -0,0 +1,121 @@
...
+#### Serializing to the wire
+
+    func makeSomeRequest(ctx context.Context) ... {
+        if span := SpanFromGoContext(ctx); span != nil {
+            httpClient := &http.Client{}
+            httpReq, _ := http.NewRequest("GET", "http://myservice/", nil)
+
+			// Transmit the span's ContextID as an HTTP header on our outbound
+            // request.
+            opentracing.AddContextIDToHttpHeader(span.ContextID(), httpReq.Header)
```
yurishkuro: I think we're missing an opportunity here for a more generic context propagation, a metadata context that contains arbitrary key/value attributes, of which tracing context ID is just one example. Security credentials can be another example. I just started working on such generic propagation API internally, so don't have any code to share, or even clear idea how it would work. The basic idea though is that the application can opt-in for different types of metadata propagation, e.g. tracing and security credentials, each of which register a handler with DCP that's invoked at the time when incoming or outgoing requests are made, and the handlers (de)serialize the metadata from the headers into metadata context. The point where I am undecided is whether the handlers should be invoked above or inside the spans manipulation, i.e. if it's above then the span manipulation is actually an implementation of the tracing handler. Opinions are welcome.

opentracing/context.go
```diff
@@ -0,0 +1,35 @@
...
+	// Create a child context for this ContextID, and return both that child's
+	// own ContextID as well as any Tags that should be added to the child's
+	// Span.
+	//
+	// The returned ContextID type must be the same as the type of the
+	// ContextID implementation itself.
+	NewChild() (childCtx ContextID, initialChildSpanTags Tags)
+
+	// Serializes the ContextID as an arbitrary byte string.
+	Serialize() []byte
```
yurishkuro: I would suggest string, not byte, serialization by default, with other optional interfaces for other serialization formats, like bytes, or Zipkin-style 4-numbers tuple. Otherwise, the bytes need to be serialized again into a text protocol, e.g. HTTP headers, meaning the common implementation would need to do something like base64, and that makes the context ID on the wire unnecessarily opaque.

opentracing/README.md
```diff
@@ -0,0 +1,121 @@
...
+        defer sp.Finish()
+		sp.Info("called xyz")
+        ...
+    }
+
+#### Creating a root Span (i.e., without a known parent)
+
+    func xyz() {
+        ...
+        sp, ctx := opentracing.StartSpan("span_name")
```
yurishkuro: something is off here, because line 41 says StartSpan method signature is (spanName, context), and Go doesn't allow function overloading.

opentracing/dapperish/main.go
yurishkuro: I'd like that, but how does it work with clock skew adjustment? Do we need need to capture additional information in that case?

Someone asked me a related question the other day. We're using Hyperbahn routing mesh as discovery mechanism, so an RPC call has two hops through the routers ring before reaching the server. The question was how these two hops can be represented as their own spans. In the current Zipkin model they cannot be, since client/server are a single span. If we split those, then I think the client span could have three siblings, {hop1, hop2, server}. But I am not clear how to express that in this new API, or how to do clock skew adjustment without knowing exact causality of these spans, which we're not capturing.

opentracing/noop.go
```diff
@@ -0,0 +1,62 @@
...
+func (n noopSpan) StartChildSpan(operationName string, parent ...context.Context) (Span, context.Context) {
+	if len(parent) > 0 {
+		return noopSpan{}, parent[0]
+	} else {
+		return noopSpan{}, context.Background()
+	}
+}
+func (n noopSpan) SetTag(key string, value interface{})           {}
+func (n noopSpan) Info(message string, payload ...interface{})    {}
+func (n noopSpan) Warning(message string, payload ...interface{}) {}
```
yurishkuro: maybe we don't need warning: http://dave.cheney.net/2015/11/05/lets-talk-about-logging?utm_source=golangweekly

opentracing/noop.go
```diff
@@ -0,0 +1,62 @@
+// Noop implementations of the core opentracing interfaces.
+package opentracing
+
+import "golang.org/x/net/context"
+
+// noopContextID:
+type noopContextID struct{}
```
yurishkuro: for efficiency, it's best to instantiate these once and always return the same instances. I did a simple perf test in Python implementation the other day, the perf difference was about 20 times between a noop tracer and a regular tracer that had sampling rate of 0. 

https://github.com/uber-common/opentracing-go/blob/master/noop.go#L28

opentracing/raw.go
```diff
@@ -0,0 +1,67 @@
...
+	SeverityInfo Severity = iota
+	SeverityWarning
+	SeverityError
+)
+
+type RawLog struct {
+	Timestamp time.Time
+
+	// Info/Warning/Error.
+	Severity
```
yurishkuro: not sure this needs to exist

opentracing/span.go
```diff
@@ -0,0 +1,91 @@
...
+	// `context.Context` and `Span` are schematic children of that context and
+	// any `Span` found therein.
+	//
+	// If `parent` is an `opentracing.ContextID`, the returned
+	// `context.Context` descends from the `context.Background()` and the
+	// returned `Span` descends from the provided `opentracing.ContextID`.
+	//
+	// If `parent` is omitted, the returned `Span` is a "root" span: i.e., it
+	// has no known parent.
+	StartSpan(operationName string, parent ...interface{}) (Span, context.Context)
```
yurishkuro: this is not an idiomatic way to define API in Go. Why not simply define 3 different methods?

opentracing/span.go
```diff
@@ -0,0 +1,91 @@
...
+	// Like Info(), but for warnings.
+	Warning(message string, payload ...interface{})
+
+	// Like Info(), but for errors.
+	Error(message string, payload ...interface{})
+
+	// Sets the end timestamp and calls the `Recorder`s RecordSpan() internally.
+	//
+	// Calling Finish() multiple times on the same Span instance may lead to
+	// undefined behavior.
```
yurishkuro: I would make an even stronger statement - Finish() should be the last call on the span, calling any of the above methods after Finish() has been called has undefined behavior and may be ignored by the tracing system.

opentracing/README.md
```diff
@@ -0,0 +1,121 @@
...
+##### Note: the singletons are optional
+
+If global singletons make you sad, use `opentracing.NewStandardTracer(...)`
+directly and manage ownership of the `opentracing.OpenTracer` explicitly.
+
+#### Creating a Span given an existing Golang `context.Context`
+
+    func xyz(ctx context.Context, ...) {
+        ...
+        sp, ctx := opentracing.StartSpan("span_name", ctx)
```
yurishkuro: I liked Adrian's suggestion of using tags to describe spans instead of a plain string:
>```
type SpanTags map[string]string
st := opentracing.SpanTags{"name":"foo", "version":"1.1", "subsystem":"bar"}
sp, ctx := opentracing.StartSpan(st, ctx)
```

One thing I am not clear on is where we capture the _service name_ of the current service, which is very fundamental concept in Zipkin (yet modeled in a very weird way, as the Endpoint in the annotations - logs in DCP terms).

Comment (bensigelman) [inline opentracing/README.md]: Great question about the service name -- it's quasi-supported via the `SetTags()` at the `Recorder` (rather than `Span`) level, but I think it should be promoted from a mere optional tag in that interface. I also think it appropriate to rename `Recorder` since it's not quite the right thing, semantically... maybe more like `ProcessRecorder`?

Nit/thought: I hesitate to call it "service name" since there are also non-service components of distributed systems (mobile clients, for instance). Maybe "component name"???

File: opentracing/README.md (hunk 18 above)
Comment (bensigelman) [inline opentracing/README.md]: I think you saw the trick here (and I also think you didn't like it :)... basically a final `...interface{}` param. I'll comment more down below.

File: opentracing/README.md (hunk 16 above)
Comment (bensigelman) [inline opentracing/README.md]: I have a bunch of thoughts about this and am not sure if github comments are the most practical place to express them!

In any case, I think we could support something that without great effort or inconvenience given how opaque the ContextID stuff is in this proposal. The main challenge would be deciding on a way to get arbitrary structured information "inside" the ContextID implementation without either (a) making the APIs too complex for the simple cases, or (b) forcing the user to do a bunch of explicit casts/etc.

Maybe the best thing would be to actually make a separate RFC branch and comment on the diff as another faux-pull-request. I'll do that when I have time (may not be today).

File: opentracing/README.md (hunk 14 above)
Comment (bensigelman) [inline opentracing/README.md]: (thanks, fixed)

File: opentracing/README.md (hunk 15 above)
Comment (bensigelman) [inline opentracing/README.md]: true, changed.

File: opentracing/context.go (hunk 17 above)
Comment (bensigelman) [inline opentracing/context.go]: Ok, noted... I don't feel strongly about the default but I do think `[]bytes` should be an option at least. Happy to change this whenever once we have consensus about the other (IMO larger) issues.

I would be concerned with adding zipkin-style 4-number tuples here, by the way(*). Some ContextIDs won't be able to represent themselves that way, so it's problematic to add it at this layer. Thoughts?

(*): it also makes me grumpy that Zipkin sends the parent_id over the wire... should not be formally necessary to serialize along with the context, but so it goes! Minor point.

File: opentracing/dapperish/main.go (hunk 5 above)
Comment (bensigelman) [inline opentracing/dapperish/main.go]: The `Recorder` layer is the best place to manage clock skew... and it seems utterly reasonable to use an annotation on the client and/or server side to clarify that there's an RPC between a parent and child span (at which point you can do clock skew correction in a semantically equivalent way to how Zipkin does it, even with two spans rather than one). Does that make sense?

As for the multi-hop case: this sounds like just the sort of thing that one could use span tags for... make sure that there's a clear way to identify middle-hop spans (via a well-specified tag) and infer "parentage" (causality) via something other than an explicit parent_id if those are unavailable. Happy to write more if that's not clear.

File: opentracing/noop.go (hunk 20 above)
Comment (bensigelman) [inline opentracing/noop.go]: makes sense!

File: opentracing/noop.go (hunk 19 above)
Comment (bensigelman) [inline opentracing/noop.go]: I officially do not feel strongly about this and am happy to do whatever others think is most reasonable (as long as there's a way to distinguish errors from non-errors). I can see arguments either way for various levels of fidelity around log severity.

Comment (bensigelman) [inline opentracing/noop.go]: BTW: the convention from google-internal systems was nice in my view: `Info` for anything interesting, `Warning` for things that seem broken but might have to do with external processes, `Error` for things that seem broken *internally* (in the current process, regardless of other processes), and `Fatal` if there's a risk of corrupting state.

File: opentracing/raw.go (hunk 21 above)
Comment (bensigelman) [inline opentracing/raw.go]: fair point, see above.

File: opentracing/span.go (hunk 23 above)
Comment (bensigelman) [inline opentracing/span.go]: ok, changed!

File: opentracing/span.go (hunk 22 above)
Comment (bensigelman) [inline opentracing/span.go]: Ha... I actually started with three different methods but collapsed them after getting grumpy about `StartRootSpan` (or whatever it would be called)... I still haven't totally made peace with `context.Context` and could imagine a `StartSpan(opName)` that does something sketchy with the stack to figure out a possible parent, even in the absence of `context.Context`. We did this internally at my company (as an experiment) and it didn't generate "false positive" parents in the systems we tested against.

I also liked the idea of supporting multiple parents using this sort of varargs idiom.

If we decided to add the `Tags` to the `StartSpan` signature, though, there'd be a great argument to move back to the separate signatures, etc, etc.

By the way, there is certainly plenty of "prior art" for golang APIs that make use of reflection-based tricks like this. Though I wouldn't argue that Rob Pike probably disapproves of them :)

**Comment by bensigelman** [comment]
@yurishkuro awesome comments, thank you! I did a few tiny quick/easy things, but I don't have time right now for the more thoughtful/deep changes, esp around PivotTracing-like ContextIDs, tags, service identification, and so forth.

I will try to get to those before Monday in any case.

Thanks again,
Ben

File: opentracing/context.go (hunk 17 above)
Comment (yurishkuro) [inline opentracing/context.go]: > but I do think []bytes should be an option at least

Agreed, since binary and text are the two prevalent encodings. HTTP instrumentation should choose text, a thrift-based protocol might choose bytes, thus avoiding potentially unnecessary intermediate encoding.

> Some ContextIDs won't be able to represent themselves that way, so it's problematic to add it at this layer.

I was thinking as an optional interface, e.g https://github.com/uber-common/opentracing-go/blob/master/zipkin.go.

> it also makes me grumpy that Zipkin sends the parent_id over the wire

I thought that was to help with cases where some spans of the trace didn't make it to storage. Having parent span ID allows to put spans in some sort of casual relationship in the UI. It's like an error correction code. Won't help much if storage is missing two levels of the tree.

File: opentracing/span.go (hunk 22 above)
Comment (yurishkuro) [inline opentracing/span.go]: I'm coming from a position of having the caller to be explicit about what it is that they think they are doing. While iterating on `uber/open-tracing` we actually changed the top-level span creation from `BeginRootSpan, BeginSpan(id)` to `BeginTrace, JoinTrace` signatures, and we kept `span.BeginChildSpan()` as is. Such API seems to present a clearer mental picture to developers doing instrumentation.

Comment (bensigelman) [inline opentracing/span.go]: I like `BeginTrace`/`JoinTrace`... I will try some things out and see how things fit together.

File: opentracing/dapperish/main.go (hunk 5 above)
Comment (yurishkuro) [inline opentracing/dapperish/main.go]: > use an annotation on the client and/or server side to clarify that there's an RPC

I am ok with using annotations for representing structural concerns at the RAW layer, but given that this is a "construction" layer, I would prefer a strongly typed way of expressing structural relationships, otherwise every instrumenter needs to know the exact name of the annotation, and its exact value type, which is Zipkin's cs/cr mess all over again.

What probably makes the multi-hop case difficult is that it's essentially a data pipeline scenario, which, as we already said earlier, is indeed difficult in the span model. Again, I have the same issue with using custom tags to solve it, because construction layer should be explicit about structural relationships so that all instrumented frameworks produce the data consistently.

One way I think the span model can work is by recording explicit _elder sibling_ relationship. 

```
 client span
+------------------------------------------------+
    +------+   +------+   +-------------------+
     hop 1      hop 2      server
```
Each router records is "hop" span as the time spent in the router. Hop-2 must record not only that it's a child of "client span", but that it also strictly follows "hop-1", and "Server" strictly follows "hop-2". Then the clock skew adjustment can be done by assuming that all 4 network intervals between these spans are equal.  

Unfortunately, at the moment I do not see an easy way to describe that relationship, especially in the "server" span which has no knowledge that it was reached as a sibling of hop-2 and not as a direct child of "client span". Can I take you up on your offer to explain how you thought you could solve this with tags?

File: opentracing/README.md (hunk 24 above)
Comment (yurishkuro) [inline opentracing/README.md]: Re the Recorder tags, that may not be the right place to store "service name" or whatever we decide to call it, because from the efficiency considerations I can imagine a recorder to be instantiated once per process/application (especially given the optional "singleton" approach :wink:  ), but an application can expose multiple services.

What if we make `component` and `operation` to be required arguments of _span creation methods_? 

tl;dr;

In Zipkin the service name is encoded in the `Endpoint{host, port, serviceName}` descriptor. Given that an app can listen to multiple ports, in our uber/open-tracing API we opted to take this descriptor in BeginTrace/JoinTrace, as those are the entry points to the service, and then BeginChildSpan inherits the descriptor. If the child span is a local span, it can explicitly provide its `component name`, in the begin-options.

To step back a bit, though, the reason service name is important in Zipkin is because it's the main filter in the UI, and it's also used to build service dependency diagram. If I had a service with 10 different URL routes, I actually wouldn't mind seeing a dependency diagram per service+spanName. That suggests that perhaps service name isn't so special vs. the operation/spanName. I could even ask for aggregation by any other tag on the span, e.g. component+operation+version.

File: opentracing/dapperish/main.go (hunk 5 above)
Comment (bensigelman) [inline opentracing/dapperish/main.go]: First of all, I agree completely that causality should be described at higher levels of the API, not just in raw annotations.

Anyway, I'm imagining that the client (i.e., the hyperbahn client) could do this... if we model each hop as two spans – one on the client, one in some routing server – obviously each server span is a blocking child of the corresponding client span. The trick is to describe the causal relationships between the various (hop1, hop2, etc) client-side spans. One possibility would be to add a `BlockedOnCompletion` annotation that points to a previously-completed `ContextId`; if that's too much bookkeeping, then the [hyperbahn] client can merely emit special `Log` statements that explicitly specify that future span children depend on previously completed span children (without actually specifying span_ids or similar for those children). I am thinking this would look vaguely like `Span.AddCausalityHint(opentracing.FutureChildrenDependOnPreviouslyCompletedChildren)`, though hopefully with a more concise name!

Not sure if the above is clear or not.

File: opentracing/README.md (hunk 24 above)
Comment (bensigelman) [inline opentracing/README.md]: I don't want to force every span creation to include a "component"... that will just force the caller to pass that string around.

I like the idea of tying things back to the process (or something close to it)... there's plenty to be gained by tracking the particular binary build and interference effects at a per-process level.

This is not to say that we shouldn't be encouraging people to use `Span` tags for filtering and aggregation in various tracing UIs. All I'm really arguing for is the existence of a "process-like concept" that is factored out of all spans created *by* that process-like thing, whether it be a mobile client or a server or whatever.

FYI, when a single process exposes multiple RPC services, my opinion (not scientific -- just a personal bias) is that those multiple RPC services should still all report through a single process-wide recorder; it's just that they should mark their top-level spans with a "service" tag.

File: opentracing/dapperish/main.go (hunk 5 above)
Comment (yurishkuro) [inline opentracing/dapperish/main.go]: > obviously each server span is a blocking child of the corresponding client span

That's actually different from what I was thinking. A naive way to model this would be with fully nested spans
```
* start client span
** start router1 server span
*** start router1 client span (as client of router2)
**** start router2 server span
***** start router2 client span (as client of target service)
****** start actual service server span
```
This is perfectly doable, but looses the dependency between top and bottom services. I consider hops through the routers to be a part of the overall RPC call, thus they are not nested. And unlike  regular service spans that start on receipt of request and end on emitting a response, the router spans start on receipt of request and end on sending another request to the next hop, i.e. they represent the time spent in the router while passing on the message. To reuse my previous diagram:
```
 client span
+----------------------------------------------------------------------------+
    +------+   +------+   +-------------------+ +----------+ +------------+
     hop 1      hop 2      server                back-hop-2   back-hop-1
```

>> Span.AddCausalityHint(opentracing.FutureChildrenDependOnPreviouslyCompletedChildren)

But how would parent know the ordering of the children?  That's where I think X-Trace's way of recording casuality via event IDs comes very handy. There was a paper on combining span and event model: http://cs.brown.edu/~rfonseca/pubs/leavitt.pdf

File: opentracing/README.md (hunk 24 above)
Comment (yurishkuro) [inline opentracing/README.md]: +1 for "process" - that would be captured at construction of the Recorder, right? We need to formalize it in the API then.

Is your concern with capturing the "component" because of data redundancy? On one hand I agree with it, since if the component is recorded at a parent span but not on children spans, then children can be inferred to be long to the same component. On the other hand, it's not clear to me that instrumentation code always knows that relationship, i.e. a component might have "operations" A and B, both public, but A also calls B internally. Instrumentation of B may need to emit the component name in direct calls, but can skip it in calls from A, which sounds rather complicated.

File: opentracing/README.md (hunk 16 above)
Comment (yurishkuro) [inline opentracing/README.md]: how about `ContextID.addMetaTag(k.v)`? And in this case maybe rename ContextID to just Context.

Comment (bensigelman) [inline opentracing/README.md]: I am hoping to send out RFCv1.1 in light of your [excellent] feedback... in it, I renamed `ContextID` as `TraceContext`, and my method name is `SetTraceTag(k, v)`, but same idea regardless... My biggest concern is how to deal with programmers who don't understand the [considerable] cost of adding "Trace Tags" (since they are sent in-band and thus need to be kept really small). Google had a few things like this that were sent along the dapper code path, but it was far from a general-purpose API (in order to avoid accidental abuse).

Comment (bensigelman) [inline opentracing/README.md]: Oh, PS: I went with `ContextID` rather than `Context` because of the potential for confusion with `context.Context`... But `TraceContext` should get the job done, I hope.

File: opentracing/README.md (hunk 24 above)
Comment (bensigelman) [inline opentracing/README.md]: Yeah, the [pending, not-sent-out-yet] `ProcessRecorder` interface has the following method:

```
	// Every process in the system must have a name. It is up to the
	// ProcessRecorder implementation to determine how this name is specified.
	ProcessName() string
```

My concern with capturing the component is just that it may be too much to require of programmers who are in the middle or the bottom of the stack... e.g., if I am writing an ORM thats sits above a database, what component am I in? The ORM component, or some particular web service that called a model layer that then called me? There is the problem of passing the strings around, of course, but also just the need to explain this additional concept...

I'm not sure I can make a terribly strong argument against it, honestly. It just feels like we are already asking programmers to be awfully patient as we explain new terminology and concepts to them, so I want to minimize that unless there's a *clear* need for additional "required concepts"... that said, I am 100% in favor of adding a "component" tag to spans where appropriate and building UIs that make it easy to filter and aggregate across components.

Happy to talk more about this if you want, and let me know if it seems like I'm misunderstanding something.

File: opentracing/README.md (hunk 16 above)
Comment (yurishkuro) [inline opentracing/README.md]: hehe, so you caved in to using the word "tracing" after all. I'm on board. I don't think what we're doing here can be usefully expressed outside of tracing-related API.

Comment (bensigelman) [inline opentracing/README.md]: correct :)

I realized I was [perhaps] being too pedantic about "context propagation"... I still don't love the term "tracing" since it is overloaded, but for better or worse, that's what people seem to call the thing that we're doing here!

File: opentracing/dapperish/main.go (hunk 5 above)
Comment (bensigelman) [inline opentracing/dapperish/main.go]: Oh, I see. Yes, I would have just modeled these [naively, as you say] as fully nested spans. I had misunderstood previously and thought the client was doing some smart multi-hop stuff on its own.

`<soapbox>`
BTW, I have seen that paper and frankly have to roll my eyes, at least a little... *The single biggest problem* with tracing systems is rarely their data model: it is getting busy, overworked programmers to adopt them. The paper concedes that spans are nice because programmers have a pre-existing mental model for them. The paper then goes on to explain how the event model is more expressive, which – given the paper's definition of the span model – is obvious and hardly requires a formal proof!

But it's entirely possible to express multiple-parent DAG structures using the model we are discussing here... i.e., we can define a particular sort of `RawLog` that has a payload value for key `HappensBeforeSpanId`. All of a sudden, the span model can express anything the event model can express, and – voila – they are equivalent in terms of their expressive capabilities. Except that the default mode of operation in the span model is easier for busy programmers to understand at a glance.

As the paper puts it:

```The downside of the X-Trace model is that it is not necessarily as intuitive for programmers to use as the span model. The call-stack view that the span model provides has a level of simplicity that the event model cannot match.```

Right. So let's just use a span-based programming model which can provide for event-oriented features as a special case... thus "making the easy things easy and the hard things possible" (as they say).
`</soapbox>`

File: opentracing/README.md (hunk 24 above)
Comment (yurishkuro) [inline opentracing/README.md]: I was writing a long reply, and in the process convinced myself that you're right, we should not prescribe a particular way of expressing hierarchy of span _significance_. Some spans need various attributes like component, version, operation, sub-module, etc., while others may only need a name. But I would like to make one thing clear in the docs - a span with just a name is implicitly assumed to be a _local_ span. Without that I think we will create massive name clashes that the UI won't be able to resolve. For example, if both service A and service B have a local span called "orm", each with custom implementation of ORM, it doesn't make sense for the service dependency diagram to show that both A and B depend on the same "orm" component, since it's a name clash. On the other hand, if they both use Hibernate, then instrumentation can emit component=hibernate and that _would_ make sense to show in the deps diagram.

> I am 100% in favor of adding a "component" tag to spans

Do you have a suggestion for API to do that?  What I did in uber/opentracing-go was to make begin-like methods take an optional Options object containing various common concepts that we want to encode as annotations, but without exposing developers to the fact of known how to format those annotations: https://github.com/uber-common/opentracing-go/blob/master/options.go#L37 It included component name, async flag, and "peer" address to record which remote end point an RPC client span is calling (the server accordingly records address of the client who called it, both mapped to SA,CA annotations in Zipkin).

Comment (bensigelman) [inline opentracing/README.md]: +1 to your idea that spans at the boundary of a process (e.g., RPC spans) will be a mess if their only attribute is a name.

I could make a suggestion or three about how to reify all of this in an API but worry about doing so without taking the time to instrument a real system to better understand the prevalence of various use cases and scenarios. One possibility would be to optionally accept a key:value tag map at span start/begin time, and to just have a bunch of predefined keys for programmers to reuse as they see fit... that's a middle-ground approach that encourages consistency without adding clutter to the core APIs.

File: opentracing/utils.go [hunk 25]
```diff
@@ -0,0 +1,27 @@
+package opentracing
+
+import (
+	"encoding/base64"
+	"fmt"
+	"net/http"
+)
+
+const (
+	OpenTracingContextIDHeader = "OpenTracing-Context-Id"
```
yurishkuro: Q: is there any value in the standard header name? We're having an internal debate about it. IMO, if two services are running with two different DCP implementations, the server service is not going to be able to make sense of the header sent by the client service, therefore having a universal header name is counterproductive. So maybe the header name should be an implementation detail of the Recorder?

One case where the common header might be useful is if the two tracing systems have the exact same representation and meaning of the context ID, but that seems rather contrived.

bensigelman: That's a good question. The API as its presently written does not actually have the `Recorder` doing the work of storing/fetching to/from HTTP headers (which is just one of many possible examples). So even though I agree the value for the key is not going to mean anything if the two communicating processes have different recorder types, I do think there's some marginal value in having a utils file (or even package) that does the grunt work of getting things on and off the wire for various protocols.

Or that was my thought. I.e., note how this file is part of the main `opentracing` package rather than the `dapperish` implementation example.

yurishkuro: Yes, totally agree in having util methods for common behavior. Just questioning whether the actual header name needs to be impl specific. I think one of your goals was to keep the Raw API minimal, and have the shared Construction API to do the common tasks. So we want to decide whether serialization of the TraceContext into the on-the-wire protocol is a common task or an implementation detail of the Raw API. If I am not mistaken, current `ContextID` belongs to the Raw API, and it has `toBytes`. If we had the following API in _some_ component of Raw layer, would be be sufficient for serialization?
```Go
type TraceContextPickler interface {
    // takes care of serialization to text headers
    func asStringMap(context TraceContext) map[string]string 

    func fromStringMap(headers map[string]string) (TraceContext, error)

    // takes care of byte stream serialization
    func asBinary(context TraceContext) []byte  // or perhaps accept target byte buffer

    func fromBinary(bytes []byte) (TraceContext, error)
}
```

Then the common util method can be `toRequestHeaders(httpReq, context, pickler)`

bensigelman:
RFC part two is in a separate PR: https://github.com/opentracing/api-golang/pull/2/files

opentracing/utils.go
bensigelman: I don't really have super-strong feelings about this... definitely see how what you have above makes sense, and also see how what's in RFC 1 and RFC 2 makes sense (as far as wire serialization is concerned). Do you think one is clearly better than the other? If so, can you help me get "off the fence"?

bensigelman:
(Since PR #2 was actually a diff against the `changes/rfc` branch, that merge has affected *this* PR (#1), and I will now merge *it* into `master`)

PR #2, opened 2015-12-01
DCP RFC #2

bensigelman:
I'll leave comments inline.

I used RFC #1 as the diff base... feel free to compare against `master`, obviously!


opentracing/README.md
```diff
@@ -8,13 +8,8 @@ For background:
 
 Everyday consumers of this `opentracing` package really only need to worry
 about a couple of key abstractions: the `StartSpan` function, the `Span`
-interface, and golang's `context.Context` idiom. Here are code snippets
-demonstrating some important use cases.
-
-A note about `context.Context`: the opentracing API here encourages (but does
```
bensigelman: I did less coupling to `context.Context` in this version. I think I prefer the results... it's still awkward to propagate both a golang `context.Context` *and* our own `Span`s/`TraceContext`s, but IMO this makes the `opentracing` API easier to use and understand.

opentracing/README.md
```diff
@@ -34,21 +29,21 @@ The simplest starting point is `./global.go`. As early as possible, call
...
     }
 
-#### Creating a Span given an existing Span
+#### Creating a Span given an existing Golang `context.Context`
 
-    func xyz(parentSpan opentracing.Span, ...) {
+    func xyz(goCtx context.Context, ...) {
         ...
-        sp, ctx := opentracing.StartSpan("span_name", parentSpan)
+        sp, goCtx := opentracing.JoinTrace("span_name", goCtx).AddToGoContext(goCtx)
```
bensigelman: the `AddToGoContext` is my nod to `context.Context` ease-of-use.

opentracing/README.md
```diff
@@ -56,9 +51,11 @@ directly and manage ownership of the `opentracing.OpenTracer` explicitly.
 
 #### Creating a root Span (i.e., without a known parent)
 
+Just to show that it's not required, we don't call `AddToGoContext` this time.
+
     func xyz() {
         ...
-        sp, ctx := opentracing.StartSpan("span_name")
+        sp := opentracing.StartTrace("span_name")
```
bensigelman: riffing off of the Uber terminology here (talking about the trace rather than the span).

opentracing/dapperish/trivial_recorder.go
```diff
@@ -9,25 +8,29 @@ import (
 )
 
 type TrivialRecorder struct {
-	tags map[string]string
+	processName string
+	tags        map[string]string
 }
 
-func NewTrivialRecorder() *TrivialRecorder {
+func NewTrivialRecorder(processName string) *TrivialRecorder {
```
bensigelman: note that the recorder interface is effectively forced into accepting a process name as a param (this is a good thing IMO)

opentracing/go_context.go
```diff
@@ -0,0 +1,27 @@
+package opentracing
+
+import "golang.org/x/net/context"
```
bensigelman: this file has all of the even slightly interesting `context.Context` integration code (and there's really not much of it)

opentracing/open_tracer.go
```diff
@@ -0,0 +1,37 @@
...
+	//     val tracer OpenTracer = ...
+	//
+	//     sp := tracer.StartTrace("GetFeed")
+	//
+	//     sp := tracer.StartTrace(
+	//         "HandleHTTPRequest",
+	//         "user_agent", req.UserAgent,
+	//         "lucky_number", 42)
+	//
+	StartTrace(operationName string, keyValueTags ...interface{}) Span
```
bensigelman: new name and signature here... note the `keyValueTags` idiom (which I've seen in a few places)... basically saves the programmer the trouble of typing something like `Tags{ ... }` and also degrades elegantly for the case where there are no tags to add.

opentracing/process_recorder.go
```diff
@@ -0,0 +1,26 @@
+package opentracing
+
+// A ProcessRecorder handles all of the `RawSpan` data generated via an
+// associated `OpenTracer` (see `NewStandardTracer`) instance. It also names
+// the containing process and provides access to a straightforward tag map.
+type ProcessRecorder interface {
+	// Every process in the larger distributed system must have a name. It is
+	// up to the `ProcessRecorder` implementation to determine how this name is
+	// specified.
+	ProcessName() string
```
bensigelman: see comment above about effectively requiring a process name when constructing an implementation for this interface

opentracing/standard_impl.go
```diff
@@ -0,0 +1,175 @@
...
+	"time"
+
+	"golang.org/x/net/context"
+)
+
+// Creates and returns a standard OpenTracer which defers to `rec` and
+// `ctxIdSource` as appropriate.
+//
+// See `InitGlobalTracer()`.
+func NewStandardTracer(rec ProcessRecorder, ctxIdSource TraceContextIDSource) OpenTracer {
```
bensigelman: this file is "morally equivalent" to the old version... just a straightforward implementation of the interface.

opentracing/trace_context.go
```diff
@@ -0,0 +1,164 @@
...
+
+// A TraceContextID is the smallest amount of state needed to describe a span's
+// identity within a larger [potentially distributed] trace. The TraceContextID
+// is not intended to encode the span's operation name, timing, or log data,
+// but merely any unique identifiers (etc) needed to contextualize it within a
+// larger trace tree.
+//
+// TraceContextIDs are sufficient to propagate the, well, *context* of a particular
+// trace from process to process.
+type TraceContextID interface {
```
bensigelman: one big design decision here was to separate `TraceContextID` from `TraceContext` (see below).

opentracing/trace_context.go
```diff
@@ -0,0 +1,164 @@
...
+// A `TraceContext` builds off of an implementation-provided TraceContextID and
+// adds a simple string map of "trace tags". The trace tags are special in that
+// they are propagated *in-band*, presumably alongside application data and the
+// `TraceContextID` proper. See the documentation for `SetTraceTag()` for more
+// details and some important caveats.
+//
+// Note that the `TraceContext` is managed internally by the opentracer system;
+// opentracer implementations only need to concern themselves with the
+// `TraceContextID` (which does not know about trace tags).
+type TraceContext struct {
```
bensigelman: Per the above review comment, `TraceContext != TraceContextID`. The difference is that the `TraceContext` proper handles this new/powerful/interesting notion of "trace tags". Hopefully the use cases will be evident to people actually reading this review :)  If not, I can provide some motivating examples. I don't question the utility, though I do have some reservations about misuse... we'd probably need to set some pretty strict limits on the size and number of the trace tags.

opentracing/trace_context.go
```diff
@@ -0,0 +1,164 @@
...
+// IMPORTANT NOTE #1: `SetTraceTag()` will only propagate trace tags to
+// *future* children of the `TraceContext` (see `NewChild()`) and/or the `Span`
+// that references it.
+//
+// IMPORTANT NOTE #2: Use this thoughtfully and with care. Every key and value
+// is copied into every local *and remote* child of this `TraceContext`, and
+// that can add up to a lot of network and cpu overhead for large strings.
+//
+// Returns a reference to this TraceContext for chaining, etc.
+func (t *TraceContext) SetTraceTag(key, value string) *TraceContext {
```
bensigelman: (worthwhile comment above)

opentracing/trace_context.go
```diff
@@ -0,0 +1,164 @@
...
+
+	ctxId, childSpanTags := t.Id.NewChild()
+	return &TraceContext{
+		Id:        ctxId,
+		traceTags: newTags,
+	}, childSpanTags
+}
+
+func (t *TraceContext) SerializeASCII() string {
+	// XXX: implement correctly if we like this API
```
bensigelman: NB: I am not actually bothering to serialize/deserialize the tags until people tell me they like the concept! Obviously this is not hard to do if we decide to move forward with it.

opentracing/trace_context.go
bensigelman: (I didn't want to force TraceContextID implementors to do the presumably redundant work of serializing the trace tags... that can be the job of the `opentracing` lib.

bensigelman:
As before, there are some hacked-up godocs available at https://godoc.org/gopkg.in/opentracing/api-golang.v2/opentracing ...

opentracing/trace_context.go
yurishkuro: people do like it! :smile:   But there are different way to serialize, e.g.
* a single opaque string stored in a single HTTP header - simple, but may run afoul of header size limits 
* a single opaque string chopped into pieces of portable size and sent in multiple HTTP headers, X-TraceCtx-1, X-TraceCtx2, ...
* each tag in its own header, e.g. `X-TraceCtx-{tagName}: {tagValue}`

opentracing/process_recorder.go
yurishkuro: +1 for requiring process name, but `ProcessRecorder` sounds rather confusing - why not `TraceRecorder` or just `Recorder`?

bensigelman: Yeah, [apparently] I wanted to indicate the intended scope in the name, and also have the type name account for the `ProcessName` method somehow. I'll mull it over...

opentracing/trace_context.go
bensigelman: I was imagining trace tags totaling less than 1K in all (which I think is fine for all reasonable HTTP servers?). So yes, I was thinking a single opaque string. If that turns out to be infeasible, then I guess a single opaque string chopped into pieces.

opentracing/trace_context.go
yurishkuro: Ben, any particular reason not to have TraceContext as an interface, similar to OpenTracer and Span?

bensigelman: Nothing profound... I was feeling uneasy about the number of abstractions/interfaces and thought this one was small enough to build into the package directly.

opentracing/trace_context.go
yurishkuro: I just finished refactoring my Python implementation, that now includes `TraceContext` concept. We could actually take it as a starting point for opentracing/api-python, since structurally it's very close to this Go version, and has lots of unit tests. 

However, for context serialization I went with one tag per header for now, because it makes http requests more readable, especially for people who want to take a peek at the metadata they are propagating. With opaque strings it would be hard. But I have it implemented as an interface, so it can be replaced with opaque strings as well.  I am still polling people in the company who actually work with http services and are interested in propagating stuff via trace context as to their preference for serialization format.

opentracing/trace_context.go
yurishkuro: ok. I don't have a strong preference as well, was just concerned that once we publish the API with a concrete class, it will be hard to change to an interface later. Actually, in my Python version I don't have a 'context ID' abstraction, only Trace Context, since ID and tags are always serialized together, and ID has practically no meaning to anyone by itself, I didn't want to instantiate an extra object.

opentracing/trace_context.go
resospoons: FYI: https://tools.ietf.org/html/rfc6648

tl;dr:
           Deprecating the "X-" Prefix and Similar Constructs
                        in Application Protocols

Abstract

   Historically, designers and implementers of application protocols
   have often distinguished between standardized and unstandardized
   parameters by prefixing the names of unstandardized parameters with
   the string "X-" or similar constructs.  In practice, that convention
   causes more problems than it solves.  Therefore, this document
   deprecates the convention for newly defined parameters with textual
   (as opposed to numerical) names in application protocols.

opentracing/trace_context.go
bensigelman: I think I'll leave this as-is for now... it's a 6-vs-half-dozen situation in my eyes, and so the tiebreaking vote goes to whatever's simpler.

opentracing/trace_context.go
bensigelman: Yuri, I'd be interested to learn what others at Uber prefer as far as serialization is concerned... I don't feel strongly about anything beyond definitely supporting `[]byte` as an option.

bensigelman:
So, I think once we've resolved the serialization and `TraceContext` interface-vs-struct issue, this is arguably ready to merge as a V0.1 thing. Agreed?

opentracing/trace_context.go
yurishkuro: @bensigelman 
A few people responded to my document with strong preference against opaque strings (one or chopped up), and in favor of one header per meta tag, e.g. 
```
UberCtx-MyTag1: myvalue1
UberCtx-MyTag2: myvalue2
```
Which is, incidentally, what I've implemented in the Python API. My only concern with this approach is the encoding of tag names and values, they need to be "safe" as http headers, but for now I left it as the responsibility of the instrumentation code.  Also, the tag names have to be case-insensitive in this approach.

yurishkuro:
Agreed, looks good for 0.1. We can also start tracking individual issues and have debates there instead of in the middle of the code.

opentracing/trace_context.go
bensigelman: Okay, sounds reasonable. This is [relatively] low on my own importance list since it hopefully won't affect `opentracing` user code one way or the other.

We have a few candidates in today, but I will try to adjust things to separate the headers for debugability.

By the way, one minor note: I saw that your pickler interface does both serialization and deserialization... I was going back and forth about that (vs what I did, where the `Serialize` is part of the `TraceContextID` interface)... I see it as a tradeoff between the symmetry of having serialization+deserialization adjacent vs. the type certainty of making `Serialize` a method on the context object itself. I think I might move towards the more symmetrical thing for this next rev.

yurishkuro: Yes, I view serialization as a separate concern from the primary function of a class like ContextID. Also, having serialization confined to a separate entity allows more flexibility in replacing it with a different implementation, should the need arise.

bensigelman:
Alright... I did a bunch of minor cleanup stuff, and also made a few slightly-more-interesting changes:

- I collapsed `TraceContext` (which had been a struct) and `TraceContextID` (which had been an interface) into a single `TraceContext` interface. The downside is that implementations will need to do the redundant work of figuring out what to do with "trace tags", but that might be an upside, too: maybe they're too expensive for some users, and this makes it easy to drop them while still using the rest of `opentracing`
- I made marshaling and unmarshaling their own interfaces and took that functionality out of `TraceContext` proper
- I split the process-identification parts of `ProcessRecorder` into their own interface, than embedded that in the (renamed) `Recorder`

I think this is good to go as v0.1. We can then get another language going (python? js? open to suggestions) and go from there.

Thanks, all!
Ben

opentracing/tracecontext.go
```diff
@@ -0,0 +1,82 @@
...
+
+// A simple interface to marshal a `TraceContext` to a binary byte array or a
+// string-to-string map.
+type TraceContextMarshaler interface {
+	// Converts the `TraceContext` into marshaled binary data (see
+	// `TraceContextUnmarshaler.UnmarshalTraceContextBinary()`).
+	MarshalTraceContextBinary(tcid TraceContext) []byte
+	// Converts the `TraceContext` into a marshaled string:string map (see
+	// `TraceContextUnmarshaler.UnmarshalTraceContextStringMap()`).
+	MarshalTraceContextStringMap(tcid TraceContext) map[string]string
```
bensigelman: @yurishkuro: don't said I never did anything for you/Uber. :)  Seriously, though, I am fine with this API but it's problematic to translate to HTTP headers since there's a lossy transformation there with capitalization, etc (at least in the core Golang libs). E.g., a `SetTraceTag` key like `"user"` becomes `"User"`, etc.

opentracing/dapperish/dapper.go
```diff
@@ -17,29 +21,149 @@ type DapperishContextID struct {
...
+		TraceID:   traceID,
+		SpanID:    spanID,
+		Sampled:   sampled,
+		traceTags: traceTags,
+	}, nil
+}
+
+func (d *DapperishTraceContextSource) MarshalTraceContextBinary(ctx opentracing.TraceContext) []byte {
+	dtc := ctx.(*DapperishTraceContext)
+	// XXX: support tags
```
bensigelman: (since this is just a demo impl, I'm leaving this out for now... I handled tags in the string map case.)

opentracing/defaulttracer.go
```diff
@@ -0,0 +1,57 @@
+package opentracing
+
+var (
+	defaultOpenTracer OpenTracer = noopOpenTracer{noopTraceContextSource{}}
```
yurishkuro: Given the convenience the methods in this file provide, I think we can assume most frameworks will be using them, tus relying on a global tracer instance of Construction API. For tracing systems willing to plug-in at the construction API rather than at the Raw API (e.g. due to performance considerations), I think we should have a way to set the default tracer explicitly, not just via `NewStandardTracer(recorder, ctxSource)`

opentracing/README.md
```diff
@@ -8,37 +8,33 @@ For background:
...
+##### Non-Singleton initialization
 
-If global singletons make you sad, use `opentracing.NewStandardTracer(...)`
-directly and manage ownership of the `opentracing.OpenTracer` explicitly.
+If you prefer direct control to singletons, use
+`opentracing.NewStandardTracer(...)` directly and manage ownership of the
+`opentracing.OpenTracer` implementation explicitly.
 
-#### Creating a Span given an existing Golang `context.Context`
+#### Creating a root Span (i.e., without a known parent)
```
yurishkuro: May be: `Starting a brand new trace by creating a root span (i.e., without a known parent)`

opentracing/tracecontext.go
yurishkuro: I'll never say that...

I did have to deal with normalizing names, e.g. to upper case, when reading/writing headers, because HTTP headers are case insensitive. I made a comment about it earlier, I think we should formalize it in the tags API, given that headers might be a rather popular format for meta:
```
SetTraceTag(caseInsensitiveKey, value string) TraceContext
```

opentracing/defaulttracer.go
bensigelman: Well, that's what `InitDefaultTracer` does, right?

opentracing/README.md
bensigelman: sure, done.

opentracing/defaulttracer.go
bensigelman: this is a great point -- disregard previous response (if you saw it via email), I misremembered what the code actually did here. will fix.

bensigelman:
I think I'll just merge this... if anyone (@yurishkuro or otherwise) has additional comments or concerns, I am 100% happy to revisit this code. Onwards...

opentracing/tracecontext.go
bensigelman: Also, btw: I ran this whole API by a friend who is very closely involved in the golang project proper, and he gave me a bunch of feedback about naming conventions, etc, hence the switch to `Marshal` and some other stuff.

That makes sense re the keys and case sensitivity... I wonder if we should also restrict them to a pretty narrow regexp, something like `(?i:[_a-z][-_a-z0-9]*)` (similar to a C identifier, but allowing hyphens).

yurishkuro: ah, yes, my built-in go lint was noticing various missteps :smile: I meant to bring it up. We should invest into setting up a Ci with unit tests, coverage, go lint, and go vet. Maybe even comparative benchmarks, I am not sure if Go's default benchmark supports comparisons.

I was considering keys validator, but I think we'd need to make it configurable so that people can turn if off in prod.  

I also noticed that you made the values url-encoded by default, while I was inclined to leave them as is. Don't have a strong opinion on that one. E.g. headers are allowed to have spaces, and url encoding makes those ugly. But there are also quite a few not allowed characters
```
  special        = "(" / ")" / "<" / ">" / "@" / ","
                 / ";" / ":" / "\" / DQUOTE / "/" / "[" 
                 / "]" / "?" / "=" / "{" / "}"
```
http://greenbytes.de/tech/webdav/draft-ietf-httpbis-p1-messaging-22.html#rfc.section.3.2.4.p.3

bensigelman: I'm really flexible (or at least suggestible) re how best to integrate this with HTTP headers. I tend to regard them as "a mess" and so my personal bias has generally been to cower in fear (c.f. my original proposal to base64 encode the binary context). But if there are people with strong, well-informed opinions, I will let them come up with sensible guidelines and encoding practices for HTTP headers.

opentracing/defaulttracer.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// Should be called as early as possible in main(), prior to calling the
+// `StartSpan*` (etc) global funcs below. Prior to calling `InitDefaultTracer`,
+// any Spans started via the `StartSpan*` globals are noops.
+//
+// See `NewStandardTracer(...)` to create an `OpenTracer` instance.
```
yurishkuro: I like it!  I would even remove that comment in line 11 - a developer enabling a tracing implementation in his application doesn't need to know about different ways of creating a tracer. If, say, `zipkin` implementation happens to be ok with standard tracer, it can expose a single method to the developer:
```Go
package zipkin

func NewTracer(...) (OpenTracer, error) {
    return NewStandardTracer(newZipkinReporter(...), newZipkinContextSource(...)), nil
}
```
So developers are never exposed to things like reporter and context source

bensigelman: That makes sense... in fact, the standard tracer should probably be in its own sub-package or something... it doesn't need to have special access to package-private stuff. I'll make that change.

yurishkuro: https://github.com/opentracing/api-golang/issues/4

Issue #4, opened 2015-12-04
Rethink package stucture


yurishkuro:
Something like this:
```
/ README, LICENSE
/ Makefile, CI files (.travis.xml or whatever)
/ opentracing / - construction API only, and globals, utils
/ opentracing / raw / - reporter and context source
/ opentracing / std / - standard tracer impl of construction API
/ examples /
/ examples / dapperish / - sample implementation
/ examples / ping-pong / - simple client/server pair illustrating the usage of tracing API w/ dapperish
```

bensigelman:
I'll send out a PR with parts of this... Go sometimes annoys me with stuff like this: I end up wanting the last segment of every path name to be meaningful, as programmers end up using it to qualify symbols. E.g., I would do `standardtracer` instead of `std` just so that programmers would end up writing `standardtracer.New`. And so on.

bensigelman:
(still need to deal with README/LICENSE/etc, as well as think about `opentracing/raw`)

yurishkuro:
@bensigelman as I am working through Python version, I've ran into a couple of issues with the packages and inheritance.

1. The main `opentracing` package has no need for Raw API. I think we've settled on the binding and instrumentation to be done at the `Tracer` level, so the Raw API becomes implementation detail of `standardtracer` package. Should we move it there? 

2. The RawSpan has certain efficiency gains in Go because it's embedded by value into the standard span. However, in Python it seems redundant, as it just introduces an unnecessary indirection for storing and reading the values.  I am thinking of merging it into standard span - any thoughts?

3. The main Tracer embeds Trace Context Source interface. This seems unnecessary, as it leaks the `NewRootTraceContext()` method into the Tracer interface.  What Tracer really needs is the two marshaler interfaces, but not the actual Source. There is some magic going on in the Go version where standardTracer manages to expose Source's methods directly by embedding it, but in Python that doesn't work, the tracer would have to re-implement the marshal methods and delegate to the Source implementation.

yurishkuro:
One more suggestion (as long as we're discussing the structure) - I would like to move `NewChild` method from `TraceContext` to `TraceContextSource`. The Source already has `NewRootTraceContext()` method, and I think it's better to keep context creation methods in a single interface, where they can share random number generator and the knowledge of how context ID is structured.

bensigelman:
@yurishkuro thanks for the thoughts/suggestions...

[...]

PR #5, opened 2015-12-04
create a standardtracer package (and some minor fallout)

bensigelman:
See #4 and the corresponding points that @yurishkuro made in #2.


examples/dapperish/dapper.go
```diff
@@ -9,8 +9,15 @@ import (
 	"sync"
 
 	"github.com/opentracing/api-golang/opentracing"
+	"github.com/opentracing/api-golang/opentracing/standardtracer"
 )
 
+func NewDapperishTracer(processName string) opentracing.OpenTracer {
```
bensigelman: (per suggestion!)

bensigelman:
I am going to :zzz: now, but I'm open to criticism about package names, etc :)

I do think that this is a good general direction, though... nice to keep the `opentracing` package proper as thin as possible.

examples/dapperish/dapper.go
yurishkuro: I think golint might complain about stutter on this one. You probably want just this: `dapperish.NewTracer()`

bensigelman: Ok, sure.

yurishkuro:
lgtm

PR #6, opened 2015-12-04
delint

bensigelman:
`golint` is so insanely pedantic, but when in Rome...


bensigelman:
(there's still plenty of work to do to make the `godoc` output look better, but I would like to wait until everything has stabilized more before dealing with that)

Issue #7, opened 2015-12-05
Serializing trace context into binary protocols


yurishkuro:
I am interested in making [TChannel](https://github.com/uber/tchannel/blob/master/docs/protocol.md) implementations instrumented with opentracing API. The tracing is built into the protocol as a first-class citizen, the "trace ID" is accessible at a fixed position in the frame without parsing the whole request, which is critical when bouncing the requests through the routers to allow routers to participate in the traces. The current limitation of 64:64:64:8 bit length will be lifted in future revs, but the limitation of a fixed length will remain.

The trace context marshaller currently does not expose separate serialization for a trace ID, only for the whole context with metadata. 
 # should we create dedicated methods to serialize just the trace ID as a binary blob?
 # can we impose a certain limit on the size of that blob?

bensigelman:
It sounds like the metadata can't be guaranteed to fit given the fixed (and presumably small) length. As such, could you substitute a custom Marshaler+Unmarshaler that skips the trace tags / metadata?

Or is there somewhere aside from that narrow region where the metadata could go? (Sounds unlikely given what you've indicated)

- Ben (mobile)

yurishkuro:
Specifically in TChannel protocol the trace ID is encoded in a fixed position with a fixed length, so that routers can access and update it efficiently. It also supports arbitrary app headers where I expect metadata to go to, but it's not being parsed out by the routers. There are different ways we can support it, one is by splitting the methods, another is to change the binary marshaller to return a tuple, e.g.
```Go 
func MarshalTraceContextBinary(tcid TraceContext) (contextId []byte, metadata []byte)
```
yet another way is to define an optional interface that a TraceContextMarshaler can implement, that exposes split contextID+metadata methods, then TChannel would only work with tracing systems that support that extra interface.

Personally, I think since the metadata (tags) is already separated in the context API itself, it's not a stretch to have marshallers treat contextID and metadata separately, at least in binary marshaller.

bensigelman:
I like that too. The reason I separated the id proper from the metadata a
few revs ago was because only the latter is actually needed to *identify* a
span. So there is a clear semantic division, and it's fine with me to reify
that. I'll make that change before Monday (or at least send out the PR).

On Saturday, December 5, 2015, Yuri Shkuro <notifications@github.com> wrote:

> Specifically in TChannel protocol the trace ID is encoded in a fixed
> position with a fixed length, so that routers can access and update it
[...]

PR #8, opened 2015-12-07
split the trace tags out from the core context ID in marshal/unmarshal

bensigelman:
(Per discussion in #7)

### DIFF
--- a/examples/dapperish/dapper.go
+++ b/examples/dapperish/dapper.go
@@ -5,7 +5,6 @@ import (
 	"encoding/binary"
 	"fmt"
 	"strconv"
-	"strings"
 	"sync"
 
 	"github.com/opentracing/api-golang/opentracing"
@@ -39,10 +38,9 @@ type DapperishTraceContext struct {
 const (
 	// Note that these strings are designed to be unchanged by the conversion
 	// into standard HTTP headers (which messes with capitalization).
-	fieldNameTraceID   = "Traceid"
-	fieldNameSpanID    = "Spanid"
-	fieldNameSampled   = "Sampled"
-	fieldNameTagPrefix = "Tag-"
+	fieldNameTraceID = "Traceid"
+	fieldNameSpanID  = "Spanid"
+	fieldNameSampled = "Sampled"
 )
 
 // NewChild complies with the opentracing.TraceContext interface.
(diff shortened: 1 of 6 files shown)


opentracing/tracecontext.go
```diff
@@ -63,10 +90,31 @@ type TraceContextMarshaler interface {
...
 	// TraceContextMarshaler.MarshalTraceContextStringMap()) into a TraceContext.
-	UnmarshalTraceContextStringMap(marshaled map[string]string) (TraceContext, error)
+	//
+	// The first parameter contains the marshaler's serialization of the core
+	// identifying information in a TraceContext instance.
+	//
+	// The first parameter contains the marshaler's serialization of the trace
+	// tags (per `SetTraceTag` and `TraceTag`) attached to a TraceContext
+	// instance.
+	UnmarshalTraceContextStringMap(
```
yurishkuro: Perhaps add a comment that it's ok to pass the same map as a value for both parameters, e.g. http headers map. The method only takes those keys from the map that it deems relevant.

opentracing/tracecontext.go
```diff
@@ -63,10 +90,31 @@ type TraceContextMarshaler interface {
...
+	) (TraceContext, error)
+
 	// Converts the marshaled string:string map (see
 	// TraceContextMarshaler.MarshalTraceContextStringMap()) into a TraceContext.
-	UnmarshalTraceContextStringMap(marshaled map[string]string) (TraceContext, error)
+	//
+	// The first parameter contains the marshaler's serialization of the core
+	// identifying information in a TraceContext instance.
+	//
+	// The first parameter contains the marshaler's serialization of the trace
```
yurishkuro: s/first/second/

yurishkuro:
looks great.

bensigelman:
done and done -- thanks.

examples/dapperish/dapper.go
```diff
@@ -63,20 +62,20 @@ func (d *DapperishTraceContext) NewChild() (opentracing.TraceContext, opentracin
 }
 
 // SetTraceTag complies with the opentracing.TraceContext interface.
-func (d *DapperishTraceContext) SetTraceTag(key, val string) opentracing.TraceContext {
+func (d *DapperishTraceContext) SetTraceTag(caseInsensitiveKey, val string) opentracing.TraceContext {
```
adriancole: I'd prefer we elaborate on this somewhere, even if this is a separate issue. For example [http/2 treats transmission of mixed case keys an error](https://http2.github.io/http2-spec/#HttpHeaders), even if comparison (at app layer) is permitted to be insensitive.


> Just as in HTTP/1.x, header field names are strings of ASCII characters that are compared in a case-insensitive fashion. However, header field names MUST be converted to lowercase prior to their encoding in HTTP/2. A request or response containing uppercase header field names MUST be treated as malformed (Section 8.1.2.6).

opentracing/tracecontext.go
```diff
@@ -39,12 +39,16 @@ type TraceContext interface {
 	// TraceContext, and that can add up to a lot of network and cpu
 	// overhead.
 	//
+	// IMPORTANT NOTE #3: Trace tags are case-insensitive: implementations may
```
adriancole: this is where my comment meant to be wrt lc transmission. Impact is that case preservation of keys is unsupported.

opentracing/utils.go
```diff
@@ -7,9 +7,12 @@ import (
 )
 
 const (
-	// OpenTracingContextHTTPHeaderPrefix precedes all opentracing-related HTTP
-	// headers.
-	OpenTracingContextHTTPHeaderPrefix = "Opentracing-Context-"
+	// OpenTracingContextIDHTTPHeaderPrefix precedes the opentracing-related
+	// ContextID HTTP headers.
+	OpenTracingContextIDHTTPHeaderPrefix = "Open-Tracing-Context-Id-"
```
adriancole: let's downcase these, knowing that http/2 will

adriancole:
LGTM

PR #14, opened 2015-12-21
restrict trace tag keys to make HTTP headers easier to manage

bensigelman:


opentracing/tracecontext.go
```diff
@@ -43,11 +48,18 @@ type TraceContext interface {
 	// wish to use them as HTTP header keys (or key suffixes), and of course
 	// HTTP headers are case insensitive.
 	//
+	// `restrictedKey` MUST match the regular expression
+	// `(?i:[a-z0-9][-a-z0-9]*)` and is case-insensitive. See
+	// CanonicalizeTraceTagKey().
+	//
```
slimsag: I think the parameter is named `caseInsensitiveKey` not `restrictedKey` ? Also it'd be nice to describe the restriction in english rather/in-addition-to regexp for those of us who are not as literate :)

opentracing/tracecontext.go
```diff
@@ -43,11 +48,18 @@ type TraceContext interface {
...
 	// Returns a reference to this TraceContext for chaining, etc.
 	SetTraceTag(caseInsensitiveKey, value string) TraceContext
 
 	// Gets the value for a trace tag given its key. Returns the empty string
 	// if the value isn't found in this TraceContext.
+	//
+	// `restrictedKey` MUST match the regular expression
+	// `(?i:[a-z0-9][-a-z0-9]*)` and is case-insensitive.
```
slimsag: same here

opentracing/tracecontext.go
slimsag: The dapperish implementation panics if the key is invalid, should all implementations do that? If so, I think it'd be good ot document it here too.

bensigelman: I deliberately didn't specify the behavior here, though I would guess most "real" implementations wouldn't crash (since they don't *have* to).

Good call re the parameter, I will fix that.

bensigelman: I kind of like the regexp, personally :)

I'll do both.

bensigelman: PS: I shouldn't have merged this so quickly... trying to do a few quick things during holiday chaos with fam...

opentracing/tracecontext.go
```diff
@@ -132,3 +144,14 @@ type TraceContextSource interface {
 	// to NewRootTraceContext().
 	NewRootTraceContext() TraceContext
 }
+
+var kTraceTagRegexp = regexp.MustCompile("^(?i:[a-z0-9][-a-z0-9]*)$")
+
+// Returns the canonicalized version of trace tag key `key`, and true if and
+// only if the key was valid.
+func CanonicalizeTraceTagKey(key string) (string, bool) {
+	if !kTraceTagRegexp.MatchString(key) {
```
yurishkuro: I think the common mismatch, especially in WSGI environments, would be underscores instead of dashes, which can be easily corrected without panicing.

Another useful trick would be to cache the raw keys into their canonical form, to avoid repetitive regexp evals, since the actual key values are most likely going to be limited in number. I've seen this done in Python [tornado.httutil.HTTPHeaders](http://tornadokevinlee.readthedocs.org/en/latest/_modules/tornado/httputil.html#HTTPHeaders)

yurishkuro: also, toLower should be called before regexp, since it won't match otherwise

opentracing/tracecontext.go
slimsag: Ah okay, cool. That makes sense, I agree having them both is nicer. Thanks a ton for this @bensigelman !

opentracing/tracecontext.go
bensigelman: The regexp starts with `(?i:` which specifies case-insensitive matching...

I hear you on caching, though I don't want to prematurely optimize...

Issue #9, opened 2015-12-10
OpenTracer.StartTrace could use map[string]interface{} to be more type safe


slimsag:
I've use log packages like [log15](https://github.com/inconshreveable/log15) which "do bad things" when you provide an uneven number of arguments.

Right [here](https://github.com/opentracing/api-golang/blob/master/opentracing/opentracer.go#L31) we open ourselves up to this. Where "do bad things" is really up to the implementation.

I think it would be nicer to change `keyValueTags` from `...interface{}` to `map[string]interface{}`.

bensigelman:
I struggle with this one... it's such a mouthful to write `map[string]interface{}{...}`.

A compromise would be to have that `keyValueTags` be of type `opentracing.Tags`, per https://github.com/opentracing/api-golang/blob/master/opentracing/raw.go#L9 ... then callers could just do something like this: http://play.golang.org/p/jFSw-sFp76

```
package main

import "fmt"

type Tags map[string]interface{}

func takesTags(t Tags) {
	for k, v := range t {
		fmt.Printf("%v -> %v\n", k, v)
	}
}

func main() {
	takesTags(Tags{
		"forty-two": 42,
		"a string":  "abcd",
	})
}
```

Best of both worlds?

bensigelman:
(I guess it would be more like `someFunction(..., opentracing.Tags{ ... })`, but still a little more palatable than `map[string]interface{}`)

slimsag:
@bensigelman I think that `opentracing.Tags{...}` would be better too. I agree writing out `map[string]interface{}{}` is a mouthful. My motivation for this is that using the map type is more sane (even in the form `type Tags map[string]interface{}`) just due to the added type safety, as it is the native Go representation for such an object.

bensigelman:
For me it's less about type safety and more about making it easy for people to generate and augment these tags things as they see fit... That's of course doable but annoying if it involves rolling/unrolling the flattened version.

yurishkuro:
+1 for type. A negative of using a map is an extra object allocation / gc pressure (I assume vararg slice is read off the stack).

slimsag:
For ideas sake, another solution could be:

```
package main

import "fmt"

type Tags map[string]interface{}

// Tags allows you to write Tags("key", "value", "key2", "value2") and returns a map.
func Tags(pairs ...interface{}) map[string]interface{} {
}

func takesTags(t map[string]interface{}) {
    for k, v := range t {
[...]

PR #15, opened 2015-12-21
set tags with help from the compiler

bensigelman:
@slimsag @yurishkuro thoughts?

### DIFF
--- a/opentracing/opentracer.go
+++ b/opentracing/opentracer.go
@@ -11,29 +11,22 @@ type OpenTracer interface {
 	// without specifying a parent Span that can be used to incorporate the
 	// newly-returned Span into an existing trace.
 	//
-	// `keyValueTags` is a [possibly empty] list of <string, interface{}> tag
-	// pairs which are applied to the returned `Span`, just as if they were
-	// applied in successive calls to `Span.SetTag()`. If the even-numbered
-	// `keyValueTags` parameters are not of `string` type, the implementation
-	// is undefined (and may `panic()`).
-	//
 	// Examples:
 	//
 	//     val tracer OpenTracer = ...
 	//
 	//     sp := tracer.StartTrace("GetFeed")
 	//
-	//     sp := tracer.StartTrace(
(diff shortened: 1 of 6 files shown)


opentracing/opentracer.go
```diff
@@ -11,29 +11,22 @@ type OpenTracer interface {
...
 	//
 	//     val tracer OpenTracer = ...
 	//
 	//     sp := tracer.StartTrace("GetFeed")
 	//
-	//     sp := tracer.StartTrace(
-	//         "HandleHTTPRequest",
-	//         "user_agent", req.UserAgent,
-	//         "lucky_number", 42)
+	//     sp := tracer.StartTrace("HandleHTTPRequest").
```
bensigelman: I don't feel strongly, but I would weakly vote for merging this PR.

opentracing/standardtracer/standardimpl.go
```diff
@@ -41,6 +41,16 @@ func (s *standardSpan) SetTag(key string, value interface{}) opentracing.Span {
 	return s
 }
 
+func (s *standardSpan) SetTags(tags opentracing.Tags) opentracing.Span {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	for k, v := range tags {
+		s.raw.Tags[k] = fmt.Sprint(v)
```
yurishkuro: @bensigelman  Ben, didn't you originally want to allow repeated tags?

yurishkuro:
lgtm

slimsag:
LGTM -- I do think this is an improvement.

opentracing/standardtracer/standardimpl.go
bensigelman: at one point, though I had some cold feet when I started to codify it... felt like it was making the simple case un-simple (and it'll be easier to add repeated tags later than to remove them).

yurishkuro:
Fixes #9

adriancole:
agreed. this is a move forward

PR #16, opened 2015-12-21
clarify restrictedKey

bensigelman:
@slimsag is this better?

### DIFF
--- a/opentracing/tracecontext.go
+++ b/opentracing/tracecontext.go
@@ -51,7 +51,9 @@ type TraceContext interface {
 	// `restrictedKey` MUST match the regular expression
 	// `(?i:[a-z0-9][-a-z0-9]*)` and is case-insensitive. That is, it must
 	// start with a letter or number, and the remaining characters must be
-	// letters, numbers, or hyphens. See CanonicalizeTraceTagKey().
+	// letters, numbers, or hyphens. See CanonicalizeTraceTagKey(). If
+	// `restrictedKey` does not meet these criteria, SetTraceTag() results in
+	// undefined behavior.
 	//
 	// Returns a reference to this TraceContext for chaining, etc.
 	SetTraceTag(restrictedKey, value string) TraceContext
--- a/opentracing/tracecontext.go
+++ b/opentracing/tracecontext.go
@@ -49,18 +49,18 @@ type TraceContext interface {
 	// HTTP headers are case insensitive.
 	//
 	// `restrictedKey` MUST match the regular expression
-	// `(?i:[a-z0-9][-a-z0-9]*)` and is case-insensitive. See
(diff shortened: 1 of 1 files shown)


yurishkuro:
+1

opentracing/tracecontext.go
```diff
@@ -49,18 +49,18 @@ type TraceContext interface {
...
 	// `restrictedKey` MUST match the regular expression
-	// `(?i:[a-z0-9][-a-z0-9]*)` and is case-insensitive. See
-	// CanonicalizeTraceTagKey().
+	// `(?i:[a-z0-9][-a-z0-9]*)` and is case-insensitive. That is, it must
+	// start with a letter or number, and the remaining characters must be
+	// letters, numbers, or hyphens. See CanonicalizeTraceTagKey().
 	//
 	// Returns a reference to this TraceContext for chaining, etc.
-	SetTraceTag(caseInsensitiveKey, value string) TraceContext
+	SetTraceTag(restrictedKey, value string) TraceContext
```
slimsag: Much better, thank you @bensigelman :+1: 

I still think we should mention that implementations should panic on invalid keys, though?

slimsag: To be clear, my concern is primarily "As a user, if I accidently call SetTraceTag with an invalid key will the app crash or gracefully ignore it etc etc etc?"

bensigelman: I'll compromise and officially call it "undefined behavior"... I like that more than specifying crash/no-crash, since different platforms may really demand different behavior.

opentracing/tracecontext.go
```diff
@@ -49,18 +49,20 @@ type TraceContext interface {
 	// HTTP headers are case insensitive.
 	//
 	// `restrictedKey` MUST match the regular expression
-	// `(?i:[a-z0-9][-a-z0-9]*)` and is case-insensitive. See
-	// CanonicalizeTraceTagKey().
+	// `(?i:[a-z0-9][-a-z0-9]*)` and is case-insensitive. That is, it must
```
adriancole: I'd prefer documenting this as down-cased, or that case won't be preserved. case-insensitivity is more about comparison than storage.

bensigelman: I'm sure there's a better way to word it, but the idea was that `TraceAttribute` (or whatever we end up calling it) will do case-insensitive matching on the `restrictedKey`.

yurishkuro: @adriancole I think we're specifically concerned with comparison. The storage format can be implementation specific: lower case, or capitalized As-Many-Http-Libraries-Do, or whatever. The main contract is that if you set an attribute with key="ABC-xyz", you can retrieve it with key="abc-Xyz".

PR #18, opened 2015-12-25
Renames OpenTracer to just Tracer; fixes minor lint issues

### LINKED ISSUE #17: Rename OpenTracer to Tracer
the package is already called `opentracing`, the usage of the interface is `opentracing.Tracer`, which does not require 'open' prefix.

yurishkuro:
fixes #17

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+
+
+.DEFAULT_GOAL := test
+
+.PHONY: test
+test:
+	go test ./...
+
+.PHONY: lint
+lint:
+	golint ./...
+
+.PHONY: vet
+vet:
+	go vet ./...
+
+.PHONY: example
+example:
+	go build -o build/dapperish-example ./examples/dapperish/.
+	./build/dapperish-example
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,13 @@
+# IntelliJ project files
+.idea/
+api-golang.iml
+api-golang.ipr
+api-golang.iws
+
+# Test results
+*.cov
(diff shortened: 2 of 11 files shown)


yurishkuro:
@bensigelman

Makefile
```diff
@@ -0,0 +1,20 @@
+
```
bensigelman: weak vote against having a Makefile (either the Makefile is super simple and thus not necessary or it is complex and a total monstrosity because, well, it's a Makefile :).

Makefile
```diff
@@ -0,0 +1,20 @@
...
+lint:
+	golint ./...
+
+.PHONY: vet
+vet:
+	go vet ./...
+
+.PHONY: example
+example:
+	go build -o build/daperish-example ./examples/dapperish/.
```
bensigelman: (if we stick with `make`, `s/daperish/dapperish/g`

bensigelman:
the `s/OpenTracer/Tracer/g` part works for me.

Makefile
yurishkuro: I think Makefile lowers the barrier to entry: `make example`. It's trivial now, but we're missing proper setup for unit test coverage and CI integration, and those add a bit more boilerplate.

Makefile
yurishkuro: fixed

Makefile
adriancole: my 2p is that the example could be in a README vs make, unless make is a commonly used tool in golang

Issue #12, opened 2015-12-11
Binary tag values


slimsag:
I saw in the docs for SetTag:

```
    // Adds a tag to the span. The `value` is immediately coerced into a string
    // using fmt.Sprint().
    //
    // If there is a pre-existing tag set for `key`, it is overwritten.
    SetTag(key string, value interface{}) Span
```

I presume this was to intentionally not allow binary `[]byte` tag values? Was there discussion around this that I missed, and could anyone link me to it?

bensigelman:
There was no discussion about it that I'm aware of... For what it's worth, I was more concerned with the `value` not-being a nested type (i.e., not a struct) and less concerned with it being a string or a `[]byte`.

Do you have a specific pre-existing use-case for `[]byte`, by the way? I can imagine some for programmatic APIs that run over trace data.

slimsag:
Currently in Appdash we allow encoding any arbitrary `struct` value, where the value is an arbitrary `[]byte` payload provided by the user.

In specific, this is very useful because clients can attach gob-encoded `struct` payloads with a span and our frontend/web UI can opt-in to rendering these payloads specially instead of just in a standard "key: value" display that we might use otherwise. We don't use this feature very extensively right now... so we wouldn't be at a strict loss without it... but I do think it would open up many possibilities in general.

For example, I might sent a space-efficient binary stack trace with a span and want to unpack it into a text form for display in the UX. I do suppose I could e.g. base64 encode during such a situation, but that seems wasteful unless we have a good reason to not support binary values.

yurishkuro:
I like that, @slimsag. Where do you encode the rendering hint? Is it in the tag name, or in the binary blob itself?

Also, how portable is this approach?  I don't think you can encode Go struct from Java, for example.

slimsag:
@yurishkuro Yes, the rendering hint is in the tag name. We consider the tag name to be the "data type name", e.g. "HTTPRequest" would unmarshal into a `HTTPRequest struct`. Portability is not a problem if you choose an encoding type that is language-agnostic. Protobuf etc would all do fine.

bensigelman:
I am convinced that we should support `[]byte` payloads for these tag annotations. How do we want to express it in the API? (`interface{}` is tricky since I'm not aware of a surefire way to convert an arbitrary `interface{}` into a `[]byte`) It would make me sad to add more methods, but I guess we could add `SetBinaryTag(key string, value []byte)` in parallel to the current `SetTag(...)`.

[...]

PR #19, opened 2015-12-26
Adds ext/tags mechanism for adding common tags to the span

yurishkuro:
@bensigelman

### DIFF
--- a/examples/dapperish.go
+++ b/examples/dapperish.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"github.com/opentracing/api-golang/opentracing"
+	"github.com/opentracing/api-golang/examples/dapperish"
 )
 
 func client() {
@@ -78,7 +79,7 @@ func server() {
 }
 
 func main() {
-	opentracing.InitDefaultTracer(NewTracer("dapperish_tester"))
+	opentracing.InitDefaultTracer(dapperish.NewTracer("dapperish_tester"))
 
 	go server()
 	go client()
--- a/examples/dapperish.go
+++ b/examples/dapperish.go
@@ -12,8 +12,8 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/opentracing/api-golang/opentracing"
 	"github.com/opentracing/api-golang/examples/dapperish"
+	"github.com/opentracing/api-golang/opentracing"
 )
 
 func client() {
--- a/examples/dapperish/dapper.go
+++ b/examples/dapperish/dapper.go
@@ -1,4 +1,4 @@
-package main
+package dapperish
 
 import (
 	"bytes"
--- a/opentracing/ext/tags.go
+++ b/opentracing/ext/tags.go
@@ -5,22 +5,23 @@ import (
 )
 
 var (
-	// RPC tags as symmetrical, each can be emitted by either client-side of server-side
+	// PeerXXX tags can be emitted by either client-side of server-side to describe
+	// the other side/service in a peer-to-peer communications, like an RPC call.
 
-	// RPCService records the service name of the RPC peer
-	RPCService = &stringTag{"rpc.service"}
+	// PeerService records the service name of the peer
+	PeerService = &stringTag{"peer.service"}
 
-	// RPCHostname records the host name of the RPC peer
-	RPCHostname = &stringTag{"rpc.hostname"}
+	// PeerHostname records the host name of the peer
+	PeerHostname = &stringTag{"peer.hostname"}
 
-	// RPCHostIPv4 records IP v4 host address of the RPC peer
-	RPCHostIPv4 = &uint32Tag{"rpc.ipv4"}
+	// PeerHostIPv4 records IP v4 host address of the peer
+	PeerHostIPv4 = &uint32Tag{"peer.ipv4"}
 
-	// RPCHostIPv6 records IP v6 host address of the RPC peer
-	RPCHostIPv6 = &stringTag{"rpc.ipv6"}
+	// PeerHostIPv6 records IP v6 host address of the peer
+	PeerHostIPv6 = &stringTag{"peer.ipv6"}
 
-	// RPCPort records port number of the RPC peer
-	RPCPort = &uint16Tag{"rpc.port"}
+	// PeerPort records port number of the peer
+	PeerPort = &uint16Tag{"peer.port"}
 )
 
 type stringTag struct {
--- a/opentracing/ext/tags.go
+++ b/opentracing/ext/tags.go
@@ -1,0 +1,51 @@
+package ext
+
+import (
+	"github.com/opentracing/api-golang/opentracing"
+)
+
+var (
+	// RPC tags as symmetrical, each can be emitted by either client-side of server-side
+
+	// RPCService records the service name of the RPC peer
+	RPCService = &stringTag{"rpc.service"}
+
+	// RPCHostname records the host name of the RPC peer
+	RPCHostname = &stringTag{"rpc.hostname"}
+
+	// RPCHostIPv4 records IP v4 host address of the RPC peer
+	RPCHostIPv4 = &uint32Tag{"rpc.ipv4"}
+
+	// RPCHostIPv6 records IP v6 host address of the RPC peer
+	RPCHostIPv6 = &stringTag{"rpc.ipv6"}
+
+	// RPCPort records port number of the RPC peer
+	RPCPort = &uint16Tag{"rpc.port"}
+)
+
+type stringTag struct {
+	Key string
+}
+
+// Add adds a string tag to the `span`
+func (tag *stringTag) Add(span opentracing.Span, value string) {
+	span.SetTag(tag.Key, value)
+}
+
+type uint32Tag struct {
+	Key string
+}
+
+// Add adds a uint32 tag to the `span`
+func (tag *uint32Tag) Add(span opentracing.Span, value uint32) {
+	span.SetTag(tag.Key, value)
+}
+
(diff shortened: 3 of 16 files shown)


examples/dapperish/dapper.go
```diff
@@ -1,4 +1,4 @@
-package main
+package dapperish
```
yurishkuro: I had to rename the package to be able to import it into tests, since we don't have any other working implementation of the trace context source.

opentracing/standardtracer/standardimpl.go
```diff
@@ -37,7 +37,7 @@ func (s *standardSpan) StartChild(operationName string) opentracing.Span {
 func (s *standardSpan) SetTag(key string, value interface{}) opentracing.Span {
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	s.raw.Tags[key] = fmt.Sprint(value)
+	s.raw.Tags[key] = value
```
yurishkuro: Removed conversion to string, as it was making the new tests fail. I think conversion should be done by the Recorder, if necessary.

examples/dapperish/dapper.go
yurishkuro: hm, looks like this broke go-build, looking into it

examples/dapperish.go
```diff
@@ -78,7 +79,7 @@ func server() {
 }
 
 func main() {
-	opentracing.InitDefaultTracer(NewTracer("dapperish_tester"))
+	opentracing.InitDefaultTracer(dapperish.NewTracer("dapperish_tester"))
```
yurishkuro: moved one level up to keep it in `main` package, while the actual dapperish implementation is in `dapperish` subpackage

opentracing/ext/tags.go
```diff
@@ -0,0 +1,49 @@
+package ext
```
yurishkuro: This is the main change in this PR

opentracing/ext/tags_test.go
```diff
@@ -0,0 +1,35 @@
+package ext_test
+
+import (
+	"testing"
+
+	"github.com/opentracing/api-golang/examples/dapperish"
```
bensigelman: maybe we should rename `examples` to `testing`... seems weird to have anything under `api-golang/opentracing/` depending on anything under `api-golang/examples/`, even if it's just a test.

yurishkuro: really it was just me cheating. I've already added testutils package, and for this test I only need minimal set of functionality from the Source implementation, so we could just copy a DaperishTraceContextSource under test utils

bensigelman: that seems right.

opentracing/standardtracer/standardimpl.go
bensigelman: That seems better, agreed.

opentracing/ext/tags.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"github.com/opentracing/api-golang/opentracing"
+)
+
+var (
+	// RPCTargetService records target service name for an RPC call
+	RPCTargetService = &stringTag{"rpc.target"}
```
bensigelman: How about `"rpc.target.service"`, `"rpc.target.hostname"`, etc? (and the peer can fill in `"rpc.source.hostname"` and `"...ipv[46]"`, etc, albeit without a `service`)

bensigelman: One other thing: I don't like how dapper+zipkin represent an RPC as a single span... it seems like a better fit (IMO) to model it as two closely-related spans. What do you think about that? Too painful of a transition? If so, though, I could imagine annotations here that are specialized for the source and target roles specifically... perhaps their `Add()` methods would take different sets of parameters.

opentracing/ext/tags.go
```diff
@@ -0,0 +1,49 @@
...
+	// RPCPort records port number of RPC target service
+	RPCPort = &uint16Tag{"rpc.port"}
+)
+
+type stringTag struct {
+	Key string
+}
+
+// Add adds a string tag to the `span`
+func (tag *stringTag) Add(span opentracing.Span, value string) {
```
bensigelman: per the above, we could have something like `RPCSourceAnnotations` with an `Add(span, peerHostname, peerService, peerIPv4, peerIPv6, peerPort)` sort of signature. no strong feelings, just an idea.

testutils/assert/assert.go
```diff
@@ -0,0 +1,57 @@
+package assert
```
bensigelman: what do you think about vendoring in one of the catch-all golang test packages?

testutils/recorder.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"sync"
+
+	"github.com/opentracing/api-golang/opentracing"
+)
+
+// InMemoryRecorder is a simple thread-safe implementation of opentracing.Recorder
+// that stores all reported spans in memory, accessible via reporter.GetSpans()
+type InMemoryRecorder struct {
```
bensigelman: gracias

bensigelman:
LGTM with the caveat that all of this stuff will probably have to change as we move forward with proper bindings, etc. Thanks!

opentracing/ext/tags.go
yurishkuro: Just pushed an update, that made your comments hidden.  I basically changed the name to just `rpc.service`, with the intention of making these names symmetrical, independent of client / server side, They just refer to the 'peer' of the current span's role, I think it makes it cleaner and doesn't pollute the tags namespace, what do you think?

I also don't like Zipkin's way of merging client and server spans into one. On the plus side, current API works even with Zipkin, as I have just proven to myself by migrating our Python client to a Python version of opentracing API (fyi - I am waiting for Uber OSS team approval to make the repo public). Changing Zipkin to two-span model will be somewhat painful, I think there's a lot of assumptions, especially in the UI, based on that model. But I think @adriancole might be up to it with v2 initiative :-)

But I think we may need some additional tag, like `is_client`, to make the tracer distinguish client-side from server-side spans. Without it Zipking tracer can't emit the right ss/sr/cs/cr annotations. In most cases Zipkin tracer can infer it by assuming start/join-trace are server-side, and start_child is client-side, but there are cases when start-trace needs to be client-side. My previous API was allowing an `is_client` flag in the start_trace method.

bensigelman: There is some intrinsic asymmetry (e.g., the "port" of the client is almost always ephemeral and rarely interesting, the client "service" is impossible to determine in some RPC systems), and – as you mention – some non-intrinsic asymmetry that nevertheless crops up due to implementation choices, like the ss/sr/cs/cr annotations.

But: I think we could still do well to try and consolidate... asymmetry or not, the fields have the same semantic meaning on client and server even if only a subset is available in either case. Howa bout calling them `"rpc.peer.*"` instead of `"rpc.*"`, though? Or even `"peer.*"` if you want to keep things more concise. Dapper newbies were sometimes confused by the fact that the peer was the one recording critical identity information.

bensigelman: Oh, another idea: we could record the `"rpc.peer.process_guid"` for implementations that expose one, then pass that guid over the wire along with the in-band context data. The advantage here would be a pointer – within the tracing system – to more comprehensive information about the peer than just the host:service:port. (I realize this probably sounds like a lot of work in a zipkin world, but wanted to mention it... Thankfully, given the nature of `ext/`, this can be either/or.

opentracing/ext/tags.go
yurishkuro: not sure how useful it would be, since nobody would usually supply all 5 args, as they overlap. Right now people can set just the tags they have. Also, since this is a standard helper package, as opposed to implementation-specific, it would still need to break apart the 5 args and do 5 individual `set_tag` calls. I think it would make sense to have richer signatures only if we go with an abstract factory pattern.

I see how it might be helpful for source/target separation, but see my comment above about using symmetric tag names for client/server.

testutils/assert/assert.go
yurishkuro: we could, although the whole concept of vendoring dependencies makes me ill, I have to deal with it in our everyday Go development and I can't wait until we switch to a sane dep mgr. I've learned to appreciate Python's virtualenv.

bensigelman: I wouldn't want to risk illness...

Yes, the vendoring situation is a mess! Also bugs me that the golang authors didn't include these utterly basic/obvious `assert`-like testing utilities that absolutely everyone wants to use. Oh well.

bensigelman:
Oh, per #12: can you update the `Span.SetTag` comment?

opentracing/ext/tags.go
yurishkuro: Changing to `peer.*`

I'll hold off on process-guid, as I am not yet sure how that can be captured. In fact, it sounds like something that the tracing system can capture automatically, without participation from instrumentation code. In order to propagate it with in-band context we'd need yet another type of context data, one whose lifespan is one p2p call, as opposed to the lifespan of the whole trace.

adriancole: Made this comment somewhere else, but I kindof like google-cloud-tracing's SpanKind for annotating the "side" of the span (no pun intended).

peer is ok as it solves the need for a separate namespace for client vs server. The tradeoff is that some folks will feel the word peer less clear than just saying server. Ex. consider those writing client libraries... they only talk about the server and visa versa. Using peer optimizes for those writing both sides (ex things like finagle, grpc, etc), which I suspect is a subset of the total audience of the api.

Issue #11, opened 2015-12-10
Goroutine safety guarantees


slimsag:
From looking at the interfaces currently; I do not see much mention of what can / cannot be called concurrently. When dealing with multiple implementations of the same interfaces, I think it is especially important to be partisans of defining what both implementations and users of these implementations should expect.

I'll have more time to look into this over the next few days, so apologies for no specific examples at this moment.

yurishkuro:
Should we not assume all methods must be thread-safe?

slimsag:
That sounds okay to me. But I think we should document this somewhere.

bensigelman:
I agree that it should be documented... I'll close this when I get around to it (sadly dealing with some non-`opentracing` stuff today :).

slimsag:
Fixed by #21

PR #21, opened 2015-12-28
clarify about concurrency / safety

bensigelman:


bensigelman:
Per #11.

yurishkuro:
LGTM. 

@slimsag did you want to see anything else?

slimsag:
LGTM.

We will also want to have this be a part of the Go package docs at (https://godoc.org/github.com/opentracing/api-golang/opentracing) but this can be done as a follow-up. I'll see if I can send a PR for this tomorrow.

bensigelman:
@slimsag makes sense re the godocs. We need a package-level comment there; hopefully it can point to the main `opentracing.github.io` site for general background.

PR #22, opened 2015-12-29
move `Recorder` and `Raw*` into `standardtracer` and `s/Default/Global/`

bensigelman:


examples/dapperish/dapper.go
```diff
@@ -19,8 +19,8 @@ func NewTracer(processName string) opentracing.Tracer {
 		NewTraceContextSource())
 }
 
-// DapperishTraceContext is an implementation of opentracing.TraceContext.
-type DapperishTraceContext struct {
+// TraceContext is an implementation of opentracing.TraceContext.
+type TraceContext struct {
```
bensigelman: `go lint` didn't like the `dapperish.DapperishTraceContext` stutter.

opentracing/noop.go
```diff
@@ -4,7 +4,6 @@ import "golang.org/x/net/context"
 
 type noopTraceContext struct{}
 type noopSpan struct{}
-type noopRecorder struct{}
```
bensigelman: unused

bensigelman:
See #4

examples/dapperish.go
```diff
@@ -12,7 +12,9 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/opentracing/api-golang/examples/dapperish"
+	// `go imports` was stumbling over dapperish since it's both a package name
+	// and a source file.
+	dapperish "github.com/opentracing/api-golang/examples/dapperish"
```
slimsag: There is no `go imports` command, did you mean `goimports` ?

Also, `goimports` only looks at `package` declarations -- not source file names..

bensigelman: oh, right, `goimports` -- I hook it up through `vim` and never invoke it manually.

TL;DR, whenever I save a file it pipes it through `goimports` and it was removing the line despite using the `dapperish` package. This fixed it. I can investigate further if you're curious.

slimsag: Hmm. It should only be removing the import if `dapperish` isn't used in the code anywhere.. In general I think we should avoid such hacks / workarounds for bugs in tools; but since this is in the examples / unrelated to the main project I think it's OK to leave it as is

bensigelman: yeah, I will try to upgrade my vim stuff and see if that fixes it. out of curiosity I ran `goimports` directly and it worked fine.

slimsag:
I only gave it a brief look, but overall LGTM.

examples/dapperish.go
bensigelman: ... and after updating the vim stuff the issue went away. I'll update the PR.

yurishkuro:
LGTM, but it seems the `raw.go` should also move into standardtracer, along with the reporter.

bensigelman:
Oh, right -- neglected to do that part. One minute...

bensigelman:
(done)

bensigelman:
I dealt with a few stupid merge conflicts, but I suspect this still LGTY since the only meaningful change was getting raw.go into standardtracer/.

Issue #20, opened 2015-12-28
debug flag


yurishkuro:
EDIT: the actual use case: a mobile developer has a toggle in the app that allows him to mark all requests the app makes to the server with "please trace" flag (e.g. http header or by other means). When the server receives such request and starts a new trace, it wants to mark that trace as "debug" to ensure that it is preserved regardless of the sampling decisions.

----
It would be nice for start_trace method to accept an optional debug flag. Ultimately, all tracing systems will have to do some form of sampling, either before or after the trace. A debug flag allows users to force the sampling of the trace regardless of other sampling decisions.

While we can handle debug flag as an optional tag using the approach in #19, for systems that make upfront sampling decisions there are certain housekeeping operations that can be skipped (for efficiency reasons) when the span is being created if the tracer decided not to sample it. Examples may include adding various tags to the span, that would be a waste of CPU clocks and memory for non-sampled span. Therefore, setting a debug flag after span creation may cause those spans to be missing the tags they would've gotten had it been known that they are going to be sampled.

bensigelman:
Why not make debug-ness the responsibility of the Recorder? (That's what I'd had in mind) I completely agree that it's an important use case, but I'm not clear why it needs to be part of the StartTrace signature.

yurishkuro:
I updated the description with the actual use case. The "debug"-ness is a property of an individual trace controlled by some custom property of the incoming request. It's not a mode of the Recorder, the same tracer/recorded should be able to accommodate both debug and non-debug traces.

bensigelman:
Dapper dealt with this by propagating a "trace_mask" along with the "trace_id" and "span_id" over the wire. The mask encoded information about the sampling rate and whether there was an "explicit" sampling or the usual pseudorandom sampling.

Somewhat related: I think there's definitely value (is there even an alternative??) to propagating whether and how the current context is sampled for systems like zipkin/dapper that do a priori sampling, I wouldn't want to equate that with a "debug" context. I.e., maybe all debug contexts are explicitly sampled, but all explicitly sampled contexts are not debug contexts.

yurishkuro:
Zipkin also encodes debug / sampled as bits in the `flags` trace context field, but that doesn't help with the question of how the debug bit gets set when starting a new trace.

Interesting point on debug context, it indeed can mean more than just forced sampling, but how do we mark a context as debug? Users don't even have a say in context creation, all they do is call start_trace.

bensigelman:
[...]

PR #23, opened 2015-12-29
add StartSpanWithContext, remove TraceContext.NewChild, s/TraceTag/TraceAttribute/

bensigelman:
See #20

### DIFF
--- a/opentracing/tracecontext.go
+++ b/opentracing/tracecontext.go
@@ -14,10 +14,10 @@ import (
 // TraceContexts are sufficient to propagate the, well, *context* of a
 // particular trace between processes.
 //
-// TraceContext also support a simple string map of "trace tags". These trace
-// tags are special in that they are propagated *in-band*, presumably alongside
-// application data. See the documentation for SetTraceAttribute() for more
-// details and some important caveats.
+// TraceContext also support a simple string map of "trace attributes". These
+// trace attributes are special in that they are propagated *in-band*,
+// presumably alongside application data. See the documentation for
+// SetTraceAttribute() for more details and some important caveats.
 type TraceContext interface {
 	// SetTraceAttribute sets a tag on this TraceContext that also propagates
 	// to future TraceContext children per TraceContext.NewChild.
@@ -27,20 +27,20 @@ type TraceContext interface {
 	// app can make it, transparently, all the way into the depths of a storage
 	// system), and with it some powerful costs: use this feature with care.
 	//
-	// IMPORTANT NOTE #1: SetTraceAttribute() will only propagate trace tags to
-	// *future* children of the TraceContext (see NewChild()) and/or the
-	// Span that references it.
+	// IMPORTANT NOTE #1: SetTraceAttribute() will only propagate trace
+	// attributes to *future* children of the TraceContext (see NewChild())
+	// and/or the Span that references it.
 	//
 	// IMPORTANT NOTE #2: Use this thoughtfully and with care. Every key and
 	// value is copied into every local *and remote* child of this
 	// TraceContext, and that can add up to a lot of network and cpu
 	// overhead.
 	//
-	// IMPORTANT NOTE #3: Trace tags are case-insensitive: implementations may
-	// wish to use them as HTTP header keys (or key suffixes), and of course
-	// HTTP headers are case insensitive.
+	// IMPORTANT NOTE #3: Trace attributes keys have a restricted format:
+	// implementations may wish to use them as HTTP header keys (or key
+	// suffixes), and of course HTTP headers are case insensitive.
 	//
-	// `restrictedKey` MUST match the regular expression
+	// As such, `restrictedKey` MUST match the regular expression
 	// `(?i:[a-z0-9][-a-z0-9]*)` and is case-insensitive. That is, it must
 	// start with a letter or number, and the remaining characters must be
 	// letters, numbers, or hyphens. See CanonicalizeTraceAttributeKey(). If
@@ -67,7 +67,7 @@ type TraceContextMarshaler interface {
 	// the core identifying information in `tc`.
 	//
 	// The second return value must represent the marshaler's serialization of
-	// the trace tags, per `SetTraceAttribute` and `TraceAttribute`.
+	// the trace attributes, per `SetTraceAttribute` and `TraceAttribute`.
 	MarshalTraceContextBinary(
 		tc TraceContext,
 	) (
@@ -82,7 +82,7 @@ type TraceContextMarshaler interface {
 	// the core identifying information in `tc`.
 	//
 	// The second return value must represent the marshaler's serialization of
-	// the trace tags, per `SetTraceAttribute` and `TraceAttribute`.
+	// the trace attributes, per `SetTraceAttribute` and `TraceAttribute`.
 	MarshalTraceContextStringMap(
 		tc TraceContext,
 	) (
@@ -101,7 +101,7 @@ type TraceContextUnmarshaler interface {
 	// identifying information in a TraceContext instance.
 	//
 	// The second parameter contains the marshaler's serialization of the trace
-	// tags (per `SetTraceAttribute` and `TraceAttribute`) attached to a
+	// attributes (per `SetTraceAttribute` and `TraceAttribute`) attached to a
 	// TraceContext instance.
 	UnmarshalTraceContextBinary(
 		traceContextID []byte,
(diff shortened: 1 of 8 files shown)


opentracing/tracecontext.go
```diff
@@ -51,18 +43,18 @@ type TraceContext interface {
...
-	// letters, numbers, or hyphens. See CanonicalizeTraceTagKey(). If
-	// `restrictedKey` does not meet these criteria, SetTraceTag() results in
-	// undefined behavior.
+	// letters, numbers, or hyphens. See CanonicalizeTraceAttributeKey(). If
+	// `restrictedKey` does not meet these criteria, SetTraceAttribute()
+	// results in undefined behavior.
 	//
 	// Returns a reference to this TraceContext for chaining, etc.
-	SetTraceTag(restrictedKey, value string) TraceContext
+	SetTraceAttribute(restrictedKey, value string) TraceContext
```
bensigelman: on the fence about calling this `SetTraceAttribute` or just `SetAttribute`... went for the more explicit thing since this – while powerful – will not be called frequently.

opentracing/tracer.go
```diff
@@ -23,10 +23,25 @@ type Tracer interface {
...
+	// `ctx` MUST be created by the `Tracer`'s embedded `TraceContextSource`.
+	//
+	//
+	// Note that the following calls are equivalent
+	//
+	//     tracer.StartSpanWithContext(opName, tracer.NewRootTraceContext())
+	//
+	//     tracer.StartTrace(opName)
+	//
+	StartSpanWithContext(operationName string, ctx TraceContext) Span
```
bensigelman: I realized that it's a misnomer to call this `StartTraceWithContext` since `ctx` may not be a root context. In fact, `StartTrace` and `JoinTrace` can both be implemented in terms of `StartSpanWithContext`.

bensigelman:
(There will be some minor merge conflicts with #22 but I wanted to keep the PRs separate)

opentracing/tracecontext.go
```diff
@@ -16,26 +16,18 @@ import (
 //
 // TraceContext also support a simple string map of "trace tags". These trace
 // tags are special in that they are propagated *in-band*, presumably alongside
```
yurishkuro: s/tags/attributes/g

opentracing/tracecontext.go
yurishkuro: TraceAttribute sgtm.

testutils/trace_context.go
```diff
@@ -7,43 +7,48 @@ import (
 // SimpleTraceContextSource is a dummy implementation of TraceContextSource.
 type SimpleTraceContextSource struct{}
 
-// NewRootTraceContext imlements NewRootTraceContext of opentracing.TraceContextSource.
+// NewRootTraceContext implements NewRootTraceContext of opentracing.TraceContextSource.
 func (source *SimpleTraceContextSource) NewRootTraceContext() opentracing.TraceContext {
 	return nil
```
yurishkuro: should this be symmetric with NewChild and return (span, tags)?

yurishkuro:
LGTM

testutils/trace_context.go
bensigelman: Hard to say. I was sad about the (span, tags) thing for NewChild but felt like I was painted into a corner by the motivating use case. I didn't have such a motivating real-world use case for this one, though I don't like the smell of the asymmetry.

opentracing/tracecontext.go
bensigelman: fixed a bunch of stuff in this file, thanks.

testutils/trace_context.go
yurishkuro: we should keep track of the use cases that influence API shape somehow, maybe by (see #xxx) in comments. I don't recall why we needed child tags to be returned from context creation.

bensigelman: it allows the child to include a "parent_id" span tag...

opentracing/tracecontext.go
adriancole: prefer just Attribute as chaining looks redundant

opentracing/tracecontext.go
```diff
@@ -14,55 +14,47 @@ import (
 // TraceContexts are sufficient to propagate the, well, *context* of a
 // particular trace between processes.
 //
-// TraceContext also support a simple string map of "trace tags". These trace
-// tags are special in that they are propagated *in-band*, presumably alongside
-// application data. See the documentation for SetTraceTag() for more details
-// and some important caveats.
+// TraceContext also support a simple string map of "trace attributes". These
```
adriancole: The word trace hints at scope (if coming from zipkin/dapper). Is it assumed that these attributes will add or replace regardless of span depth?

opentracing/tracecontext.go
```diff
@@ -14,55 +14,47 @@ import (
...
 	//
-	// The returned TraceContext type must be the same as the type of the
-	// TraceContext implementation itself.
-	NewChild() (childCtx TraceContext, childSpanTags Tags)
-
-	// SetTraceTag sets a tag on this TraceContext that also propagates to
-	// future TraceContext children per TraceContext.NewChild.
-	//
-	// SetTraceTag() enables powerful functionality given a full-stack
+	// SetTraceAttribute() enables powerful functionality given a full-stack
```
adriancole: omg we are full-stack devs!

opentracing/tracecontext.go
```diff
@@ -14,55 +14,47 @@ import (
...
 	// opentracing integration (e.g., arbitrary application data from a mobile
 	// app can make it, transparently, all the way into the depths of a storage
 	// system), and with it some powerful costs: use this feature with care.
 	//
-	// IMPORTANT NOTE #1: SetTraceTag() will only propagate trace tags to
-	// *future* children of the TraceContext (see NewChild()) and/or the
-	// Span that references it.
+	// IMPORTANT NOTE #1: SetTraceAttribute() will only propagate trace
```
adriancole: ic.. this is the confusing part. PropagationAttribute would make more sense in my brain, as it only affects downstream, not the enclosing trace.

opentracing/tracecontext.go
adriancole: after looking at the rest of the code, I don't think Trace clarifies things immediately, especially when this only applies downstream. I'd strongly prefer a word like PropagationAttribute or similar as it doesn't imply trace-scope

opentracing/tracer.go
adriancole: can't think of anything better at the moment. Thanks for taking a stab.

opentracing/tracecontext.go
bensigelman: @adriancole (cc @yurishkuro) I hear you on wanting a word that gets at the downstream-ness. We could have a setter/getter pair like `PropagateAttribute` and `Attribute` (or `getAttribute` in languages that include `get` in their accessor names). Thoughts on that?

opentracing/tracecontext.go
bensigelman: webscale

opentracing/tracecontext.go
yurishkuro: Elsewhere I mentioned that I had people asking me about upstream propagation as well, e.g. for controlling an http cookie from a downstream layer. I don't think it would be technically very difficult to support, subject to some conflict resolution policy. Given that possibility, do we actually need to emphasize downstream-ness of these attributes?

It might be more useful to have an additional dimension to the Trace Attributes that clarifies the propagation semantics: downstream all the way, downstream one level only, upstream, bidirectional, etc. Unfortunately, I don't have concrete use cases just yet for different scenarios, will try to acquire some this week.

bensigelman: So, imagine this tree of spans:

```
              A
              |
         +----+----+
         |         |
         B         C
         |         |
      +--+--+   +--+--+
      |     |   |     |
      D     E   F     G
```

If someone decides to set an attribute "upstream" in `F`, `D` or `E` may have already finished... And of course there's no formal requirement (in the data model) that children finish before their parents. So I do think the upstream case is more complex than the downstream case. If we want to try to future-proof the API, we could call it `PropagateAttributeDown` or `PropagateAttributeDownstream`?? A mouthful for sure, though I'm comforted by how infrequently this will be called.

yurishkuro: @uber/rpc

yurishkuro: I still think upstream propagation semantic is fairly intuitive. It has certain limitations that people must understand before using it, of course.

> there's no formal requirement (in the data model) that children finish before their parents.

If an RPC is made with expectation of a result returned to the caller, then I think it's reasonable to assume the parent span must last longer than the children. Caveat emptor.

@bensigelman are you suggesting PropagateAttributeDown as a method name?  Why not this:
```
setTraceAttribute(key, value, propagationMode)
``

bensigelman: Well, some spans are not RPCs... and even when they are, I saw a relatively common pattern at Google where an RPC server would return to the RPC client _before_ doing an insert into a cache service... as such, the server could take longer than the client.

Re `setTraceAttribute`, I was responding to @adriancole's comment: "I'd strongly prefer a word like PropagationAttribute or similar as it doesn't imply trace-scope"

PR #24, opened 2015-12-29
opentracing: make HTTP header prefix constants more readable

slimsag:
- Remove `OpenTracing` prefix as the package is already `opentracing`.
- Add some whitespace between the comments.

### DIFF
--- a/opentracing/utils.go
+++ b/opentracing/utils.go
@@ -7,12 +7,13 @@ import (
 )
 
 const (
-	// OpenTracingContextIDHTTPHeaderPrefix precedes the opentracing-related
-	// ContextID HTTP headers.
-	OpenTracingContextIDHTTPHeaderPrefix = "Open-Tracing-Context-Id-"
-	// OpenTracingTagsHTTPHeaderPrefix precedes the opentracing-related
-	// trace-tags HTTP headers.
-	OpenTracingTagsHTTPHeaderPrefix = "Open-Tracing-Trace-Tags-"
+	// ContextIDHTTPHeaderPrefix precedes the opentracing-related ContextID HTTP
+	// headers.
+	ContextIDHTTPHeaderPrefix = "Open-Tracing-Context-Id-"
+
+	// TagsHTTPHeaderPrefix precedes the opentracing-related trace-tags HTTP
+	// headers.
+	TagsHTTPHeaderPrefix = "Open-Tracing-Trace-Tags-"
 )
 
 // AddTraceContextToHeader marshals TraceContext `ctx` to `h` as a series of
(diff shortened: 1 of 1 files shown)


bensigelman:
thanks!

yurishkuro:
To be honest, I think these constants should not be in `opentracing/utils.go`. The methods in the utils should take the header names as arguments, and the constants should be in `opentracing/standard`, if exist at all. Remember that `opentracing/standard` is not a functional implementation by itself, it still requires a real implementation for context source that is ultimately responsible for serialization. Therefore, I would completely expect two different tracing implementations to use different header names, if only to avoid false positives in interoperability.

slimsag:
@yurishkuro good point about moving them to `opentracing/standard`. I agree with that.

re: `two different tracing implementations to use different header names` I want to understand further, shouldn't the headers and values generally be interoperable between implementations? or do we assume that one impl. might filter out tags that it cannot handle in the api-golang layer but not as part of demarshaling the tags/headers? maybe you see some other compatibility issue and I just missed it?

yurishkuro:
The interoperability between tracing systems has not been the goal of the API so far. It maybe an interesting direction down the road, but for now the main philosophy was that if a company has 100 different pieces of software all instrumented with OTAPI, it can trivially swap the underlying tracing implementations, as long as the swap is done across all 100 services.

The goal of interoperability would be different: to have N services run on tracing system A, and the other 100-N services on system B. The goal itself is somewhat questionable, as you don't get complete traces in either A or B. But assuming it is a goal, the challenge would be to come up with a single standard format for trace context serialization, including standardizing on http header names, etc. The current API is not trying to address that at all - if all 100 services run the same tracing system, then that system can do serialization of the context any way it sees fit, including custom header names like B3 - Big Brother Bird.

slimsag:
I see your point. I'm aware of the differences between what we're doing here vs. a single standard format for serialization. I was primarily questioning if there were any direct incompatibilities with multiple tracing impls sharing the same headers ... but I see how it's a slippery slope to go down, and a single common tracing format is not something I think we should delve into right now due to the very large scope.

I'll send a PR later to remove these header consts entirely.

opentracing/utils.go
```diff
@@ -7,12 +7,13 @@ import (
...
 const (
-	// OpenTracingContextIDHTTPHeaderPrefix precedes the opentracing-related
-	// ContextID HTTP headers.
-	OpenTracingContextIDHTTPHeaderPrefix = "Open-Tracing-Context-Id-"
-	// OpenTracingTagsHTTPHeaderPrefix precedes the opentracing-related
-	// trace-tags HTTP headers.
-	OpenTracingTagsHTTPHeaderPrefix = "Open-Tracing-Trace-Tags-"
+	// ContextIDHTTPHeaderPrefix precedes the opentracing-related ContextID HTTP
+	// headers.
+	ContextIDHTTPHeaderPrefix = "Open-Tracing-Context-Id-"
```
adriancole: bikeshed: prefer OpenTracing to Open-Tracing. there's no Open org that we're a part of, unlike google cloud which have Cloud-Foo and Cloud-Bar

bensigelman: FYI, I went with `Open-Tracing` because I was sad when I saw how golang's http libs turned `OpenTracing` into `Opentracing`. Not a good reason, I admit, but there you have it :)

Also, I don't have strong feelings about this -- happy to change it if someone else does!

yurishkuro:
@slimsag I booked a separate issue for that, #29

PR #26, opened 2015-12-29
opentracing/ext: simplify internal tag types

slimsag:
- No need for struct because there is only one field.
  - Do not place the values on the heap / use pointer as they are small.

### DIFF
--- a/opentracing/ext/tags.go
+++ b/opentracing/ext/tags.go
@@ -9,44 +9,38 @@ var (
 	// the other side/service in a peer-to-peer communications, like an RPC call.
 
 	// PeerService records the service name of the peer
-	PeerService = &stringTag{"peer.service"}
+	PeerService = stringTag("peer.service")
 
 	// PeerHostname records the host name of the peer
-	PeerHostname = &stringTag{"peer.hostname"}
+	PeerHostname = stringTag("peer.hostname")
 
 	// PeerHostIPv4 records IP v4 host address of the peer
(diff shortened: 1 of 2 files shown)


opentracing/ext/tags.go
```diff
@@ -9,44 +9,38 @@ var (
...
 
 	// PeerPort records port number of the peer
-	PeerPort = &uint16Tag{"peer.port"}
+	PeerPort = uint16Tag("peer.port")
 )
 
-type stringTag struct {
-	Key string
-}
+type stringTag string
```
slimsag: If others are opposed to writing `string(theTag)` instead of `theTag.Key` I can revert that part of this change; but I def. think we should not be using pointer values for these.

yurishkuro:
Nice!  I forgot about this Go's feature.  LGTM.

bensigelman:
Also LGTM, thanks.

opentracing/ext/tags.go
```diff
@@ -9,44 +9,38 @@ var (
 	// the other side/service in a peer-to-peer communications, like an RPC call.
 
 	// PeerService records the service name of the peer
-	PeerService = &stringTag{"peer.service"}
+	PeerService = stringTag("peer.service")
```
adriancole: Not related to your change.. This is sortof like the "sa" and "ca" annotations in zipkin. https://github.com/openzipkin/zipkin/blob/master/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift#L144

PR #30, opened 2016-01-06
%s/marshal/encode/g, %s/unmarshal/decode/g

bensigelman:


opentracing/codec.go
```diff
@@ -0,0 +1,71 @@
+package opentracing
```
bensigelman: (stuck this stuff in its own file)

opentracing/tracecontext.go
```diff
@@ -62,81 +62,11 @@ type TraceContext interface {
...
-	) (TraceContext, error)
-}
-
 // TraceContextSource is a long-lived interface that knows how to create a root
-// TraceContext and marshal/unmarshal any other.
+// TraceContext and encode/decode any other.
 type TraceContextSource interface {
-	TraceContextMarshaler
-	TraceContextUnmarshaler
+	TraceContextEncoder
```
bensigelman: @yurishkuro there's no formal reason why `TraceContextSource` needs to embed the codec interfaces... of course it's fine for a single type to implement all three, but any objection to removing the embedding here?

opentracing/codec.go
yurishkuro: +1

opentracing/tracecontext.go
yurishkuro: I'm ok with removing them here, but are you going to add them to the Tracer then?  Because the Tracer interface was the only place these codec methods were accessible by clients.  Actually, how would the standard tracer get a hold of these, do you need to pass them to the constructor separately?

bensigelman: At the moment, I think the only code that actually requires the encoder/decoder is the helper for HTTP headers...

Maybe the best thing would be to expose them via accessor methods (rather than embedding) on the TraceContextSource. That way we don't force users to have a single type implement the Source methods as well as the codec methods.

bensigelman: (then again, the Source could just delegate... I don't know, it's six vs half-dozen.)

bensigelman: I looked at this some more. IMO we should leave this as-is. There are arguments to split it out but I don't want to force users to keep track of a `Tracer` impl separately from the encoder and decoder; and if `Tracer` is going to provide access to the encoder+decoder, what we have here is arguably the simplest thing, and that always seems like a good place to start.

bensigelman:
Ok to merge?

opentracing/tracecontext.go
yurishkuro: +1 to keep as is, it's workable now, even in Python, with a delegation. Even though Python supports multiple inheritance, calling constructors of 3 superclasses is tricky (encoders need the header names as ctor args), so I ended up embedding them and delegating.

yurishkuro:
lgtm

PR #38, opened 2016-01-19
RFC about https://github.com/opentracing/opentracing.github.io/issues/30

bensigelman:


bensigelman:
(please provide high-level feedback at https://github.com/opentracing/opentracing.github.io/issues/30, not on this PR... that discussion is already fragmented across two Issues, I don't want to make things more confusing!)

span.go
```diff
@@ -47,6 +35,46 @@ type Span interface {
...
+	// RecordEvent() is equivalent to
+	//
+	//   Log(time.Now(), LogData{Event: event})
+	//
+	// if payload is unspecified; otherwise, only one payload argument is
+	// accepted, and RecordEvent() is equivalent to
+	//
+	//   Log(time.Now(), LogData{Event: event, Payload: payload[0]})
+	//
+	RecordEvent(event string, payload ...interface{})
```
yurishkuro: seems this would be more in line with info/error if this was called just `Event`

dankosaur: @yurishkuro +1 -- Had the same thought, based on this naming was wondering if `Event` is reserved as some sort of data structure

bensigelman: (ok, changed!)

bensigelman: (ok, changed!)

span.go
```diff
@@ -47,6 +35,46 @@ type Span interface {
...
+
+	// Info() is equivalent to
+	//
+	//   Log(time.Now(), LogData{Message: fmt.Sprint(args...)})
+	//
+	Info(args ...interface{})
+
+	// Infof() is equivalent to
+	//
+	//   Log(time.Now(), LogData{Message: fmt.Sprintf(format, args...)})
```
yurishkuro: this is a bit confusing. Seems like these logging methods do not allow structured payloads, like a json, if it's always going to be transformed into a combined message after executing sprintf. Or does this allow lazy evals?

bensigelman: Yes, the idea here (per option A) was to restrict application payload data (like a json, as you say) to either `Event("payload_identifier", appPayload)` or the catch-all `Log(time.Now(), ...)`.

I have been looking at code we have where application payloads are involved, and for the most part there's either a pretty-printed log message with format params XOR a payload; rarely both.

bensigelman: PS: I am not eager to sort out lazy-eval without reliable profiling data from multiple systems/deployments since it's basically an optimization.

PPS: Not sure it's wise to include both `Info` and `Infof` (etc)... perhaps we should just choose an idiom and stick to it.

PR #39, opened 2016-01-22
Like #38, but with less functionality

bensigelman:
See https://github.com/opentracing/opentracing-go/pull/38 and https://github.com/opentracing/opentracing.github.io/issues/30#issuecomment-173756516 .

### DIFF
--- a/span.go
+++ b/span.go
@@ -35,35 +35,6 @@ type Span interface {
 	// Suitable for serializing over the wire, etc.
 	TraceContext() TraceContext
 
-	// Log() records `data` to this Span.
-	//
-	// See LogData for semantic details.
-	Log(data LogData)
-
-	// Info() is equivalent to
-	//
-	//   Log(time.Now(), LogData{Message: fmt.Sprint(args...)})
-	//
-	Info(args ...interface{})
-
-	// Infof() is equivalent to
-	//
-	//   Log(time.Now(), LogData{Message: fmt.Sprintf(format, args...)})
-	//
-	Infof(format string, args ...interface{})
-
-	// Error() is equivalent to
-	//
-	//   Log(time.Now(), LogData{Message: fmt.Sprint(args...), Severity: ERROR})
-	//
-	Error(args ...interface{})
-
-	// Errorf() is equivalent to
-	//
-	//   Log(time.Now(), LogData{Message: fmt.Sprintf(format, args...), Severity: ERROR})
-	//
-	Errorf(format string, args ...interface{})
-
 	// Event() is equivalent to
 	//
 	//   Log(time.Now(), LogData{Event: event})
@@ -74,6 +45,11 @@ type Span interface {
 	//   Log(time.Now(), LogData{Event: event, Payload: payload[0]})
 	//
 	Event(event string, payload ...interface{})
+
+	// Log() records `data` to this Span.
+	//
+	// See LogData for semantic details.
+	Log(data LogData)
 
 	// A convenience method. Equivalent to
 	//
@@ -87,10 +63,10 @@ type Span interface {
 	AddToGoContext(goCtx context.Context) (Span, context.Context)
 }
 
-// See Span.Log(). Every LogData instance should specify at least one of
-// Message, Event, or Payload.
+// See Span.Log(). Every LogData instance should specify at least one of Event
+// and/or Payload.
 type LogData struct {
-	// The timestamp of the log event; if set to the default value (the unix
+	// The timestamp of the log record; if set to the default value (the unix
 	// epoch), implementations should use time.Now() implicitly.
 	Timestamp time.Time
 
@@ -105,15 +81,6 @@ type LogData struct {
 	// perspective.
 	Event string
 
-	// Message (if non-empty) is a free-form debugging string, much in keeping
-	// with general practices around console-oriented logging of human-readable
-	// messages.
-	Message string
-
-	// Severity will most often be non-V0 when `Message` is non-empty, but that
-	// is not a formal requirement.
-	Severity Severity
-
 	// Payload is a free-form potentially structured object which Tracer
 	// implementations may retain and record all, none, or part of.
 	//
@@ -122,12 +89,3 @@ type LogData struct {
 	// semantic flags to a Log() implementation.
 	Payload interface{}
 }
-
-type Severity int
-
-const (
-	// Loosely modeled after https://github.com/google/glog/blob/v0.3.4/src/glog/log_severity.h#L47
-	INFO    = 0
-	WARNING = 1
-	ERROR   = 2
-)
--- a/span.go
+++ b/span.go
@@ -39,12 +39,13 @@ type Span interface {
 	//
 	//   Log(time.Now(), LogData{Event: event})
 	//
-	// if payload is unspecified; otherwise, only one payload argument is
-	// accepted, and Event() is equivalent to
+	LogEvent(event string)
+
+	// EventWithPayload() is equivalent to
 	//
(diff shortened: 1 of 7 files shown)


span.go
```diff
@@ -47,6 +35,22 @@ type Span interface {
...
+	// accepted, and Event() is equivalent to
+	//
+	//   Log(time.Now(), LogData{Event: event, Payload: payload[0]})
+	//
+	Event(event string, payload ...interface{})
+
+	// Log() records `data` to this Span.
+	//
+	// See LogData for semantic details.
+	Log(data LogData)
```
bensigelman: not sure what to call this in the slimmed-down version of this API... it is important to support explicit timestamps as a boundary case, though.

span.go
```diff
@@ -58,3 +62,30 @@ type Span interface {
...
 }
+
+// See Span.Log(). Every LogData instance should specify at least one of Event
+// and/or Payload.
+type LogData struct {
+	// The timestamp of the log record; if set to the default value (the unix
+	// epoch), implementations should use time.Now() implicitly.
+	Timestamp time.Time
+
+	// Event (if non-empty) should be the stable name of some notable moment in
```
adriancole: Are we expecting empty string events? Or do you mean if this field is not set. I can understand the latter (or school me if empty is null or none in go)

span.go
```diff
@@ -58,3 +62,30 @@ type Span interface {
 	// TraceContext.
 	AddToGoContext(goCtx context.Context) (Span, context.Context)
 }
+
+// See Span.Log(). Every LogData instance should specify at least one of Event
+// and/or Payload.
+type LogData struct {
+	// The timestamp of the log record; if set to the default value (the unix
+	// epoch), implementations should use time.Now() implicitly.
+	Timestamp time.Time
```
adriancole: So here we are saying the boundary, not the unit. Maybe make docs clear if that's the case. Ex not necessarily microseconds if that's the intent. If Timestamp already includes a unit (in go), maybe it is implied.

span.go
adriancole: In languages that support overloading, id expect them to just add a timestamp arg and skip the intermediary.

span.go
```diff
@@ -58,3 +62,30 @@ type Span interface {
...
+	// perspective.
+	Event string
+
+	// Payload is a free-form potentially structured object which Tracer
+	// implementations may retain and record all, none, or part of.
+	//
+	// If included, `Payload` should be restricted to data derived from the
+	// instrumented application; in particular, it should not be used to pass
+	// semantic flags to a Log() implementation.
+	Payload interface{}
```
adriancole: Payload is interesting as we are saying this could be unused, used in process, or sent out of band somewhere. In other words "whether or not payload is used, and how it is used, is implementation specific."

We could perhaps add an example, in some footnote of a use case of this, as otherwise it feels like a junk column in a database ;) 

Ex the derived data you mention, maybe an example of an event-specific context that isn't a tag. As tags lack timestamps I could guess this could support time-correlated tags. Iirc the major call for this was more about arbitrarily complex objects. At any rate, either here or some doc someplace, I'd elaborate on this payload concept (if that hasn't already occurred)

span.go
bensigelman: `time.Time` is the canonical way to represent a moment in time in Go. It actually offers nanosecond precision (which is pretty silly given that modern CPUs *don't* offer nanosecond precision!).

FYI: https://golang.org/pkg/time/#Time

span.go
bensigelman: I mean if it's not set, basically, yeah. Go makes it extremely annoying to pass pointers to literal strings... you would want to write

```
  ...
  Event: &"my_literal_string",
  ...
```

, but that doesn't compile. This does, but is super gross:

```
func strPtr(s string) { return &s }

  ...
  Event: strPtr("my_literal_string"),
  ...
```

gah.

span.go
bensigelman: yeah, I think that makes sense... for this PR I mainly wanted to get the interface out there for discussion, but the "semantic API" page would need to include something about this.

span.go
```diff
@@ -47,6 +35,22 @@ type Span interface {
...
+	// Event() is equivalent to
+	//
+	//   Log(time.Now(), LogData{Event: event})
+	//
+	// if payload is unspecified; otherwise, only one payload argument is
+	// accepted, and Event() is equivalent to
+	//
+	//   Log(time.Now(), LogData{Event: event, Payload: payload[0]})
+	//
+	Event(event string, payload ...interface{})
```
yurishkuro: I suggest `Event` and `EventWithPayload`, instead of referring to `payload[0]`. The need to define two methods is not the fault of the API, it's the fault of the language.

span.go
```diff
@@ -58,3 +62,30 @@ type Span interface {
 	// TraceContext.
 	AddToGoContext(goCtx context.Context) (Span, context.Context)
 }
+
+// See Span.Log(). Every LogData instance should specify at least one of Event
+// and/or Payload.
+type LogData struct {
```
yurishkuro: I really think we should start a page on the `.io` site where we define the Scope of the version X of the API by listing the Use Cases and the "canonical" way of implementing them via the API. Then all API changes can be validated on whether they satisfy the established use cases.

For example, the implication of the previous `Error()` log function was that the span is explicitly marked as "failed", i.e. an indication to the UI to display the span or the log entry differently, and possibly for sampler to capture the trace.

span.go
```diff
@@ -58,3 +62,30 @@ type Span interface {
...
+	// Event (if non-empty) should be the stable name of some notable moment in
+	// the lifetime of a Span. For instance, a Span representing a browser page
+	// load might add an Event for each of the Performance.timing moments
+	// here: https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming
+	//
+	// While it is not a formal requirement, Event strings will be most useful
+	// if they are *not* unique; rather, tracing systems should be able to use
+	// them to understand how two similar Spans relate from an internal timing
+	// perspective.
+	Event string
```
yurishkuro: To clarify, are we saying that the conventional logging, with a readable message, is no longer a part of the API?

bensigelman: This is an RFC, per https://github.com/opentracing/opentracing.github.io/issues/30#issuecomment-173756516

The idea is to focus on the propagation API and the non-controversial parts of the logging API, get some experience integrating into applications and tracers, then add human-readable logging afterwards. I.e., I would see this as a strategic deferral, not a removal.

span.go
bensigelman: I think this is an excellent idea... you previously mentioned opinionated recipes for doing various common tracing-related tasks, which is a related excellent idea.

span.go
bensigelman: agreed entirely about the language aspect... annoying. your suggestion sounds fine to me, I can live with either.

span.go
yurishkuro: empty string is the default "zero" value for a string field in Go, so it's fine to treat it as "not set".

yurishkuro:
so if we change `...interface{}` to a single value, lgtm.

bensigelman:
@yurishkuro

I got rid of the `...interface{}` hack... I also did `s/Event/LogEvent/` in order to both:

1. Make the fallback `Log(...)` method naming sensible
1. Future-proof against the possible inclusion of formatted log strings (like the ones backed out earlier)

In other languages that support overloading this would/will be simpler, of course.

@adriancole I added a "for example" about payloads, btw :)

yurishkuro:
lgtm.

bensigelman:
I fixed the build and tests to accommodate this API change... I'll merge and send out a change for the docs site in a bit.

Issue #33, opened 2016-01-13
Scope of OpenTracing, scope of API repositories, RPC or Tracing


kriskowal:
We on Uber’s RPC group are embarking on a project to generalize context propagation from incoming requests to outgoing requests. We would like to start a discussion about the architecture and scope of open tracing.

@yurishkuro tells us that this undertaking is within the charter of the open tracing working group, which appears to:

“Provide a library in various languages that has common types for trace spans and contexts, and the logic for propagating incoming spans to outgoing spans. Provide a sufficiently general library that bindings for arbitrary clients, servers, and trace reporters can be developed independently and swapped as need arises.”

Instrumenting specific clients (e.g., the Node.js built-in HTTP client), servers (e.g., the built-in Go HTTP server), and trace reporters (e.g., one in Python for Zipkin) is out of scope but some will likely be contributed to the commons anyway. Wire formats for transmitting spans and annotations appear to be out of scope.

Apart from dealing with trace propagation, the tracing library provides an affordance for simple context propagation, copying attributes from incoming spans to outgoing spans. For generalized RPC context propagation, copying from incoming to outgoing requests will not be sufficient for most attributes, and undesirable for performance in other cases, as each hop would bloat fan-out requests.

- Request properties like an auth token and auth params should be copied from incoming to outgoing.
- Request context properties like “caller”, “callee”, and “method” should not be propagated, provided either at call sites of the instrumented RPC library, or inferred from the RPC IDL (in our case, Thrift). For completeness, these request properties include retry flags, speculative execution configuration, and shard keys for handle-or-forward stateful services.
- Request context properties like “timeout” need to be converted to a “deadline” on arrival, copied to outgoing request contexts, then converted back to “timeout” at the time of each attempt, bearing in mind that the IDL for the outgoing endpoint might have a shorter timeout that what remains.
- Response context properties like invalidated cache keys would need to be merged.

A context propagation library will also be in a good position to forward cancellation to the downstream call graph and perhaps even abort local work.

We have a few options for stacking this architecture. The tracing library as stands provides a tightly coupled solution for general stack propagation, but does not address cases like “timeout” propagation. The common case is that various kinds of request context need unique logic for propagation.

- We could separate the concern of trace propagation entirely from request context propagation, and use the trace library to serve one of many separable concerns (“cancellation”, “timeouts”, “auth”, “tracing”).
- We could merge all of these concerns into a single tracing library. This is attractive if these concerns are sufficiently common.
[...]

PR #40, opened 2016-01-24
RFC: far fewer types (but comparable portability)

bensigelman:
This is not for merging as-is – the `standardtracer` package is a mess, for example – but is definitely worth looking at and thinking about.

The goal here was to experiment with whether some of the formality around `TraceContext` could be removed from the API entirely. In my eyes, `TraceContext` (or `SpanContext` if we were to call it that -- whatever) is a worthwhile concept for an OpenTracing **implementation**, but I am not so sure it needs to leak out as an abstraction at the top level (as it has to date).

Also, this should be obvious: if we decide to go with this approach, it will affect all languages. I'm starting with Go because (a) Go makes it easy to think about interfaces without getting bogged down in boilerplate, and (b) there are working examples I could compile against as a sanity check.

I'll make some other comments inline.

Some people who might care about this: @yurishkuro @adriancole @michaelsembwever @slimsag @dankosaur @bcronin


README.md
```diff
@@ -64,10 +64,11 @@ If you prefer direct control to singletons, manage ownership of the
 Additionally, this example demonstrates how to get a `context.Context`
 associated with any `opentracing.Span` instance.
 
-```
+```go
     func xyz(goCtx context.Context, ...) {
         ...
-        sp, goCtx := opentracing.JoinTrace("span_name", goCtx).AddToGoContext(goCtx)
+        goCtx, sp := opentracing.ContextWithSpan(
```
bensigelman: FWIW, I was just tinkering with whether `Span` really needed that `AddToGoContext` helper method... the change here is not related to the main event in this PR.

README.md
```diff
@@ -84,8 +85,8 @@ associated with any `opentracing.Span` instance.
 
             // Transmit the span's TraceContext as an HTTP header on our
             // outbound request.
-            opentracing.AddTraceContextToHeader(
-                span.TraceContext(),
+            opentracing.PropagateSpanInHeader(
+                span,
```
bensigelman: important difference: we propagate `Span` instances, not `TraceContext`s.

README.md
```diff
@@ -98,23 +99,16 @@ associated with any `opentracing.Span` instance.
...
+```go
     http.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
-        // Grab the TraceContext from the HTTP header using the
-        // opentracing helper.
-        reqTraceCtx, err := opentracing.TraceContextFromHeader(
-                req.Header, opentracing.GlobalTracer())
-        var serverSpan opentracing.Span
-        var goCtx context.Context = ...
+		// Join the trace in the HTTP header using the opentracing helper.
+        serverSpan, err := opentracing.JoinTraceFromHeader(
```
bensigelman: important difference: we "JoinTrace..." on the receiving end of a propagated `Span`, much like the previous `JoinTrace` methods do for `Span` instances.

important difference: since there's no `TraceContext` intermediary anymore, we must provide the operation name at "de-propagation" time.

codec.go
```diff
@@ -1,71 +0,0 @@
-package opentracing
-
-// TraceContextEncoder is a simple interface to encode a TraceContext as a
-// binary byte array or a string-to-string text map.
-type TraceContextEncoder interface {
```
bensigelman: this stuff was half-moved, half-rewritten in `spanpropagation.go`.

examples/dapperish.go
```diff
@@ -19,10 +18,10 @@ import (
 func client() {
 	reader := bufio.NewReader(os.Stdin)
 	for {
-		span := opentracing.StartTrace("getInput")
-		ctx := opentracing.BackgroundGoContextWithSpan(span)
+		ctx, span := opentracing.BackgroundContextWithSpan(
```
bensigelman: this file gives a sense of how user code differs... mainly just that `TraceContext` is no longer there.

gocontext.go
```diff
@@ -2,31 +2,31 @@ package opentracing
...
-// GoContextWithSpan seturns a new `context.Context` that holds a reference to
+// ContextWithSpan returns a new `context.Context` that holds a reference to
 // the given `Span`.
 //
-// NOTE: We use the term "GoContext" to minimize confusion with TraceContext.
-func GoContextWithSpan(ctx context.Context, span Span) context.Context {
-	return context.WithValue(ctx, activeSpanKey, span)
+// The second return value is simply the `span` passed in:
+// this can save some typing and is only provided as a convenience.
+func ContextWithSpan(ctx context.Context, span Span) (context.Context, Span) {
```
bensigelman: since there's no longer a `TraceContext` type, we can finally refer to `GoContext` as `Context` :)

span.go
```diff
@@ -21,6 +24,48 @@ type Span interface {
...
+	//
+	// As such, `restrictedKey` MUST match the regular expression
+	// `(?i:[a-z0-9][-a-z0-9]*)` and is case-insensitive. That is, it must
+	// start with a letter or number, and the remaining characters must be
+	// letters, numbers, or hyphens. See CanonicalizeTraceAttributeKey(). If
+	// `restrictedKey` does not meet these criteria, SetTraceAttribute()
+	// results in undefined behavior.
+	//
+	// Returns a reference to this Span for chaining, etc.
+	SetTraceAttribute(restrictedKey, value string) Span
```
bensigelman: (now part of Span proper)

spanpropagator.go
```diff
@@ -0,0 +1,78 @@
+package opentracing
+
+// SpanPropagator is responsible (a) for encoding Span instances in a manner
```
bensigelman: this file is the most interesting/controversial piece of the PR... naming and semantics are both tricky here since nothing is symmetrical and we need to support models that create child spans across RPC boundaries and models that use a single span for both parties in an RPC.

spanpropagator.go
```diff
@@ -0,0 +1,78 @@
...
+	) (
+		traceContextID map[string]string,
+		traceAttrs map[string]string,
+	)
+
+	// JoinTraceFromBinary starts a new Span with the given `operationName`
+	// that's joined to the Span that was binary-encoded as traceContextID and
+	// traceAttrs (see SpanPropagator.PropagateSpanAsBinary()).
+	JoinTraceFromBinary(
+		operationName string,
```
bensigelman: again, note that the `operationName` is required since we need to return a `Span` instance.

standardtracer/raw.go
```diff
@@ -8,7 +8,7 @@ import (
 
 // RawSpan encapsulates all state associated with a (finished) Span.
 type RawSpan struct {
```
bensigelman: Do not waste any time reading any of the diffs under `standardtracer/`... I made the changes I needed to in order to get a working example and nothing more :)

If we want to support the `standardtracer` idiom going forward, I would suggest that `TraceContext` move into this package... it makes a lot of things cleaner for implementations even if we decide it's not needed for OpenTracing callers.

tracecontext.go
```diff
@@ -1,103 +0,0 @@
...
-// it within a larger trace tree.
-//
-// TraceContexts are sufficient to propagate the, well, *context* of a
-// particular trace between processes.
-//
-// TraceContext also support a simple string map of "trace attributes". These
-// trace attributes are special in that they are propagated *in-band*,
-// presumably alongside application data. See the documentation for
-// SetTraceAttribute() for more details and some important caveats.
-type TraceContext interface {
```
bensigelman: very important diff: this file is gone

tracer.go
```diff
@@ -26,22 +26,7 @@ type Tracer interface {
 	// Like `StartTrace`, but the returned `Span` is made a child of `parent`.
 	//
 	// The `parent` parameter can either be a `context.Context` or an
-	// `opentracing.TraceContext`. In the former case, the implementation
-	// attempts to extract an `opentracing.Span` using `SpanFromGoContext()`.
+	// `opentracing.Span`. In the former case, the implementation attempts to
+	// extract an `opentracing.Span` using `SpanFromContext()`.
 	JoinTrace(operationName string, parent interface{}) Span
```
yurishkuro: looks like we don't need this method anymore. If parent is a context, devs can do
```go
SpanFromContext(context).StartChild(...)
```
and if it's already a span then just call `span.StartChild()`

spanpropagator.go
```diff
@@ -0,0 +1,78 @@
...
+	//
+	// The first return value must represent the SpanPropagator's encoding of
+	// the core identifying information in `sp`.
+	//
+	// The second return value must represent the SpanPropagator's encoding of
+	// trace attributes, per `Span.SetTraceAttribute`.
+	PropagateSpanAsBinary(
+		sp Span,
+	) (
+		traceContextID []byte,
```
yurishkuro: to bury the notion of "trace context" completely, we may want to refer to this argument as `spanID`

bensigelman: oh yeah, that was just an oversight. for sure.

tracer.go
bensigelman: good point... will do if we move forward with this.

yurishkuro:
@bensigelman good job! I like this direction, makes the API simpler to use, and solves the "contains" vs. "identified by" conundrum.

There is one edge case that now won't work (the "debug" trace), but it wasn't very elegant in the previous version either, so we can think about it separately. Maybe once we extend the API to allow externally provided timestamps, the debug use case will be solved there as well.

+1

bensigelman:
@yurishkuro, would the debug case be satisfied by a tag map made available at Span initialization time? Agreed that such a thing could probably be folded in to the forthcoming API for after-the-fact span recording...

yurishkuro:
> would the debug case be satisfied by a tag map made available at Span initialization time?

It could be, but as we've seen it also makes for an ugly api (in some languages), and the use case for pre-construction tags is pretty weak, so I wouldn't rush there.

README.md
adriancole: I'd expect this hook to be most commonly called by integrators, as tracers often bundle an http library and/or plugins that start or resume a trace. Some of these set the operation name while reading from headers, but in many cases they don't because operation name can be changed, specified elsewhere or later.

PS headers in zipkin include trace_id, id, parent id, sampled and (while rarely implemented) the debug flag. Brave has an optional way to send the span name in a header iirc.

examples/dapperish.go
```diff
@@ -19,10 +18,10 @@ import (
...
+			opentracing.StartTrace("getInput"))
 		// Make sure that global trace tag propagation works.
-		span.TraceContext().SetTraceAttribute("User", os.Getenv("USER"))
+		span.SetTraceAttribute("User", os.Getenv("USER"))
```
adriancole: don't mean to distract the issue, but I will :package: 

I *really* like Tag Sets where propagation is simply a property of a tag. The "Trace" part of the name "TraceAttribute" has resulted in heaps of discussion, notably as no-one can guess by its name that it a propagation tag, but also the common "is it a part of the trace?" Ex. in zipkin trace id, id, parent id, and sampled decision are "trace attributes", ie only one field has anything to do with the trace. 

I really would rather we be able to deconflate this whole thing by saying `Span.propagateTag("User", os.Getenv("USER"))`, or worse but still better `Span.propagateAttribute("User", os.Getenv("USER"))`. I have consistently found TraceAttribute an unnecessarily distracting term.

examples/dapperish.go
```diff
@@ -36,8 +35,8 @@ func client() {
 
 		httpClient := &http.Client{}
 		httpReq, _ := http.NewRequest("POST", "http://localhost:8080/", bytes.NewReader([]byte(text)))
-		opentracing.AddTraceContextToHeader(
-			span.TraceContext(), httpReq.Header, opentracing.GlobalTracer())
+		opentracing.PropagateSpanInHeader(
```
adriancole: Existing tracers send via multiple headers, but I guess this is named PropagateSpanInHeader vs PropagateSpanInHeaders bc in go, headers are in an object named Header?

gocontext.go
adriancole: hah

span.go
adriancole: I do like like that we don't need to think about context as a user of a span.

For zipkin users to propagate debug, they'd need to access something like this directly or indirectly.

We probably need to be a bit less chatty with docs, or move them somewhere else, since this is now on a user type. For example, users should be more encouraged about baggage that isn't a trace identifier (eventhough that is supported).

span.go
```diff
@@ -36,25 +81,29 @@ type Span interface {
...
-	// Sets the end timestamp and calls the `Recorder`s RecordSpan()
-	// internally.
-	//
-	// Finish() should be the last call made to any span instance, and to do
-	// otherwise leads to undefined behavior.
-	Finish()
+// Tags are a generic map from an arbitrary string key to an opaque value type.
+// The underlying tracing system is responsible for interpreting and
+// serializing the values.
+type Tags map[string]interface{}
```
adriancole: I'm just ignoring this part.. ping me if it is not go-specific

spanpropagator.go
adriancole: requiredAttrs? bc if the intent is for these not to be dropped, but also fixed length, at least zipkin expects the same few attrs propagated. trace id, span id, parent id, and sampled. While we can argue about whether parent id can be sent, in reality the collectors make no attempt to look up a parent id when missing, and lack of parent id assumes the span is a root span. if we don't propagate sampled, we can royally muck up span trees as. 

Since we describe this as "core identifying information", implementations should be able to decide what that is. Ex. "user-id" certainly isn't, and even if some may not like what zipkin has as core propagation tags, they are crucial for bug prevention.

spanpropagator.go
adriancole: this is because we've made the operationName un-settable. In zipkin, it can be set late or by a different component that the one handling the transport.

spanpropagator.go
```diff
@@ -0,0 +1,78 @@
...
+	// JoinTraceFromBinary starts a new Span with the given `operationName`
+	// that's joined to the Span that was text-encoded as traceContextID and
+	// traceAttrs (see SpanPropagator.PropagateSpanAsBinary()).
+	//
+	// It's permissible to pass the same map to both parameters (e.g., an HTTP
+	// request headers map): the implementation should only decode the subset
+	// of keys it's interested in.
+	JoinTraceFromText(
+		operationName string,
+		traceContextID map[string]string,
```
adriancole: same comment, but less cluttered place to suggest an alternative: coreAttrs, userAttrs

tracer.go
adriancole: I need to catch up on the span.startChild topic, although here probably isn't a great place to do that.

examples/dapperish.go
michaelsembwever: Yes.
Or 

    Span.setCascadedTag("User", os.Getenv("USER"))

It's an attribute that child tags will also inherit, regardless of network/process boundaries.

I find the whole "Distributed Context Propagation" at the API level, has grown into an unneeded (psychological) [schema](https://en.wikipedia.org/wiki/Schema_%28psychology%29).

To the user of the API it can be simplified just as an attribute the casades, even over the network/process boundaries.

( The "Distributed Context Propagation" schema is still valuable in Specification though. )

adriancole:
I like that we are calling this a propagation api vs a codec.. that's a better choice. There are some implementation glitches that this doesn't cause, but maybe just highlights better.

For example, span operationName often lives in code that doesn't have the responsibility of reading headers or thrift envelopes.

Also, I've made some comments that we should really say "core" attributes/propagated tags vs assuming they will be identifiers. Here are the core attributes in zipkin. Note that optional below doesn't mean it is optional to propagate. It means that it can be null.

```
  1: i64 trace_id,
  2: i64  span_id,
  3: optional i64 parent_span_id,
  5: optional bool sampled
```

https://github.com/twitter/finagle/blob/develop/finagle-thrift/src/main/thrift/tracing.thrift#L246

span.go
michaelsembwever: See the discussion (further up)[https://github.com/opentracing/opentracing-go/pull/40/files#r50631798] asking for better naming of the method `SetTraceAttribute(..)`.

tracer.go
michaelsembwever: See https://github.com/opentracing/opentracing-java/pull/7/files#r50501336

Or https://github.com/opentracing/opentracing.github.io/pull/34#discussion_r50616849 
Or https://github.com/opentracing/opentracing.github.io/pull/34#discussion_r50616728

adriancole:
Another late comment. It is probably acknowledged, but this api is what binds us or overlaps with other context apis. Ex. in finagle `HttpContext.read` is used to get a context which can also be used for ancillary things to tracing, such as deadlines. Maybe use-cases doc needs to enumerate that we don't expect to replace context apis with this, rather interact with them.

adriancole:
Suggest we just contain the zipkin debug flag to finagle as it is rarely
implemented outside it. I don't mind opening an issue to help change to a
simpler policy of always reading the sampled flag.

debug is very complex compared to sampled=0 or 1, and only really impacts
the collection tier, which we assume is down-sampling.

I made notes about this
https://github.com/opentracing/opentracing.github.io/pull/34#discussion_r50632814

spanpropagator.go
yurishkuro: I have no objections to late binding of operation name

examples/dapperish.go
bensigelman: I, too, am unnerved by the naming around "Trace Attributes"... if we pursue the basic model of this PR, something like `SetCascadedTag` is worth considering seriously (since both the cascaded and "non-cascaded" tags apply to the same object now: the Span).

@adriancole @michaelsembwever @yurishkuro thoughts on something like the following three methods on Span?

```
    SetTag(string, BasicType)
    SetCascadingTag(string, BasicType)
    Tag(string) --> string
```

Some problems right off the bat:
- If "cascading tags" (trace attributes) are to be used as HTTP headers without bizarro escaping, we run headlong into the lengthy/complex caveats about casing, hyphens, etc. And it seems like the key restrictions for "plain" tags and cascading tags should be identical.
- What to do if there's a "cascading" tag and a user calls `SetTag` with the same key? Who wins?

examples/dapperish.go
bensigelman: yeah, I don't like it, but that's what the Go maintainers decided, apparently.

span.go
bensigelman: (@adriancole nothing to see here...)

spanpropagator.go
bensigelman: @adriancole @yurishkuro it's more than `requiredAttrs`... there are other "required" parts of a Span that don't get sent in-band during propagation (the operation name, the timing info, etc).

Maybe we should use language consistent with whatever we end up calling `JoinTraceFrom(Binary|Text)`. It is borderline tautological, but the truest way to describe this field is that it's the minimum amount of state needed to join back to the trace on the other side of the propagation boundary.

spanpropagator.go
bensigelman: Yeah, me neither... I even have a few (unrelated) scenarios where I think it would be needed.

In Go this presents some annoying problems since overloading is disallowed by the compiler. @yurishkuro thoughts? I believe this would affect everywhere that an operation name is required at Span creation time. My personal preference would be to create a single, less convenient API for starting spans (much like that [logging RFC](https://github.com/opentracing/opentracing-go/pull/39) did with `LogData`) that handles all of the cases via a large `Options`-like struct.

bensigelman:
Thanks, all, for the comments so far. Here are the important outstanding issues as I see them... in no particular order:
- what to call and/or change about `SetTraceAttribute` (leave as-is? rename? combine with tags? what about the key restrictions?)
- how to refer to the parts of a Span that need to get sent in-band during propagation
- how we should allow people to set operation names after Spans have already started
- whether/how to change method names around starting and joining traces

One other nice side-effect of this proposal: the hacks where we returned a child context *and* tags for the forthcoming Span are gone; even better, that Span does not even need to represent out-of-band identity fields like `parent_id` in the Span Tags.

Finally: nobody has leapt to the defense of TraceContext at this level of the API... seems like it's a goner.

spanpropagator.go
yurishkuro: I myself haven't run into use cases where operation name was not know at the point of starting the span. Interested to hear what they are. But as far as the API, if we add an extra `setOperationName` method to the span, then the `JoinTraceFromBinary` doesn't need to change, people can just provide an empty string for the name. I do want to have `xxxWithOptions` methods for handling more exotic cases like external timestamps, but that approach still binds everything at the span creation.

adriancole: >
> I myself haven't run into use cases where operation name was not know at
> the point of starting the span. Interested to hear what they are.
>
I think there's a difference between use-cases and interop use-cases.

Ex. in finagle the operation name is an event (like most things). In brave
there's other code that's responsible for setting the span name. Either
would need to break to ensure the responsibility of setting the name of an
operation was at span creation time.

WRT to the question of why systems including above were designed to permit
the name to be late-bound (use cases)? Well, it would probably take some
hours to research this.

yurishkuro: another possibility is to simply drop operation_name as a requirement for a Span and delegate it to be a "standard" tag. Solves both ugly-api and late-binding problems.

dankosaur: ^ I like this.  It is parallel to other tags that will be recommended as part of the semantic span API.  However, this is important enough that it should be required and have an option to provide in constructor.

examples/dapperish.go
dankosaur: :+1: -- `Attribute` vs `Tag` was a strange terminology distinction when I first started getting into this project; nothing unusual but the kind of naming that software is unfortunately rife with.  So I think it can be clearer.

However, I'm not sure folding `Attributes` into `Span Tags` (via `CascadingTag` or similar) is going to solve the problem.  A `Tag` is data added to a span, whereas an `Attribute` is data added to a `Context`.  Based on that `Attributes` feel like they are not part of `Span` creation.  We do want to propagate them, but they are not `Span`-centric.

It makes sense to think of the relevant part of `DCP` as propagating a `Span` when you have a zipkin-like model where client and server report data into the same `Span`, but when considering more generalized tracing systems, `Context` is the thing that is propagated, and `Span`s are created to report information on execution as the `Context` passes through.  We can add `Span` info to the `Context` before propagating to inform the structure of the tree and `Span`s, but that doesn't make the `Context` a `Span`.

I realize this is arguably scope question currently being debated in #33 as well, but clarity is the goal of this line of discussion and I think it will be clearer if we don't conflate these two concepts.

bensigelman:
Per https://github.com/opentracing/opentracing.github.io/issues/24, I cleaned up the incredibly gross parts of this (i.e., `standardtracer`), fixed some comments, fixed tests, and would now like to merge this.

I recognize that we are not done talking about `SetTraceAttribute`, but I also think that can be resolved in parallel.

I'll add a few comments inline...

examples/dapperish.go
bensigelman: (/me files a motion to decouple this important, unfinished discussion from the current PR)

spanpropagator.go
bensigelman: (I went with

```
    contextSnapshot []byte,
    traceAttrs []byte,
```

for now... happy to take suggestions!

standardtracer/spanpropagator.go
```diff
@@ -0,0 +1,194 @@
...
+	err = binary.Write(contextBuf, binary.BigEndian, sc.TraceID)
+	if err != nil {
+		panic(err)
+	}
+	err = binary.Write(contextBuf, binary.BigEndian, sc.SpanID)
+	if err != nil {
+		panic(err)
+	}
+
+	// Handle the attributes.
```
bensigelman: FYI, I haven't tested this yet and so it probably has some subtle bug. I'll write a unittest once we agree that trace attrs are here to stay. :)

standardtracer/standardcontext.go
```diff
@@ -0,0 +1,48 @@
+package standardtracer
+
+import "sync"
+
+type StandardContext struct {
```
bensigelman: This is low-importance since the whole `standardtracer` thing is just a convenience for tracing system authors, but FWIW this version of `standardtracer` is way less flexible since it hardcodes this Dapper-like `StandardContext` struct. We can push the old `TraceContext` in here and make everything way more generic if there's a demand. I didn't think it was worth the time and effort at this point in time.

standardtracer/recorder.go
```diff
@@ -1,32 +1,9 @@
 package standardtracer
 
-// ProcessIdentifier is a thin interface that guarantees all implementors
-// represent a ProcessName and accepts arbitrary process-level tag assignment
-// (e.g., build numbers, platforms, hostnames, etc).
-type ProcessIdentifier interface {
```
bensigelman: I removed this since a tracing impl can add this if it feels like it, but ProcessIdentifier has no bearing on RawSpan or its direct dependencies so thus can be omitted here.

standardtracer/standardcontext.go
yurishkuro: @bensigelman do you think it's even worth having a standard tracer?  We haven't done it in other languages.

tracer.go
dawallin: I don't like the startChild function being on the Span. It mixes Span creation with behavior. I think it is better to collect all creation functions into the Tracer.

tracer.go
```diff
@@ -22,26 +23,4 @@ type Tracer interface {
 	//         SetTag("lucky_number", 42)
 	//
 	StartTrace(operationName string) Span
-
-	// Like `StartTrace`, but the returned `Span` is made a child of `parent`.
-	//
-	// The `parent` parameter can either be a `context.Context` or an
-	// `opentracing.TraceContext`. In the former case, the implementation
-	// attempts to extract an `opentracing.Span` using `SpanFromGoContext()`.
-	JoinTrace(operationName string, parent interface{}) Span
```
dawallin: I agree that the JoinTrace function is maybe not needed any more. But I think all creation functions for a Span should be collected here in the Tracer. To have the StartChild function in the Span mixes the creation and behavior responsibility of the Span. I vote to move the StartChild into here, e.g. StartTrace(operationName, Span).

bensigelman: this is not a rhetorical question, so apologies if it sounds snarky (it's not intended to): how much time have you spent looking at calling code for tracing systems?

My experience is that the vast majority (95%? 99%?) of it would only have to use the `Span` API we have now (and, in Go, perhaps the stuff in gocontext.go), and I like that simplicity for the user's mental model.

standardtracer/standardcontext.go
bensigelman: I'm starting to doubt it :)

I can move all of this to the `dapperish` package if that's better, though I'd prefer to make that change in a separate PR regardless.

bensigelman: (I *do* think there's value in a util package that various tracer impls can reuse...)

bensigelman:
(this is now rebased against https://github.com/opentracing/opentracing-go/pull/39)

bensigelman:
Per https://github.com/opentracing/opentracing-go/pull/40#issuecomment-174408383 , I think this is basically ready for merge unless someone wants to object. There are still some open issues, but they were not created by this PR (the trace attribute questions and whether we can consolidate span-creation mechanisms).

dkuebric:
I think the biggest change here, as you point out, is the removal of `TraceContext` as an external interface.  It's been seen as a stumbling block, but I think removing it introduces a strange confusion as well--a `Span` is a reportable observation representing execution within my application, but also I need to propagate it in order to report something somewhere else?

We currently have this definition of a `Span`, which seems good and clear:
> A Span represents a logical unit of work in the system that has a start time and a duration.

Our definition of `TraceContext` is currently pretty complex, so I won't quote it here, but I think it boils down to:
> A TraceContext provides identifying information that is associated with each Span reported, tying the entire trace together.  It is propagated throughout the request, allowing it to carry not only Span identifying information, but also other Attributes added by an OT user.

Curious if anyone else shares this feeling (or if I'm an outlier)?

(PS -- I think this is pretty relevant to the discussion on #33 as well.)

bensigelman:
Well, there is `SpanPropagator` which is responsible for getting spans across process boundaries... so in a sense it takes the place of `TraceContext` and the `Codec` and `TraceContextSource` pieces, all of which were only practically useful as (a) a place to stash the baggage / trace attrs, and (b) encapsulate the parts of a Span that propagate over the wire.

But my larger point is that there's still a type in this revised API that concerns itself with propagation. It's just not an encapsulated member of the Span type.

bensigelman:
@dkuebric do you want to veto this or are you more "thinking out loud"?

Are there other dissenters to the core idea of removing TraceContext from the API?

If not, does someone want to LGTM this?

yurishkuro:
It LGTM - other discussions can continue, but this doesn't make things worse imo, yet leaves less things to learn/worry about.

PR #41, opened 2016-02-03
Use struct{} instead of int as context key

tschottdorf:
The line

```go
  val := ctx.Value(activeSpanKey)
```

previously allocated to pass the integer
`activeSpanKey` as an interface. With this
change, no such allocation is necessary.

This showed up during benchmarks in
cockroachdb/cockroach.

### DIFF
--- a/gocontext.go
+++ b/gocontext.go
@@ -2,9 +2,9 @@ package opentracing
 
 import "golang.org/x/net/context"
 
-type contextKey int
+type contextKey struct{}
 
-const activeSpanKey contextKey = iota
+var activeSpanKey = contextKey{}
 
 // ContextWithSpan returns a new `context.Context` that holds a reference to
 // the given `Span`.


gocontext.go
```diff
@@ -2,9 +2,9 @@ package opentracing
 
 import "golang.org/x/net/context"
 
-type contextKey int
+type contextKey struct{}
 
-const activeSpanKey contextKey = iota
+var activeSpanKey contextKey = struct{}{}
```
yurishkuro: this should be 

```go
var activeSpanKey = contextKey{}
```

tschottdorf: done.

yurishkuro:
Last request - could you please retitle/rewrite commit message per http://chris.beams.io/posts/git-commit/? Along the lines "Use struct{} instead of int as context key"

tschottdorf:
Sure, done. Consider adding a `CONTRIBUTING.md` or you're likely going to make that last request repeatedly for new contributors.

PR #42, opened 2016-02-03
Fix multiple bugs in standard SpanPropagator

tschottdorf:
Fixed the following bugs:

- JoinTraceFromBinary reconstructed, but didn't use the
  attributes map
- JoinTraceFromBinary contained calls to binary.Read which
  attempted to read into type `int`, which is a runtime panic.
- JoinTraceFromText did not create a new span, but returned a
  Span with identical SpanID and no parent instead.

Added a test that exercises all of the above.

Two further commits contain unrelated minor fixes.

### DIFF
--- a/standardtracer/spanpropagator.go
+++ b/standardtracer/spanpropagator.go
@@ -178,11 +178,6 @@ func (s *tracerImpl) JoinTraceFromText(
 		return nil, fmt.Errorf("Only found %v of 3 required fields", requiredFieldCount)
 	}
 
-	lcAttrsMap := make(map[string]string, len(attrsMap))
-	for k, v := range attrsMap {
-		lcAttrsMap[strings.ToLower(k)] = v
-	}
-
 	return s.startSpanGeneric(
 			operationName,
 			&StandardContext{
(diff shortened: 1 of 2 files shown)


standardtracer/spanpropagator.go
```diff
@@ -135,7 +136,7 @@ func (s *tracerImpl) JoinTraceFromBinary(
 				SpanID:       randomID(),
 				ParentSpanID: propagatedSpanID,
 				Sampled:      sampledByte != 0,
-				traceAttrs:   make(map[string]string),
```
bensigelman: yikes -- obviously this code was never tested (I was more trying to convince myself that this stuff could be written sanely).

standardtracer/spanpropagator_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"testing"
+	"time"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/standardtracer"
+	"github.com/opentracing/opentracing-go/testutils"
+)
+
+func TestSpanPropagator(t *testing.T) {
```
bensigelman: you are a champion, thank you.

bensigelman:
LGTM, thank you again.

PR #43, opened 2016-02-03
Export NoopTracer

tschottdorf:
It is used as the default value for the
package-level singleton, but allowing
external packages to make use of it is
a benefit.

Minor other code changes to have passing
`go vet ./...` and `golint`.

Any plans for CI on this repo? The current
state of affairs is somewhat sad.

### DIFF
--- a/globaltracer.go
+++ b/globaltracer.go
@@ -1,7 +1,7 @@
 package opentracing
 
 var (
-	globalTracer Tracer = noopTracer{}
+	globalTracer Tracer = NoopTracer{}
 )
 
 // InitGlobalTracer sets the [singleton] opentracing.Tracer returned by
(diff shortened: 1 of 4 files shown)


bensigelman:
Per your question, the individual platform repos mainly exist in order to verify that the (platform-independent) semantic spec makes sense. Once we get there, we'll call it "OpenTracing 1.0RC" and work on solidifying these repos / adding CI / integrating with users / integrating with more implementations. There will presumably be some light churn during that period, we'll make changes, and v1.0 will be the result.

So, TL;DR, CI and such things will happen soon but we were planning to resolve the semantic API questions (like the one you just raised on Gitter) first.

bensigelman:
And, LGTM.

PR #44, opened 2016-02-03
Fixes (span).Sampled propagation bug in StandardTracer

bg451:
The sampled status of a span wasn't being recorded in PropagateSpanAsBinary, so I went ahead and added that even though I remember seeing discussion that StandardTracer might be removed.


standardtracer/spanpropagator.go
```diff
@@ -93,6 +104,10 @@ func (s *tracerImpl) JoinTraceFromBinary(
 	if err != nil {
 		return nil, err
 	}
+	err = binary.Read(contextReader, binary.BigEndian, &sampledByte)
```
bensigelman: details, details...

PR #45, opened 2016-02-04
Fix the fix in #44

tschottdorf:
namely setting Sampled on the child whenever
it was **not** set on the parent.

Caught by the test from #42. Much can be won
by a simple one-liner setup on Travis/CircleCI,
pre-stabilization or not. The current situation
is frankly a little insane.

Happy to set that up, just need the corresponding
rights on the repository.


yurishkuro:
@tschottdorf I added you to the repo.

yurishkuro:
fix looks good.

bg451:
Oops I missed that, thanks for catching it!

PR #48, opened 2016-02-06
RFC for Injection/Extraction in Go

bensigelman:
C.f. https://github.com/opentracing/opentracing.github.io/pull/45 and https://github.com/opentracing/opentracing.github.io/issues/44#issuecomment-179654948.

There are some good things and some less-good things here... I will add a few comments inline.

I didn't adjust the standardtracer, noop.go, etc: just trying to illustrate the user-facing pieces of this.


bensigelman:
(@michaelsembwever @yurishkuro @tschottdorf @dkuebric I think all of you have taken an interest in this)

span.go
```diff
@@ -52,6 +53,10 @@ type Span interface {
 	// See LogData for semantic details.
 	Log(data LogData)
 
+	// Return a WireInjector for the given WireEncoding, or nil if the Span
+	// does not support such an encoding.
+	WireInjectorForEncoding(enc WireEncoding) WireInjector
```
bensigelman: Ultimately I thought this was better on `Span`. First of all, there's the convenience aspect: the user *definitely* has a `Span` on hand at injection time, and only possibly has a `Tracer` on hand, esp given how annoying Go can be regarding `net.Context` and so forth (and my desire not to paint anyone into a corner with singletons).

Second of all, per Gitter chatting with @tschottdorf, it's quite possible that there will be different concrete impls of `Span` for different sampling use-cases in an OT implementation, so putting the Injector access point here allows these different impls to return respectively different injectors as needed.

span.go
```diff
@@ -145,3 +150,13 @@ func CanonicalizeTraceAttributeKey(key string) (string, bool) {
 	}
 	return strings.ToLower(key), true
 }
+
+// InjectSpan is a helper that takes some of the stuttering out of the common
+// Span.WireInjectorForEncoding calling pattern.
+func InjectSpan(sp Span, enc WireEncoding, carrier ...interface{}) error {
```
bensigelman: (just a convenience to reduce stuttering)

tracer.go
```diff
@@ -23,4 +21,8 @@ type Tracer interface {
 	//         SetTag("lucky_number", 42)
 	//
 	StartTrace(operationName string) Span
+
+	// Return a WireExtractor for the given WireEncoding, or nil if the Tracer
+	// does not support such an encoding.
+	WireExtractorForEncoding(enc WireEncoding) WireExtractor
```
bensigelman: extraction, of course, needs to be on the Tracer, not on the Span (since there *is* no Span at this point)

utils.go
```diff
@@ -16,18 +17,20 @@ const (
...
+func InjectSpanInHeader(
 	sp Span,
 	h http.Header,
-	propagator SpanPropagator,
-) {
-	contextIDMap, tagsMap := propagator.PropagateSpanAsText(sp)
-	for headerSuffix, val := range contextIDMap {
+) error {
+	var traceState, attrs map[string]string
+	if err := InjectSpan(sp, WIRE_ENCODING_SPLIT_TEXT, &traceState, &attrs); err != nil {
```
bensigelman: ... just a simple example of how injection would look for the common-case of HTTP headers. See the extraction complement below.

wire.go
```diff
@@ -0,0 +1,68 @@
...
+	//   1) The "tracer state" for the Span (for example, in Dapper this would
+	//      include a trace_id, a span_id, and a bitmask representing the
+	//      sampling status for the given trace)
+	//   2) Any Trace Attributes (per Span.SetTraceAttribute)
+	//
+	// The encoded data is separated in this way for a variety of reasons; the
+	// most important is to give OpenTracing users a portable way to opt out of
+	// Trace Attribute propagation entirely if they deem it a stability risk.
+	//
+	// The `carrier` for injection must be two pointers to `[]byte`s for the
```
bensigelman: I really dislike how this RFC requires such subtle documentation of the `carrier` params (as I've called them here), esp since the compiler won't be able to check any of this for the user. Doubly bad since we're dealing with output parameters / pointers-to-slices.

wire.go
```diff
@@ -0,0 +1,68 @@
...
+// queues and other IPC mechanisms are also reasonable places to use a
+// WireInjector.
+//
+// The specific encoding for an injected Span depends on the WireEncoding.
+// OpenTracing defines a common set of WireEncodings, and each has an expected
+// carrier type and format. See the WireEncoding enum comments for details.
+//
+// See WireExtractor and Span.WireInjectorForEncoding.
+type WireInjector interface {
+	InjectSpan(span Span, carrier ...interface{})
```
bensigelman: the simplicity/elegance here is a partial counterbalance to my unhappiness with https://github.com/opentracing/opentracing-go/pull/48/files#diff-7e1c728cab70a6a846841fa62fd6e430R18

wire.go
```diff
@@ -0,0 +1,68 @@
+package opentracing
+
+// WireEncoding identifies a WireInjector/WireExtractor pair.
+type WireEncoding int
+
+const (
```
bensigelman: Note that OpenTracing-portable encodings will still need to be declared here if we do this as an enum.

Another approach – probably a *better* approach – would be to follow the `net.Context` approach which allows for a kind of "global type-namespace" of encodings without forcing anyone to deal with `package reflect`. Example: https://github.com/golang/blog/blob/master/content/context/userip/userip.go#L34

bensigelman:
My net feelings on this:
- definitely subtle, but also definitely something that vanilla OT users wouldn't need to see
- if we use more of a `net.Context` pattern, there's a little more hope for extensibility... e.g., Thrift or TChannel could (remotely) declare a WireEncoding-compatibly "key" and OT implementations could implement or not implement those encodings without `opentracing-go` enumerating everything. Hm, maybe I should mock this up for clarity. (Could also be done with plain-old strings or whatever... doesn't need to use the type system)
- the subtlety around `carrier` output params will bite someone... but in my view this is still a net win
- FWIW, having written it down both ways I prefer accessing the Injector via Span rather than Tracer

wire.go
yurishkuro: One way to provide better type safety is to expect the `carrier` to be an interface defined in OT, as I showed here: https://github.com/opentracing/opentracing.github.io/pull/45#discussion_r51581185

bensigelman: @yurishkuro so it would be a single `carrier interface{}` as opposed to a vararg, and we'd have a convention that the actual param type is an OT interface?

yurishkuro: that would be my preference, as it makes the API for the end user compiler-enforced, and still keeps the mechanism extensible (I didn't even notice that you had `...` there). For common cases like two-maps, one map, binary, etc. we can provide standard factory methods in the utils.

One drawback of the interface approach is that it requires new object allocation, but it can be address via object pool.

tschottdorf:
The API feels fairly awkward. It's simply trying to be general to the point of
fixing nothing - you don't want to commit to a certain type of Span, and you
don't want to commit to a wire format. At the end of the day, the API doesn't
express the obvious: If you write a new Span, then nobody can tell you how it's
going to {,de}serialize from/to different formats.

I think a good reality check is thinking about whether anyone would actually
want to use that API or just ignore it and use their own concrete types. A
decent interface is great, but when you're talking about serialization, it
tends to get in the way.

What speaks against taking an opinionated approach to what a Span is and
letting those folks who need something different write their own? I get
referenced for wanting my own custom Span impl, but I don't really want that -
I just didn't want to fork standardtracer just yet in anticipation of code
changes useful for not only me.

The radical approach would be simply tossing `opentracing-go` but for the
`standardtracer` package - making everything concrete. In particular, give
access to everything that needs to be serialized - the various IDs, and the
attributes map - default serializations can be provided on the package level,
and everything custom can access the fields required. That makes
`standardtracer` the implementation of the spec, as opposed to the
implementation of the interface trying to implement the spec which it is now.

I'm curious to hear about some concrete use cases which you're looking to cover
here, and how the current interface would work for them.

wire.go
michaelsembwever: "Wire" isn't helpful terminology here, all you need to care about is that this is injecting and extracting a span into another object.

Whether that object is going to a wire or not, is irrelevant.
Take for example an Extractor implementation that is extracting Spans line-by-line when parsing a log file.

michaelsembwever:
> FWIW, having written it down both ways I prefer accessing the Injector via Span rather than Tracer

Apart from finding it confusing when some functional methods are in Tracer and others are in Span, i find it breaks object-orientation. The tracer class is your controller. Here you're not performing something on the Span, but performing something on some other object (eg the request header map), so it makes more sense to have it in the Tracer than in the Span.

This is the same argument as `Span.create_child`, and there you see it clearly design-wise because it naturally collapses into the `Tracer.create_span(optional<span> parent)` method and you'll left with one less method in the API the end-user has to get into their head to understand how everything works together.

span.go
michaelsembwever: on the other hand, the more parameters you have in one method the harder it is for the new-comer to understand what is actually going on. Especially when it works one way here but another way over there. 

That and it's more porcelain api that can get added once the plumbing is working.

michaelsembwever:
> At the end of the day, the API doesn't express the obvious: If you write a new Span, then nobody can tell you how it's going to {,de}serialize from/to different formats.

+1


> if we use more of a net.Context pattern, there's a little more hope for extensibility... e.g., thrift or TChannel could (remotely) declare a WireEncoding-compatibly "key" and OT implementations could implement or not implement those encodings without opentracing-go enumerating everything. Hm, maybe I should mock this up for clarity. (Could also be done with plain-old strings or whatever... doesn't need to use the type system)

The non-extendable WireEncoding doesn't make sense, and really doesn't belong in the API. The only thing the user cares about is the format, or the framework, that the Span needs to transform to/from. And it needs to be pluggable.

The original Extractor/Injector proposal kept this stuff hidden to the Tracer, and the "key" was simply the object you needed the transformation from/to. eg `tracer.getExtractor(request.class).extract(request)` 

For the user's part this kept the psychological schema they had to bear to the bare minimum. The Tracer implementation had privately a registry of extractors/injectors keyed by corresponding object. For languages that don't support types this way I'm sure it would be still clean enough to do it string-based like `tracer.getExtractor("javax.servlet.Request").extract(request)` 

And finally, in response to the stuttering effect, I'm actually in favour of it, as used properly it makes the API and how it works easier for outsiders to understand.

yurishkuro:
> you don't want to commit to a certain type of Span, and you don't want to commit to a wire format. 

That is the intention, because API cannot commit to either:
  * API does not know the internals of the Span, how implementation chose to represent the state
  * APi does not know the target "wire" format yet expected to support all of them
  * The only thing API *can* expose is the intermediate representation of the Span as either a kv-map or a binary blob, but it has to do it in a way that remains flexible w.r.t. the first two points. The intermediate `Writer` APIs (https://github.com/opentracing/opentracing.github.io/pull/45#discussion_r51581185) are able to do that, but they are not a requirement to use the injector in a more abstract way as @michaelsembwever wants

> At the end of the day, the API doesn't express the obvious: If you write a new Span, then nobody can tell you how it's going to {,de}serialize from/to different formats.

That is by design - if you write a new Span, only *you* know how to serialize it.

yurishkuro:
```
tracer.getExtractor(request.class).extract(request)
tracer.getExtractor("javax.servlet.Request").extract(request)
```

It's language specific, of the 4 languages we're targeting initially, only Java can use the first form, JS, Python, and Go can only use a constant. I do prefer string over enum, as it keeps it extensible while still allowing OT to define some common constants.

But I cannot get behind the specific example of using a class like `javax.servlet.Request`. There are plenty of http frameworks in Java that do not use `javax.servlet.Request`, but some other type. An OT implementation is the only one that knows how to convert its Span into some other representation, but it needs to know what that representation is ahead of time. If I implement a new http framework that uses a custom class to represent the Request, there must be a way in OpenTracing to still use the available injectors to inject the span into my custom Request class, until the time my framework becomes so popular that my Request class is supported natively by all tracers. And the only way to provide the *portable* injectors is to make assumptions about the details of the serialized representation and provide a common-denominator API, like the kv-map + binary that we had originally.

bensigelman:
@tschottdorf there are two areas where it's most important for OpenTracing to remain portable across (OpenTracing) implementations: (1) vanilla timing/etc Span instrumentation of application and library code, and (2) injection/extraction of the trace+span state across process boundaries. You as a (very) early user of OpenTracing are looking at `standardtracer`, but that is not the intended pattern or use case for OT as a whole. (You can glance at some motivating use cases here: http://opentracing.io/use-cases/)

Anyway, @yurishkuro spoke well to the implications of all of the above.

@michaelsembwever I strongly agree with @yurishkuro that we shouldn't expect that OpenTracing implementors write code for each web/RPC/etc framework in popular use. This is of course the motivation for using general-purpose encodings like string maps and blobs.

Let's step back... ignoring naming and whether we use strings or type introspection or enums (and so on), the **real** question here is where the brunt of the injection/extraction code will live. There are three options as I see it:

1. In the N RPC/IPC frameworks that *use* OpenTracing (probably via string-map-or-blob sorts of methods)
1. In OpenTracing implementations, where there will be N mostly-similar adapters for the N frameworks above
1. As a bunch of hacks in application code on either side of the RPC/IPC frameworks (nobody wants this, I hope)

I am in favor of moving towards (1), as then the code that pertains to "RPC/IPC Framework X" can live entirely in that repo and we don't need having this patchwork of support and non-support across the various OpenTracing impls in the world. I believe you are in favor of (2), either because you think N is small (I would strongly disagree about that), or because you think there are performance issues with (1), or perhaps because the code just looks gross with the generic maps/etc. :) Or something else?

Thoughts?

PS: I should have mentioned that naming is not an important part of this RFC, and "wire" will surely not survive to 1.0!

tschottdorf:
> That is by design - if you write a new Span, only you know how to serialize it.

That doesn't sound like a design choice, but rather an uncomfortable truth which the API is trying to hide by providing interfaces that suggest that any Span and any serializer can pair up (but in fact only those which live in the same impl can do it).

The Tracer knows the concrete type of the Span. So extracting is easy: a variety of objects can be passed to a method, and you get a `Span` back (which impl you get is determined by the tracer depending on what's in the object you're passing).
Can that be pluggable? Not really. It's hard-coded in the tracer anyway. No point in registering anything -
something plain and boring like `Extract(interface{}) (Span, error)` on the tracer makes sense.

For injecting, similar story. It doesn't really make sense to have a `Span` passed to the `Tracer`, hoping that it'll be able to serialize it somehow (after all, you don't know whether the `Span` is your impl, and type assertions there are a sign of bad abstraction). Instead, this should live on the one object that still knows itself, the Span. Since all concrete types of Span are known to Tracer, the corresponding code really lives in the Tracer impl, but is exposed via the Span: `sp.Inject(interface{}) error`.

In that proposal, there aren't any "global formats" for transmitting traces. But I think that's just how things go if you don't want to commit to what a Span actually **is**.

Here's the resulting interface (with some simple renaming I found appropriate):

```go
type Tracer interface {
  // Join attempts to obtain a Span from the given interface.
  Join (interface{}) (Span, error)
}

type Span interface {
  // Inject attempts to store the Span's information in the given interface.
  Propagate(interface{}) error
}

t := something.NewTracer()
sp := t.StartTrace("foo")
var req http.Request
// This will add the trace information to the request's headers.
if err := t.Propagate(&req); err != nil {
  return err
}

// Receiving end uses the Tracer to join a new Span onto the one
// encoded in the Request.
if sp, err := t.Join(&req); err != nil {
  // ...
}
```

The situation changes when you commit to a type of Span. For example, for `standardtracer` one could add pluggability to the tracer to account for users who are happy with the trace format provided by `standardtracer` and only need to be more flexible in how they pass those around the system.

tschottdorf:
@bensigelman our messages crossed. I'd be interested to hear what works and what doesn't with my simpler suggestion below. Ideally, you can give a concrete use case that would cause trouble - I haven't seen many concrete examples in the discussions, but that would be really helpful.

yurishkuro:
@tschottdorf If you review @bensigelman 's points 1-3 above, you will see that your API falls roughly in (2) camp. I am on the same page as Ben in preferring option (1).

On the placement of the injector methods, you, Ben, and I seem to agree that it should be on the Span directly, not on the Tracer.

Note that your API proposal is not significantly different from @michaelsembwever 's or Ben's, except that they use a factory method that returns the codec. The specific example you were asking for is just this - how can a Tracer or Span know how to marshal the span in/out of an `interface{}`?  Such API provides no help to either side; the end user doesn't know what to pass into such method, the tracer implementation doesn't know what it is being passed. With the factory method, the tracer is first told what kind of `interface{}` it's expected to marshal.

bensigelman:
@tschottdorf, @yurishkuro already pointed this out, but I will rephrase in my own words just for clarity. Let's look at the code you proposed (thank you for that, btw: makes this way more concrete):

```

**Comment by bensigelman** [comment]
@tschottdorf, @yurishkuro already pointed this out, but I will rephrase in my own words just for clarity. Let's look at the code you proposed (thank you for that, btw: makes this way more concrete):

```
t := something.NewTracer()
sp := t.StartTrace("foo")
var req http.Request
// This will add the trace information to the request's headers.
if err := t.Propagate(&req); err != nil {
  return err
}
```

This is a reasonable(*) API if and only if there are a small number of things like `http.Request` for every OpenTracing implementation to support. Unfortunately I don't think that's the case. I may not have time to finish before my toddler wakes up, but at least before EOD I will send out another version of this RFC that tries to create a little bit of a middleground between the options 1 and 2 I propose. I also want to be clear in saying that I very much like the *idea* of what you've suggested, but have trouble imagining how it's going to work with the innumerable home-grown RPC/IPC frameworks in the world: how could an OpenTracing impl know about them all?

(*): I'd rather use package-specific values like `net.Context` does – or even a plain-old namespace of `strings` – to identify formats. What you've done here might work for `http.Request`, but I can easily imagine more than one format for `[]byte`, and `t.Propagate(...)` needs a way to differentiate between them (beyond plain-old reflection).

**Comment by bensigelman** [comment]
All, please take a fresh look at this proposal... I've adjusted a bunch of stuff and tried to aim for something halfway between the "1" and "2" of my earlier comment: https://github.com/opentracing/opentracing-go/pull/48#issuecomment-181044269

File: span.go [hunk 8]
```diff
@@ -52,6 +53,24 @@ type Span interface {
...
+	// used by the `net.Context` package (see
+	// https://godoc.org/golang.org/x/net/context#WithValue).
+	//
+	// Example usage (sans error handling):
+	//
+	//     span.PropagationInjectorForFormat(
+	//         opentracing.PROPAGATION_FORMAT_GO_HTTP_HEADER).InjectSpan(
+	//         span, httpReq.Header)
+	//
+	PropagationInjectorForFormat(format interface{}) PropagationInjector
```
bensigelman: Per the concerns about the `enum` being non-extensible... Well, duh, yes, enums are not extensible! The idea was to basically force propagation formats to be centralized and defined by OpenTracing.

The model proposed here is a little more flexible: I still think the best practice will be to fall back on one of the "builtin" formats, but it's also possible – either for cleanliness or just as an optimization – for external codebases to define their own propagation formats.

FYI, for those who don't know Go very well, the type is part of the identity of an `interface{}` value... so even if two packages typedef `int` (as `BuiltinPropagationFormat` is typedef'd in this PR), the values don't compare as equal: http://play.golang.org/p/I2dnRUxofy

span.go
```diff
@@ -145,3 +164,13 @@ func CanonicalizeTraceAttributeKey(key string) (string, bool) {
 	}
 	return strings.ToLower(key), true
 }
+
+// InjectSpan is a helper that takes some of the stuttering out of the common
+// Span.PropagationInjectorForFormat calling pattern.
+func InjectSpan(sp Span, format PropagationFormat, carrier ...interface{}) error {
```
bensigelman: As before, we could just get rid of this, too.

utils.go
```diff
@@ -16,18 +17,28 @@ const (
...
+func InjectSpanInHeader(
 	sp Span,
 	h http.Header,
-	propagator SpanPropagator,
-) {
-	contextIDMap, tagsMap := propagator.PropagateSpanAsText(sp)
-	for headerSuffix, val := range contextIDMap {
+) error {
+	// First, look for a PROPAGATION_FORMAT_GO_HTTP_HEADER injector (our preference).
+	injector := sp.PropagationInjectorForFormat(PROPAGATION_FORMAT_GO_HTTP_HEADER)
```
bensigelman: @tschottdorf @yurishkuro please read this function and the one below... the idea is that propagation code could first check for its format of choice (whatever that may be), then fall back on something like `PROPAGATION_FORMAT_SPLIT_TEXT` if the OpenTracing impl doesn't know about the aforementioned "format of choice".

propagation.go
```diff
@@ -0,0 +1,103 @@
...
+	// The `carrier` for injection and extraction must be a `*TextCarrier`
+	// instance.
+	PROPAGATION_FORMAT_SPLIT_TEXT
+
+	// PROPAGATION_FORMAT_GO_HTTP_HEADER encodes the Span into a Go http.Header
+	// instance (both the tracer state and any Trace Attributes).
+	//
+	// The `carrier` for both injection and extraction must be an http.Header
+	// instance.
+	PROPAGATION_FORMAT_GO_HTTP_HEADER
```
bensigelman: I don't like the idea of OpenTracing adding tons of builtin formats, but maybe we could stick with standard-library representations of super-common things like HTTP headers, etc. Or, say, maybe `opentracing-ruby` would have a builtin format for some Rails type just because so much of Ruby code is actually Rails. Et cetera... It would be an ongoing compromise, and IPC codebase maintainers would be strongly encouraged to fall back on one of the general-purpose text/binary formats as a contingency.

tracer.go
```diff
@@ -23,4 +21,22 @@ type Tracer interface {
...
 	//
 	StartTrace(operationName string) Span
+
+	// Return a PropagationExtractor for the given `format` value, or nil if the
+	// Span does not support such a format.
+	//
+	// OpenTracing defines a common set of `format` values (see
+	// BuiltinPropagationFormat), and each has an expected carrier type.
+	//
+	// Other packages may declare their own `format` values, much like the keys
```
bensigelman: (FWIW, I would not imagine this pattern working well in non-Go... other languages could just use `String` for their `format` params)

span.go
yurishkuro: I know that naming wasn't the main objective of this PR, but to make it less scary I think a simple `Injector(format)` signature achieves the same results..

utils.go
yurishkuro: I am a bit reluctant to suggest to instrumentation coders to use a fallback mechanism as a standard practice. From "API as a standard" point of view I would rather say that whichever formats are declared in the OpenTracing API are mandatory for all implementations. Additional formats / injectors can be registered with the Tracer at start-up.

propagation.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// TextCarrier breaks a propagated Span into two pieces.
+//
+// The Span is separated in this way for a variety of reasons; the most
+// important is to give OpenTracing users a portable way to opt out of Trace
+// Attribute propagation entirely if they deem it a stability risk.
+type TextCarrier struct {
```
yurishkuro: @bensigelman why not use an interface that does not require intermediate map allocation?

```go
type TextInjector interface {
    AddTracerStateField(key, value string)
    AddTraceAttribute(key, value string)
}
// util function to construct TextInjector from maps
func NewMapBasedTextInjector(tracerState map[string]string, traceAttributes map[string]string) TextInjector {}

// util function to construct TextInjector for http.Header
func NewHttpHeaderTextInjector(header *http.Header) TextInjector {}
```

Note that this allows varying the argument to the Injector's inject method without varying the format.

The extractor part is a bit tricker, as it needs direct by-key access for tracer state, and possibly iterator over all keys for trace attributes.

On second thought, this doesn't have to *replace* TextCarrier, could be another standard format (better performance, less copying).

propagation.go
```diff
@@ -0,0 +1,103 @@
...
+// propagation formats.
+type BuiltinPropagationFormat int
+
+const (
+	// PROPAGATION_FORMAT_SPLIT_BINARY encodes the Span in a BinaryCarrier
+	// instance.
+	//
+	// The `carrier` for injection and extraction must be a `*BinaryCarrier`
+	// instance.
+	PROPAGATION_FORMAT_SPLIT_BINARY BuiltinPropagationFormat = iota
```
yurishkuro: we should use shorter names for constants, they make the usage code very hard to read.

michaelsembwever:
@tschottdorf 
```
type Tracer interface {
  // Join attempts to obtain a Span from the given interface.
  Join (interface{}) (Span, error)
}

type Span interface {
  // Inject attempts to store the Span's information in the given interface.
  Propagate(interface{}) error
}
```
This is by far the best design I've seen so far. I like the use of the words join and propagate.

I'd rather see similar concepts in similar places; reducing the risk in API compatibility by aligning methods to be where they functionally exist. You mention "the corresponding code really lives in the Tracer impl". By doing this it makes life easier for the outsider to quickly grasp the dynamics of the API, for example that join & propagate are paired methods. Probably more importantly it is more future-proof, maybe a tracing implementation will need to support `tracer.propagate(..)` where `tracer != span.tracer`. I also understand the problem of different Span implementations, but that would have greater impact that just the implementation of `span.propagate`; at the end of the day the tracer still needs to know about such custom Span implementations.

```
type Tracer interface {
  // Join attempts to obtain a Span from the given interface.
  Join (interface{}) (Span, error)
  // Inject attempts to store the Span's information in the given interface.
  Propagate(Span, interface{}) error
}
```

I'd be happy to do away with the factory methods, given the API provides the same level of clarify  (which only yours has so far) and solves the problem of delegation as explained below.

---- 

From @bensigelman 
> @michaelsembwever I strongly agree with @yurishkuro that we shouldn't expect that OpenTracing implementors write code for each web/RPC/etc framework in popular use. This is of course the motivation for using general-purpose encodings like string maps and blobs.

From @yurishkuro 
>  If I implement a new http framework that uses a custom class to represent the Request, there must be a way in OpenTracing to still use the available injectors to inject the span into my custom Request class, until the time my framework becomes so popular that my Request class is supported natively by all tracers. And the only way to provide the portable injectors is to make assumptions about the details of the serialized representation and provide a common-denominator API, like the kv-map + binary that we had originally.

It's frustrating that you're bringing this up again.
This has already been explained in that it would be solved through delegation.
General purpose injectors/extractors (propagators/joiners) would exist in the OT implementations. These general purpose implementations would match each protocol specification. Delegating to these would be the framework specific injectors/extractors. 

Such a delegation approach means
 a) approach (2) doesn't mean more code,
 b) the problem of all the different http Request frameworks becomes a rather trivial exercise,
 c) new frameworks have to write very little code. 

Therefore we need not care about the difference between (1) and (2), just let evolution take its course. 

This also goes back to @bensigelman 's argument against @tschottdorf proposal:
> This is a reasonable[1] API if and only if there are a small number of things like http.Request for every OpenTracing implementation to support. 

There might not be a small number of http.Request implementations to support, but the support of them becomes trivial through a delegation model: the bulk of the functionality is already provided by the general purpose implementations. 

---- 
@bensigelman 
I don't see the point in having `BuiltinPropagationFormat`, `BinaryCarrier`, `TextCarrier` explicitly in the API. As far as I can see it's just noise. I understand where you've coming from, but my feeling is you're trying to solve too much, and it's becoming convoluted. 

In the original injector/extractor proposal, or in @tschottdorf's join/propagate proposal, this notion of the provided general purpose implementations is implied. It is accessed through the existing API, and can be referenced instead by just the Specification and example use-case documentation.

tschottdorf:
@michaelsembwever:

> Probably more importantly it is more future-proof, maybe a tracing implementation will need to support tracer.propagate(..) where tracer != span.tracer

Is that a thing? Sounds like there's no reason to support that. Within the life of a single trace in a single context, I shouldn't just switch out the tracer. Sounds like an academic/exotic use case best advised against (?).

Coming from Go-land and disliking global singletons, I feel strongly that the `Span` should know how to serialize itself, simply because that's all I'm going to have when I actually want to do that.

span.go
tschottdorf: you need to return an error here or in the method on PropagationInjector to give things a chance to fail.

tschottdorf: sorry, overlooked the `nil`. I guess that's ok, though it looks slightly non-idiomatic. Since the whole point of this exercise, maybe the `PropagationInjector` should have a chance to return an error, though: you don't know what it does, and while your standard impls will never hit a problem, others might.

utils.go
tschottdorf: +1 on no fallbacks.

michaelsembwever:
> > Probably more importantly it is more future-proof, maybe a tracing implementation will need to support tracer.propagate(..) where tracer != span.tracer
> 
> Is that a thing?

The goal of OT is rather ambitious: to provide an API that different tracing frameworks can implement. Then on top of this all the different languages to all the different tracing solutions have to be supported. 

Not only is OT going to evolve but so are all the different tracing implementations.
You can bet your bottom dollar OT's initial API will be wrong.
Building an API for evolution, rather than today's ideal, should take precedence. I've tried to allude to this numerous times, when mentioning a separation from porcelain to plumbing api. 

Right now let's just focus on the plumbing.

> Coming from Go-land and disliking global singletons, I feel strongly that the Span should know how to serialize itself, simply because that's all I'm going to have when I actually want to do that.

Is Go-land's dislike for singletons influencing the API over first principle?

This issue also came up with `span.create_child(..)`.
First everyone agreed that it was superfluous and should be removed.
It was a duplicate of `tracer.create_span(..)`.
Then it was snuck back in, for the same reason: it was more **convenient** to have it also in Span. This aspect of convenience, in my mind, doesn't belong in the first-pass of the API.

tschottdorf:
I still think the generality here is too high and reuse basically impossible.
There's a lot of talk about "standard encodings" here, but what does that even mean if the structure of the `Span` isn't prescribed? Sure, I know what the binary encoding means for `RawSpan`, but it immediately doesn't make any sense any more once my `StandardContext` looks different. You need **some** assumptions about a trace's structure to have things make at least some sense. For example, take the attributes - that's fairly straightforward; a collection of keys and values (suitably restricted). There aren't that many ways of representing that, and code reuse is possible:

```
package binary

func EncodeAttributesFromMap(source map[string]string, dest []byte) []byte { /* ... */ }
func EncodeAttributesFromIterator(next func() (k string, v string), dest []byte) []byte { /* ... */ }
func EncodeAttributesFromPairs(dest, kvs ...string) []byte { /* ... */ }
```

That part is very amenable to reuse. In my custom tracer with a custom span, I'm somehow going to have to store the attributes, and one of those functions above will likely allow me not to have to worry about what the actual format on the wire is.

If a similar assumption held for the remainder of the trace, we could actually **define** the binary serialization format (and similarly http, and many others) and a new tracer implementation would indeed be able to support the "standard formats" without being left stranded for a lack of definition.

For creating a new span, similar thought experiment:

```go
package binary

func DecodeAttributes(source []byte, callback(string, string)) { /* ... */ }
func DecodeContext(source []byte, callback(string, interface{})) { /* ... */ }
// ...
```

with example usage:

```go
func (*binaryExtractor) ExtractSpan(opName string) (Span, error) {
  var sp customSpanImpl
  cb := func(k, v string) {
    switch k {
      case "ID":
        if _, ok := v.(int32); !ok
          return nil, errors.New("invalid type")
        }
        sp.Identifier = v.(int32)
      // ...
    }
  }
  // ...
}

```

Not perfect, but I don't see how we can talk about "default formats" when the object in question could be *anything*.

yurishkuro:
@michaelsembwever 

> It's frustrating that you're bringing this up again. This has already been explained in that it would be solved through delegation. General purpose injectors/extractors (propagators/joiners) would exist in the OT implementations. 

I don't disagree on the frustrating part. I would like to see an example of tracer implementation that does "general purpose injectors/extractors". imo that is exactly what Ben has in the `TextCarrier` example, an intermediary representation that specific frameworks can transform to their specific wire protocols.

tschottdorf:
@michaelsembwever: Sorry, I should add that I don't think the API needs to be fixed across languages. If you're in Java-land and having both items on the Tracer itself works for general Java use-cases, great. 

It's just that I don't feel so strongly that they need to be next to each other in Go. The Span goes more places than the Tracer, and it's the Span you want to serialize so it's simply more powerful to have that functionality there. And since a Span is owned by the Tracer (another tracer wouldn't even know do to anything with it, except for what's provided by the interface), I don't think there's room for confusion.

michaelsembwever:
> that is exactly what Ben has in the TextCarrier example, an intermediary representation that specific frameworks can transform to their specific wire protocols.

What Ben has is not delegation. It introduces a whole lot extra into the API.
There is still a lot to grasp around the anything structure Span and how that works with injectors/extractors. It wouldn't surprise me though that it's easier to solve once (just for injector/extractor), rather than twice (injector/extractor and carriers).

yurishkuro:
@tschottdorf 

> Not perfect, but I don't see how we can talk about "default formats" when the object in question could be anything.

What you call "default formats" I call an intermediate representation. A key-value map is an example of one such representation that any span can be marshalled into: 
* one or more keys for "tracer state" (trace id, span id, whatever the tracer wants to propagate). 
* and one or more keys for tracer attributes

Another intermediate representation is two binary blobs. One of these two intermediate representations can be injected into the majority of the existing RPC protocols, but the authors of that RPC framework, not by the OpenTracing implementation.

This was the original design, which works well in several already written tracers, but it was criticized as too specific and not extensible (fair points). This PR is an attempt to provide a more general API where instead of explicit methods producing exactly 2 representations (kv-map and binary), the injectors are more general / pluggable. The OT implementation will be required to provide a minimal set of injectors, that can be used to build more specific injectors if an RPC framework so desires. But the minimal injectors must be flexible enough be suitable for a large variety of target protocols, which makes them near identical to the same 2 intermediate representations we had before.

tschottdorf:
Ah, sorry. I overlooked the part in the diff that has the TraceState -- am I right that what I'm proposing above is identical to what @bensigelman was planning?

michaelsembwever:
@tschottdorf why can't the plugin implementations for injector/extractor (propagator/joiner) be tied to the Span implementation? 

Given "a Span is owned by the Tracer", then surely we have what it takes for Injectors/Extractors to work with different Spans.

If you write a new Span type, then you're writing a new Tracer. If you're writing a new Tracer type then you're writing the  "general purpose injectors/extractors". 

From this all higher-level injectors/extractors just work, because they delegate to the general purpose injectors/extractors you've just written.

This is done without having the `BuiltinPropagationFormat`, `BinaryCarrier`, and `TextCarrier` APIs.

Am I missing something?

tschottdorf:
@michaelsembwever I'm not sure which statement of mine you're referring to, but one of the upshots that I see with the proposal here is that for all tracers whose context and attributes are basically key-values, `opentracing` can provide standard serializers.

So for the vast majority of custom tracers, no serialization code has to be written (just some boilerplate to call the plumbing functions we could provide, example signatures of which I've collected above).

That saves work and should also aid in having different tracers "converge" on a few encodings which gives more chance of interop.

propagation.go
tschottdorf: agreed. And does it have to be an enum? Why not just

```go
package encoding

type Binary struct{}
type HTTP struct{}
```

that's what would happen for custom wire types elsewhere.

propagation.go
tschottdorf: As noted in the comment on the main thread, I don't think the explicit carriers are useful. Decent low-level functions for encoding key-value pairs are (tracers will need to write implementation boilerplate, and that's what we need to support - not another intermediate type).

span.go
tschottdorf: I'd toss this, at least for now.

michaelsembwever:
@tschottdorf 
> I'm not sure which statement of mine you're referring to, but one of the upshots that I see with the proposal here is that for all tracers whose context and attributes are basically key-values, opentracing can provide standard serializers.

Yeah, but that was always the case.
What's happened here is that a whole lot of API baggage has been added.

utils.go
bensigelman: @yurishkuro @tschottdorf I should explain this more... @yurishkuro, I agree that the formats declared within OpenTracing proper are mandatory. This example was meant to illustrate what some off-brand RPC library could do with their *own* format (i.e., use it if available, but fall back to something standard if not). So, imagine for the sake of illustration that the GO_HTTP_HEADER format is non-standard in this example.

propagation.go
bensigelman: @yurishkuro I am going to ignore your (totally reasonable) suggestion until we settle the larger questions here :)

@tschottdorf I'll reply on the main thread.

propagation.go
bensigelman: I have no strong feelings here... I was copying best practices from a `net.Context` blog post from Go-team. @tschottdorf what you have here is equally fine with me.

propagation.go
```diff
@@ -0,0 +1,103 @@
...
+///////////////////////////////////////////////////////////////////////////////
+
+// PropagationInjector is responsible for injecting Span instances in a manner
+// suitable for propagation via a format-specific "carrier" object. Typically
+// the injection will take place across an RPC boundary, but message queues and
+// other IPC mechanisms are also reasonable places to use a
+// PropagationInjector.
+//
+// See PropagationExtractor and Span.PropagationInjectorForFormat.
+type PropagationInjector interface {
```
michaelsembwever: -1 on the name. `SpanInjector`, `SpanPropagator` or just `Injector`, are all more intuitive.

utils.go
michaelsembwever: I missed this comment. sry.

The implementation looks good.

Like i've said before I'd rather see a clean separation of how a span is put on a protocol's wire from the particular framework at hand by re-using the same API and using a delegation approach.

That is the notion of a "carrier" becomes the same as the "interface" in the Injector/Extractor.

So in this situation you have the delegation model of:

    HttpHeaderSpanInjector --> SplitTextInjector
This is easily done because the interaction between the two is a `map<string,string>` and is dictated by the latter. 

> the formats declared within OpenTracing proper are mandatory

Agreed. This can be done via `SplitTextInjector`, rather than adding the Carrier API concept.

In addition the notion of the traceState and traceAttributes does not need to be explicitly represented in the API. It is represented rather by the implementation of SplitTextInjector (and BinaryInjector).

bensigelman:
@tschottdorf:

> ... am I right that what I'm proposing above is identical to what @bensigelman was planning?

I can't say *identical*, but the spirit of what you proposed is very much in line with this PR. OpenTracing makes the assumption that every Span implementation can find a way to represent itself as a pair of string maps, as a pair of `[]byte`s, and – at least in this PR – as elements in an `http.Header` instance. It seems like that's what you're getting at as well.

One thing I'd point out, since you mentioned something similar in a recent thread: I would very much prefer that OpenTracing make no assumptions about the internals of the so-called "tracer state": there are working tracers that have literally no in-band tracer state, after all... and even Dapper-like things differ in the details (e.g., Dapper's "sampling bit" was actually more like a "sampling int32" that encapsulated a small bitmask and a reciprocal sampling rate... hard to standardize stuff like that).

@michaelsembwever:

> It's frustrating that you're bringing this up again.
>
> This has already been explained in that it would be solved through delegation.
>
> General purpose injectors/extractors (propagators/joiners) would exist in the OT implementations. These general purpose implementations would match each protocol specification. Delegating to these would be the framework specific injectors/extractors.

I am sorry that you're frustrated, and I agree that it sounds like we're talking past one another. Rarely a good thing.

Nevertheless, I am still not seeing a path forward where all of the following are true:
- OpenTracing doesn't have *some* concept like `BinaryCarrier`/`TextCarrier`
- Maintainers of RPC/IPC code that want to integrate OpenTracing can do so with a guarantee of portability across compliant OpenTracing implementations
- OpenTracing as a project does not need to add as much as a single symbol/string/function on behalf of the countless RPC/IPC frameworks and mechanisms out there in the world

By incorporating something akin to the builtin formats in this PR, we lose the first bullet above but check off the second and third. In my mind that's a great trade. @michaelsembwever you seem to think there's a silver bullet solution that gives us all three, but I don't see it. Maybe you're imagining that the larger OpenTracing community will converge in some organic fashion around certain standard formats?

If what I'm suggesting still doesn't make sense, please fork your Java PR (or this one, whatever) and illustrate your basic point, but please include some example calling code (like the short diff in utils.go here... nothing significant). Or hell, just do the example calling code and we can fill in the gaps.

bensigelman:
btw, all: I don't give a sh*t about naming in this PR, so I am just going to go through and accept all of your various suggestions along those lines. I don't want people to be distracted by names when it's the shape of this API that's most important here.

michaelsembwever:
> you seem to think there's a silver bullet solution that gives us all three, but I don't see it. 

By taking the delegation approach you achieve all three points.
I explained it in https://github.com/opentracing/opentracing-go/pull/48/files#r52133616

utils.go
michaelsembwever: Code example to above comment.
(I don't know go so excuse me if it turns into psuedo)

```
// HttpHeaderFrameworkInjector
func Inject(
	sp Span,
	h http.Header,
) error {
	map<string,string> headerMap = h
	splitTextInjector.InjectSpan(sp, headerMap)
}

// SplitTextFormatInjector
func Inject(
	sp Span,
	map Map<string,string>,
) error {
	for headerSuffix, val := [span.traceId, span.spanId] {
		h.Add(ContextIDHTTPHeaderPrefix+headerSuffix, url.QueryEscape(val))
	}
	for headerSuffix, val := range span.TraceAttributes {
		h.Add(TagsHTTPHeaderPrefix+headerSuffix, url.QueryEscape(val))
	}
}
```

bensigelman: @michaelsembwever You didn't miss the comment, it wasn't there yet :)  [side note: Github would do well to fix their pull request UX -- it's damned confusing when you're trying to catch up after a day (or night) away from the firehose of individual updates]

So perhaps we are actually "in violent agreement", at least for the most part?? I hope so, anyway.

As for the "carriers": the motivating question is how the OpenTracing user who is dealing with process-boundary concerns (i.e., the person writing code like the lines we're commenting about here) is expected to pass in the `http.Header` or `map[string]string` (or whatever) that they're trying to inject into. I know that you wanted to do this – in Java, that is – by having the Injector rely on introspection. So, to be clear, something like this:

```
func InjectSpanInHeader(
	sp Span,
	h http.Header,
) error {
	// First, look for a GoHTTPHeader injector (our preference).

	//////////////////////////////////////////////////////////////////////////////
	// NOTE: we pass the http.Header *instance* in, so there's no need to pass
	// it again via a carrier.
	//////////////////////////////////////////////////////////////////////////////
	injector := sp.Injector(myHTTPHeader)
	if injector != nil {
		return injector.InjectSpan(sp)
	}

	// Else, fall back on SplitText.

	//////////////////////////////////////////////////////////////////////////////
	// NOTE: we pass both text maps in, so there's no need to pass them again
	// via a carrier.
	//////////////////////////////////////////////////////////////////////////////
	tracerState := map[string]string{}
	traceAttrs := map[string]string{}
	if injector = sp.Injector(tracerState, traceAttrs); injector == nil {
		return errors.New("No suitable injector")
	}
	if err := injector.Inject(sp); err != nil {
		return err
	}
	for headerSuffix, val := range tracerState {
		h.Add(ContextIDHTTPHeaderPrefix+headerSuffix, url.QueryEscape(val))
	}
	for headerSuffix, val := range traceAttrs {
		h.Add(TagsHTTPHeaderPrefix+headerSuffix, url.QueryEscape(val))
	}
	return nil
}
```

The above feels sketchy in Go, though it can be made to work. I preferred the security/certainty of the intermediary Injector/Extractor interfaces, personally. But this is not a huge thing for me. If we were to go with something like the above, I would just assume shorten this:

```
	if injector = sp.Injector(tracerState, traceAttrs); injector == nil {
		return errors.New("No suitable injector")
	}
	if err := injector.Inject(sp); err != nil {
		return err
	}
```

... to this:

```
	if err = sp.Inject(tracerState, traceAttrs); err != nil {
		return err
	}
```

That is, get rid of the `Injector` interface, and just make `Span.Inject()` more complex for the implementors. At least it could be more efficient that way. (Minor point: `Span.Inject()` would need variadic parameters if we want to get rid of the carrier types, which are basically just included for clarity)

bensigelman:
I pushed some simple renames per various suggestions, and Github has kindly hidden the most important threads in this PR since the diffs are "outdated".

@michaelsembwever @tschottdorf @yurishkuro ^^^

utils.go
michaelsembwever: @bensigelman 
> As for the "carriers": the motivating question is how the OpenTracing user who is dealing with process-boundary concerns (i.e., the person writing code like the lines we're commenting about here) is expected to pass in the http.Header or map[string]string (or whatever) that they're trying to inject into. I know that you wanted to do this – in Java, that is – by having the Injector rely on introspection

There was no introspection required in my code example. And neither any need for variadic parameters.

What I'm looking for is the delegation model of

    HttpHeaderSpanInjector --> SplitTextInjector

where the end-user only needs to call `span.Injector(opentracing.HttpHeader).Inject(span, httpReq.Header)`

tschottdorf:
@bensigelman I think I see the difference between our approaches now. You provide a couple of objects and demand that all tracers somehow find a way to store themselves into (and back out of) them, but you don't say how. That means that every tracer will find their own way to somehow smush their stuff into a byte array, header, or whatever. I don't see a good way of reusing the code, or the upshot.

I'm aiming to actually **define** how the data is encoded, the goal being that the tracers don't actually have to **come up** with such a way. That is done by providing useful low-level primitives for these respective encodings/objects, so that the only thing that the implementations have to do is call these helpers with the appropriate key-values that describe their tracer state and be all set (if they're doing otherworldly complicated things, they can still do their own thing, of course).

Carriers as you describe will still exist, but implicitly in the user's codebase - they will be exactly how the user stores the serialized data, not in ways we define. For example, if I happen to do this in my codebase:

```go
type WireTrace [2][]byte // intentionally odd choice
type Session struct {
  // whole lot of stuff omitted
  EncTrace WireTrace
}
```

and assuming that I just want some standard encoding that plays with bytes, then my extractor for `encoding.Binary` (or whatever we call the official binary protocol) could be this:

```go
func (*oddExtractor) JoinTrace(operationName string, carrier interface{}) (Span, error) {
  wt := carrier.(Session).EncTrace // sans error handling
  sp := &spanImpl{}
  // Now I know that wt[0] and wt[1] hold the serialized data in standard
  // opentracing binary format and I can use the plumbing methods to get
  // the data and populate `sp`.
}
// Injection is in similar spirit.
```

The assumption here is that everyone will have slightly different ways of
storing information, and there's no point in trying to coerce them into a
format. Instead, what you want is that people don't waste their time writing
serializers.

Curious to hear whether I'm overlooking some hidden benefit of the current
suggestion compared to above.

yurishkuro:
@tschottdorf 

> I'm aiming to actually define how the data is encoded, the goal being that the tracers don't actually have to come up with such a way. 

If that is your goal, then I think it's way too early to talk about it, given that we can't still agree on the shape of the end-user facing API. Note that portability of serialized span representation between tracers is decidedly a **non-goal** for OpenTracing. Unifying how tracers achieve implementation of the API is pretty much a non-goal at this stage as well, because we can expect 1000s of end users using API to instrument their apps/libs, and only 10s of developers actually concerned with implementing the tracers, so we want to focus on the 1000s. Code reuse across tracing implementations is simply a problem that doesn't exist today.

That said, your examples of actually implementing the proposed API, starting from the public API signatures, are welcome. So taking the latest one:

```go
func (*oddExtractor) JoinTrace(operationName string, carrier interface{}) (Span, error) {
  wt := carrier.(Session).EncTrace // sans error handling
  sp := &spanImpl{}
  // Now I know that wt[0] and wt[1] hold the serialized data in standard
  // opentracing binary format and I can use the plumbing methods to get
  // the data and populate `sp`.
}
```

If I understand you correctly, that is an example of "odd" extractor written by an end-user, e.g. by some RPC framework owner. I think it's in the spirit of what @michaelsembwever refers to as delegation, but it has two problems. A minor one is it's missing the starting point of user calling `tracer.extractSpan(session)` - without having a factory method with enum of sort that code will not be pretty. Another minor one is `sp := &spanImpl{}` - that's not possible since end user has no idea about span implementations, only tracer does. But the main problem is the comment. What are those plumbing methods? Since this is end-user code, it has no access to any more methods than the public API of the tracer. So what should really happen is that this custom extractor now delegates to one of the official extractors that all tracers must support:

```go
func (*oddExtractor) JoinTrace(operationName string, carrier interface{}) (Span, error) {
  wt := carrier.(Session).EncTrace // sans error handling
  binCarrier := &BinaryCarrier{wt[0], wt[1]}
  return tracer.getExtractor(SPLIT_BINARY).extractSpan(binCarrier)
}
```

And that takes us back to needing to standardize on a thing like `BinaryCarrier`.

@michaelsembwever is that how you see delegation working? If not, can you show your examples (Java is fine if you prefer)?

yurishkuro:
@bensigelman I am happy with the latest version, aside from the copying requirement in the two carriers, which we can address separately (it's not worse than today's api in that respect).

tschottdorf:
> If I understand you correctly, that is an example of "odd" extractor written by an end-user, e.g. by some RPC framework owner.

So that isn't the person that has the tracer? To make this more concrete, your end-user here is a Google dev, but the Tracer is Dapper (and not under immediate control by the end-user)?

If so, that makes sense - I wasn't aware of that. I have performance concerns but agree with @yurishkuro that this should be ok with some iterations.
I still think the enum is ugly compared to format types as I suggested, but again that's a detail.

The only wildcard here is @michaelsembwever's direction, and I must admit that I'm not quite clear what that suggestion would look like (despite repeated glances at the existing code). It may be time for a dumbed-down, concrete, example.

propagation.go
```diff
@@ -0,0 +1,100 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+// BUILTIN PROPAGATION FORMATS:
+///////////////////////////////////////////////////////////////////////////////
+
+// BuiltinFormat is the shared type for builtin OpenTracing in-band
+// propagation formats.
+type BuiltinFormat int
+
+const (
```
yurishkuro: @bensigelman is it worth adding a comment about the closeness of this enum? I.e. that the API does not prevent extensions, because the factory method's actual argument is `interface{}`

yurishkuro:
@tschottdorf 

> So that isn't the person that has the tracer? To make this more concrete, your end-user here is a Google dev, but the Tracer is Dapper (and not under immediate control by the end-user)?

Our customary terminology is:
 * **end user** is the person instrumenting some application or and RPC framework. They only have access to OpenTracing API, and they have no idea about implementation details of the tracer, such as what data it uses to represent a span on the wire, for example. The only place where the end user is exposed to a concrete implementation of the tracer is when they create a singleton instance of it and inject wherever needed as a `Tracer` API. That is the O(1) configuration that documentation refers to - if they want to swap for a different tracer (e.g. zipkin for dapper), they just change that one place with initialization, not the actual instrumentation code they've written elsewhere
* **implementor** is the developer writing Tracer implementation, Span implementation, standard injectors/extractors. We need to make their job doable, but not necessarily easy, i.e. it's much more important for the end user's job to be easy.

@bensigelman maybe we should have a section in the docs about ^^^.

bensigelman:
@tschottdorf and I had a quick conversation and I think we're pretty much on the same page: namely, that the approach here works modulo some Go-specific performance optimizations. I also got rid of the `enum` entirely.

propagation.go
bensigelman: @yurishkuro no longer relevant... I used the idiom @tschottdorf suggested (six vs half-dozen for me, but whatever)

bensigelman:
@yurishkuro 

> Our customary terminology is:
> ...
> @bensigelman maybe we should have a section in the docs about ^^^.

That seems wise, yes!

yurishkuro:
> I also got rid of the enum entirely.

is there a plan to add functionality to those types?  Otherwise this trades one line in the enum-version for two lines of code.

michaelsembwever:
from @tschottdorf 
> Carriers as you describe will still exist, but implicitly in the user's codebase - they will be exactly how the user stores the serialized data, not in ways we define. 

jackpot.

i can see that it's time I do the equilivalent PR in java to better express what I think can be simplified here (does away with the explicit concept of Carriers). I hope this PR can remain open in parallel.

bensigelman:
@michaelsembwever no problem re keeping this PR open in parallel.

bensigelman:
@yurishkuro @tschottdorf regarding the `enum`-vs-`type BlahType struct{}` thing: personally I don't care... the calling code will be identical in either case. The argument for the `type BlahType struct{}` version is that (a) it's perhaps slightly cleaner since the value literally takes zero storage and (b) it doesn't imply that there's anything special about the `type BuiltinFormat int` like the `enum` version might. The argument for the `enum` version is that it is indeed fewer LOC and there are N+1 exported "things" (be they types or values) relating to builtin formats rather than N*2 exported "things" in the `type BlahType struct{}` version.

I definitely don't care, though. :)  Users will get accustomed to either, and this is a Go-specific concern in either case.

tschottdorf:
As you pointed out, I like the symmetry with what users will have to do and it's semantically right because a `struct{}` contains no information, and that's what you want. Besides, if we changed the enum order (hypothetically, yes, but still) we'd screw everyone over. That doesn't seem right.
It's also a minor issue for me, just tickled my idiomaticity OCD.

yurishkuro:
I don't have strong feelings either way, type+const are fine. Maybe at some point Go will decide to join the rest of the 21st century languages and have generics, then the strongly typed constants could be used to imply the argument type for the injector (which is what @michaelsembwever is trying to do in Java).

bensigelman:
@bensigelman:

> @michaelsembwever no problem re keeping this PR open in parallel.

@michaelsembwever: no problem re keeping this PR open for a little while. :)  But since none of the three of us see how what you're suggesting is going to work in Go, I would love it if you could write even the calling code side in Java (i.e., no need to flesh out the APIs beneath unless it's necessary for understanding) so we can assess whether your magic bullet is applicable in Go.

Incidentally, I wrote and merged some docs about the relative priority of callers and callees in OpenTracing: http://opentracing.io/use-cases/#stepping-back-who-is-opentracing-for (@tschottdorf, would the above have cleared things up for you and saved time in the discussion above? That would be the success metric for that particular snippet!)

Thanks all...

tschottdorf:
@bhs: yep, this is helpful. I'm omitting comments on the code snippets
since you'll probably be plowing through them after API stabilization
anyway.

utils.go
```diff
@@ -51,15 +76,15 @@ func JoinTraceFromHeader(
 			if err != nil {
 				return nil, err
 			}
-			contextIDMap[strings.TrimPrefix(key, ContextIDHTTPHeaderPrefix)] = unescaped
+			carrier.TracerState[strings.TrimPrefix(key, ContextIDHTTPHeaderPrefix)] = unescaped
```
beberlei: The HTTP Headers don't start with X-, is that expected?

Also these headers are not documented in the specification. Are `Open-Tracing-Context-Id-` and `Open-Tracing-Trace-Tags-` supposed to be used for HTTP communication with headers?

bensigelman: re `X-`: http://tools.ietf.org/html/rfc6648

As for the other point: this file should probably go away. Specific header prefixes are an OT implementation issue, not an OT standards issue. (OT makes no guarantees about two communicating processes running different OT impls)

bensigelman: I like this file because it serves to illustrate changes to the (subtle!!) implications of changes to injection/extraction APIs... but it could easily move to `examples/` for now. Per http://opentracing.io/use-cases/#opentracing-priorities, implementors are a distant second priority to users at the moment.

yurishkuro: `Open-Tracing-Context-Id-` should definitely go away, we even have a ticket for it.

bensigelman:
Now that #50 is merged, I am going to rebase this and actually hook it up with the tests/examples/etc. I will also move the `Injector()` and `Extractor()` methods to `Tracer` since we have a `Span.Tracer()` method which makes it easy to write a helper for injection that only takes a `Span` instance.

bensigelman:
This is now considerably more fleshed out... I made a few small changes to the API itself and then made sure the example and tests worked. I'm pretty happy with where this is right now, though we could make some of this a tiny bit faster at the possible expense of ease-of-use. I'm open to feedback/suggestions in any case.

I did ultimately move the Injector and Extractor methods into the same interface... @michaelsembwever has made various arguments in favor, and since there's now a `Span.Tracer()` method I no longer see a counterargument from a convenience standpoint.

propagation.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+///////////////////////////////////////////////////////////////////////////////
+// BUILTIN PROPAGATION FORMATS:
+///////////////////////////////////////////////////////////////////////////////
+
+// BuiltinFormat is used to demarcate the values within package `opentracing`
+// that are intended for use with the Tracer.Injector() and Tracer.Extractor()
+// methods.
+type BuiltinFormat byte
```
bensigelman: @tschottdorf I went back to the enum, btw... it was nice to be able to refer to `BuiltinFormat` in comments about acceptable values for `format` params.

span.go
```diff
@@ -179,3 +180,15 @@ func StartChildSpan(parent Span, operationName string) Span {
 		Parent:        parent,
 	})
 }
+
+// InjectSpan is a helper that injects a Span instance into a carrier.
+//
+// See Tracer.Injector() and Injector.InjectSpan().
+func InjectSpan(sp Span, format interface{}, carrier interface{}) error {
```
bensigelman: (this is now easy to implement given `sp.Tracer()`)

standardtracer/propagation.go
```diff
@@ -159,53 +222,52 @@ func (t *tracerImpl) JoinTraceFromBinary(
 	), nil
 }
 
-func (t *tracerImpl) JoinTraceFromText(
+const (
+	tracerStateHeaderName = "Tracer-State"
+	traceAttrsHeaderName  = "Trace-Attributes"
+)
+
+func (p goHTTPPropagator) InjectSpan(
```
bensigelman: this one is marginally interesting... just shows how an injector and extractor can defer to other injectors and extractors to good effect.

propagation.go
tschottdorf: Still strikes me as a bad choice, but I've made my argument and can live with it either way.

span.go
```diff
@@ -179,3 +180,15 @@ func StartChildSpan(parent Span, operationName string) Span {
...
+	var inj Injector
+	if inj = sp.Tracer().Injector(format); inj == nil {
+		return fmt.Errorf("Unsupported PropagationFormat: %v", format)
+	}
+	inj.InjectSpan(sp, carrier)
+	return nil
```
yurishkuro: I think you should return the result of L192

utils.go
```diff
@@ -16,20 +17,33 @@ const (
 	TagsHTTPHeaderPrefix = "Open-Tracing-Trace-Tags-"
 )
```
yurishkuro: can we remove these?

utils.go
```diff
@@ -16,20 +17,33 @@ const (
...
+	// Else, fall back on SplitText.
+	if injector = sp.Tracer().Injector(SplitText); injector == nil {
+		return errors.New("No suitable injector")
+	}
+	carrier := NewSplitTextCarrier()
+	if err := injector.InjectSpan(sp, carrier); err != nil {
+		return err
+	}
+	for headerSuffix, val := range carrier.TracerState {
 		h.Add(ContextIDHTTPHeaderPrefix+headerSuffix, url.QueryEscape(val))
```
yurishkuro: I think the method should take the prefixes as arguments, to really make it general purpose util method.

noop.go
```diff
@@ -68,7 +70,25 @@ func (n NoopTracer) StartSpanWithOptions(opts StartSpanOptions) Span {
...
+	return defaultNoopInjectorExtractor
+}
+
+// Injector belongs to the Tracer interface.
+func (n NoopTracer) Injector(format interface{}) Injector {
+	return defaultNoopInjectorExtractor
+}
+
 // JoinTrace belongs to the Tracer interface.
 func (n NoopTracer) JoinTrace(operationName string, parent interface{}) Span {
```
yurishkuro: I thought it's StartSpan now.

span.go
bensigelman: definitely!

propagation.go
bensigelman: Thanks for being flexible... the old version made `Injector()` comment unwieldy...

utils.go
bensigelman: Yeah, I'd like to (a) remove this, (b) move `standardtracer` into the `examples` package or similar, and (c) further generalize the utils stuff per your other comments.

In my mind the above could and should be its own PR, though.

utils.go
bensigelman: (see https://github.com/opentracing/opentracing-go/pull/48#discussion_r52792470... I agree, FWIW)

noop.go
bensigelman: This is deadcode... removed some other stuff, too.

noop.go
```diff
@@ -68,7 +70,25 @@ func (n NoopTracer) StartSpanWithOptions(opts StartSpanOptions) Span {
...
 	return defaultNoopSpan
 }
+
+func (n noopInjectorExtractor) InjectSpan(span Span, carrier interface{}) error {
+	return nil
+}
+
+func (n noopInjectorExtractor) JoinTrace(operationName string, carrier interface{}) (Span, error) {
```
bensigelman: @yurishkuro do we want this to return the `defaultNoopSpan` or `nil` for the `Span` return val? I think `nil` is more honest, but it's a toss-up.

bensigelman:
IMO this is ready for merging unless there are further comments. I'll follow up with a change to demote standardtracer to "example code" status, clean up the utils stuff, etc.

noop.go
yurishkuro: it needs to return a noopSpan, otherwise end user code can NPE

bensigelman: Shouldn't a nil return be used to signal that no trace could be found to
join with? Or should we reserve an error val for that?

On Friday, February 12, 2016, Yuri Shkuro <notifications@github.com> wrote:

> In noop.go
> <https://github.com/opentracing/opentracing-go/pull/48#discussion_r52794134>
> :
>
> >  // JoinTrace belongs to the Tracer interface.
> >  func (n NoopTracer) JoinTrace(operationName string, parent interface{}) Span {
> >  	return defaultNoopSpan
> >  }
> > +
> > +func (n noopInjectorExtractor) InjectSpan(span Span, carrier interface{}) error {
> > +	return nil
> > +}
> > +
> > +func (n noopInjectorExtractor) JoinTrace(operationName string, carrier interface{}) (Span, error) {
>
> it needs to return a noopSpan, otherwise end user code can NPE
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opentracing/opentracing-go/pull/48/files#r52794134>.
>


-- 
(Sent via mobile)

yurishkuro: Ah, I remember now. So the original Go API I wrote had this semantics:

```
span, err := tracer.joinTrace()
if err != nil or span == nil {
  span = tracer.startTrace()
}
```

Some people complained about this being too verbose and asking "why can't joinTrace do it in one line". I think you now discovered why - if we do it all in Join, we will have to swallow the error that may potentially occur during tracer ctx deserialization (malformed header or what not). The above code was explicit and leaving error handling to the caller, which is the Go way. So I think we need to decide on the exact semantics of the extractor's Join method. My inclination is to have it deal with missing header (span==nil above), but expose decoding error, so that people may alert or it, or emit metric.

yurishkuro: In short, the semantics I'd prefer is this:
```go
span, err := tracer.extractor.JoinTrace(carrier)
if err != nil {
  // emit metric
  span = tracer.startSpan()
}
```

utils.go
yurishkuro: separate PR is fine.

propagation.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// Extractor is responsible for extracting Span instances from an
+// format-specific "carrier" object. Typically the extraction will take place
+// on the server side of an RPC boundary, but message queues and other IPC
+// mechanisms are also reasonable places to use a Extractor.
+//
+// See Injector and Tracer.Extractor.
+type Extractor interface {
+	JoinTrace(operationName string, carrier interface{}) (Span, error)
```
yurishkuro: per the comment on Noop, we need to document the exact semantics of this method

noop.go
bensigelman: I like the idea of defining a

```
var TraceNotFound = errors.New("trace not found by Extractor")
```

... so your example above would still work, but a user may be able to tell the difference between the not-found case and the parse-error case...

propagation.go
bensigelman: agreed... same with InjectSpan. I'll do this stuff when I'm back at my computer in an hour.

noop.go
yurishkuro: +1 to std error

tschottdorf: Yeah, that's the way to go.

bensigelman: (done)

bensigelman:
@yurishkuro @tschottdorf I think this is ready for a (final?) look... I'll leave it open until I get at least one explicit LGTM :)

tracer.go
```diff
@@ -28,6 +26,48 @@ type Tracer interface {
...
+	// used by the `net.Context` package (see
+	// https://godoc.org/golang.org/x/net/context#WithValue).
+	//
+	// Example usage (sans error handling):
+	//
+	//     tracer.Injector(
+	//         opentracing.GoHTTPHeader).InjectSpan(
+	//         span, httpReq.Header)
+	//
+	// NOTE: All opentracing.Tracer implementations MUST support all
```
bensigelman: (new)

standardtracer/propagation.go
```diff
@@ -0,0 +1,291 @@
...
+
+func (p *goHTTPPropagator) JoinTrace(
+	operationName string,
+	carrier interface{},
+) (opentracing.Span, error) {
+	// Decode the two base64-encoded data blobs from the HTTP header.
+	header := carrier.(http.Header)
+	tracerStateBase64, found := header[http.CanonicalHeaderKey(tracerStateHeaderName)]
+	if !found || len(tracerStateBase64) == 0 {
+		return nil, opentracing.TraceNotFound
```
bensigelman: (I added the shiny new standard errors here and there)

propagation.go
```diff
@@ -0,0 +1,133 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+// CORE PROPAGATION INTERFACES:
+///////////////////////////////////////////////////////////////////////////////
+
+var (
+	// TraceNotFound occurs when the `carrier` passed to Extractor.JoinTrace()
+	// is valid and uncorrupted but has insufficient information to join or
+	// resume a trace.
+	TraceNotFound = errors.New("opentracing: Trace not found in Extraction carrier")
```
bensigelman: voila

propagation.go
```diff
@@ -0,0 +1,133 @@
...
+// other IPC mechanisms are also reasonable places to use an Injector.
+//
+// See Extractor and Tracer.Injector.
+type Injector interface {
+	// InjectSpan takes `span` and injects it into `carrier`. The actual type
+	// of `carrier` depends on the `format` passed to `Tracer.Injector()`.
+	//
+	// Implementations may return opentracing.InvalidCarrier or any other
+	// implementation-specific error if injection fails.
+	InjectSpan(span Span, carrier interface{}) error
```
bensigelman: Please scrutinize this comment and the one for `JoinTrace` below... and `Tracer.Injector()/Extractor()`, too.

yurishkuro:
:ship: it

btw, can someone explain why github's `shipit` is a rodent?

bensigelman:
@yurishkuro https://www.quora.com/GitHub/What-is-the-significance-of-the-Ship-It-squirrel

michaelsembwever:
The following PR demonstrates how the Carrier struct isn't required in the Specification.
https://github.com/opentracing/opentracing-java/pull/10

PR #50, opened 2016-02-10
consolidate span creation methods

bensigelman:
While we're at it, add support for Span start/finish options.

@yurishkuro I am amused by the similarity between the Start/FinishOptions and some RFC Go code you sent out in the early days of the DCP discussion late last year. I didn't actually reference it, but if memory serves this is similar.

See also:
- https://github.com/opentracing/opentracing.github.io/issues/44#issuecomment-179654948
- https://github.com/opentracing/opentracing.github.io/issues/22
- https://github.com/opentracing/opentracing.github.io/issues/20, which is basically a dup of 22 just above
- https://github.com/opentracing/opentracing.github.io/issues/43 (perhaps the start options could accomplish this via an explicit Tag)

### DIFF
--- a/span.go
+++ b/span.go
@@ -33,7 +33,7 @@ type Span interface {
 	Finish()
 	// FinishWithOptions is like Finish() but with explicit control over
 	// timestamps and log data.
-	FinishWithOptions(opts *FinishOptions)
+	FinishWithOptions(opts FinishOptions)
 
 	// LogEvent() is equivalent to
 	//
@@ -129,10 +129,21 @@ type LogData struct {
 type FinishOptions struct {
 	// FinishTime overrides the Span's finish time, or implicitly becomes
 	// time.Now() if FinishTime.IsZero().
+	//
+	// FinishTime must resolve to a timestamp that's >= the Span's StartTime
+	// (per StartSpanOptions).
 	FinishTime time.Time
 
 	// BulkLogData allows the caller to specify the contents of many Log()
 	// calls with a single slice. May be nil.
+	//
+	// None of the LogData.Timestamp values may be .IsZero() (i.e., they must
+	// be set explicitly). Also, they must be >= the Span's start timestamp and
+	// <= the FinishTime (or time.Now() if FinishTime.IsZero()). Otherwise the
+	// behavior of FinishWithOptions() is undefined.
+	//
+	// If specified, the caller hands off ownership of BulkLogData at
+	// FinishWithOptions() invocation time.
 	BulkLogData []LogData
 }
 
--- a/span.go
+++ b/span.go
@@ -6,15 +6,12 @@ import (
 	"time"
 )
 
-// Span represents an active, un-finished span in the opentracing system.
+// Span represents an active, un-finished span in the OpenTracing system.
 //
-// Spans are created by the Tracer interface and Span.StartChild.
+// Spans are created by the Tracer interface.
 type Span interface {
 	// Sets or changes the operation name.
 	SetOperationName(operationName string) Span
-
-	// Creates and starts a child span.
-	StartChild(operationName string) Span
 
 	// Adds a tag to the span.
 	//
@@ -34,6 +31,9 @@ type Span interface {
 	// Finish() should be the last call made to any span instance, and to do
 	// otherwise leads to undefined behavior.
 	Finish()
+	// FinishWithOptions is like Finish() but with explicit control over
+	// timestamps and log data.
+	FinishWithOptions(opts *FinishOptions)
 
 	// LogEvent() is equivalent to
 	//
@@ -86,6 +86,9 @@ type Span interface {
 	//
 	// See the `SetTraceAttribute` notes about `restrictedKey`.
 	TraceAttribute(restrictedKey string) string
+
+	// Provides access to the Tracer that created this Span.
+	Tracer() Tracer
 }
 
 // LogData is data associated to a Span. Every LogData instance should specify
@@ -121,6 +124,18 @@ type LogData struct {
 	Payload interface{}
 }
 
+// FinishOptions allows Span.FinishWithOptions callers to override the finish
+// timestamp and provide log data via a bulk interface.
+type FinishOptions struct {
+	// FinishTime overrides the Span's finish time, or implicitly becomes
+	// time.Now() if FinishTime.IsZero().
+	FinishTime time.Time
+
+	// BulkLogData allows the caller to specify the contents of many Log()
+	// calls with a single slice. May be nil.
+	BulkLogData []LogData
+}
+
 // Tags are a generic map from an arbitrary string key to an opaque value type.
 // The underlying tracing system is responsible for interpreting and
 // serializing the values.
@@ -144,4 +159,12 @@ func CanonicalizeTraceAttributeKey(key string) (string, bool) {
 		return "", false
 	}
 	return strings.ToLower(key), true
+}
+
+// StartChildSpan is a simple helper to start a child span given only its parent (per StartSpanOptions.Parent) and an operation name per Span.SetOperationName.
+func StartChildSpan(parent Span, operationName string) Span {
+	return parent.Tracer.StartSpanWithOptions(SpanOptions{
+		OperationName: operationName,
+		Parent:        parent,
+	})
 }
--- a/span.go
+++ b/span.go
@@ -144,7 +144,7 @@ type FinishOptions struct {
 	//
 	// If specified, the caller hands off ownership of BulkLogData at
 	// FinishWithOptions() invocation time.
-	BulkLogData []LogData
+	BulkLogData []*LogData
 }
 
 // Tags are a generic map from an arbitrary string key to an opaque value type.
@@ -174,7 +174,7 @@ func CanonicalizeTraceAttributeKey(key string) (string, bool) {
 
 // StartChildSpan is a simple helper to start a child span given only its parent (per StartSpanOptions.Parent) and an operation name per Span.SetOperationName.
 func StartChildSpan(parent Span, operationName string) Span {
-	return parent.Tracer.StartSpanWithOptions(SpanOptions{
+	return parent.Tracer().StartSpanWithOptions(StartSpanOptions{
 		OperationName: operationName,
 		Parent:        parent,
 	})
--- a/tracer.go
+++ b/tracer.go
@@ -27,7 +27,7 @@ type Tracer interface {
 	//     })
 	//
 	StartSpan(operationName string) Span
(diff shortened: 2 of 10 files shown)


span.go
```diff
@@ -6,16 +6,13 @@ import (
...
+// Span represents an active, un-finished span in the OpenTracing system.
 //
-// Spans are created by the Tracer interface and Span.StartChild.
+// Spans are created by the Tracer interface.
 type Span interface {
 	// Sets or changes the operation name.
 	SetOperationName(operationName string) Span
 
-	// Creates and starts a child span.
-	StartChild(operationName string) Span
```
bensigelman: This really doesn't need to be in `Span` as long as we are willing to have a `Span.Tracer()` accessor. Since I can't imagine a `Span` impl that doesn't have access to its `Tracer` (to handle `Finish()` if nothing else), I like this change. See the new `StartChildSpan` (global) helper method.

span.go
```diff
@@ -121,6 +124,18 @@ type LogData struct {
...
+// FinishOptions allows Span.FinishWithOptions callers to override the finish
+// timestamp and provide log data via a bulk interface.
+type FinishOptions struct {
+	// FinishTime overrides the Span's finish time, or implicitly becomes
+	// time.Now() if FinishTime.IsZero().
+	FinishTime time.Time
+
+	// BulkLogData allows the caller to specify the contents of many Log()
+	// calls with a single slice. May be nil.
+	BulkLogData []LogData
```
bensigelman: (might as well...)

span.go
```diff
@@ -145,3 +160,11 @@ func CanonicalizeTraceAttributeKey(key string) (string, bool) {
 	}
 	return strings.ToLower(key), true
 }
+
+// StartChildSpan is a simple helper to start a child span given only its parent (per StartSpanOptions.Parent) and an operation name per Span.SetOperationName.
+func StartChildSpan(parent Span, operationName string) Span {
```
bensigelman: A consolation prize since this PR gets rid of `Span.StartChild`... doesn't strictly need to be here, but it seems polite.

tracer.go
```diff
@@ -14,13 +16,36 @@ type Tracer interface {
...
+	// created. May be nil.
+	Parent Span
+
+	// StartTime overrides the Span's start time, or implicitly becomes
+	// time.Now() if StartTime.IsZero().
+	StartTime time.Time
+
+	// Zero or more Tags. The restrictions on map values are identical to those
+	// for Span.SetTag(). May be nil.
+	Tags map[string]interface{}
```
bensigelman: I believe this handles that "debug" case we'd discussed at times, @yurishkuro

tracer.go
```diff
@@ -14,13 +16,36 @@ type Tracer interface {
...
+	// OperationName may be empty (and set later via Span.SetOperationName)
+	OperationName string
+
+	// Parent may specify Span instance that caused the new (child) Span to be
+	// created. May be nil.
+	Parent Span
+
+	// StartTime overrides the Span's start time, or implicitly becomes
+	// time.Now() if StartTime.IsZero().
+	StartTime time.Time
```
bensigelman: this is something we need to add to the API somehow... otherwise there's no way to do non-realtime OpenTracing.

bensigelman:
If we like this direction, I will fix things up with the build / the ramifications in the examples / etc.

Note that I am deliberately ignoring anything touched by #48... hopefully these PRs can be considered in relative isolation.

tracer.go
tschottdorf: so this is going to be equivalent to a bunch of `SetTag` calls? What's the debug case discussion about?

span.go
tschottdorf: what are the semantics for the Span's start/finish time if conflicting LogData is inserted?

span.go
tschottdorf: Also this allows putting encoding and decoding next to each other on the
tracer.

span.go
bensigelman: I'd argue that they should be undefined at the OpenTracing level.

span.go
bensigelman: indeed, I thought about that as well. @michaelsembwever would be pleased :)

tracer.go
bensigelman: TL;DR, the idea was that the "debug" span might even be a different concrete implementation type/struct/whatever, but that would benefit from *knowing* it's "debug" prior to Span init time. Something like that. I actually don't fully understand the use case, I think it's a Zipkin thing. @yurishkuro can correct what is almost certainly my partial misunderstanding above.

tracer.go
```diff
@@ -14,13 +16,36 @@ type Tracer interface {
...
-	//     var tracer Tracer = ...
+	//     var tracer opentracing.Tracer = ...
 	//
-	//     sp := tracer.StartTrace("GetFeed")
+	//     sp := tracer.StartSpan("GetFeed")
 	//
-	//     sp := tracer.StartTrace("HandleHTTPRequest").
-	//         SetTag("user_agent", req.UserAgent).
-	//         SetTag("lucky_number", 42)
+	//     sp := tracer.StartSpanWithOptions(opentracing.SpanOptions{
```
yurishkuro: missing `&` ?

tracer.go
```diff
@@ -14,13 +16,36 @@ type Tracer interface {
...
+
+// StartSpanOptions allows Tracer.StartSpanWithOptions callers to override the
+// start timestamp, specify a parent Span, and make sure that Tags are
+// available at Span initialization time.
+type StartSpanOptions struct {
+	// OperationName may be empty (and set later via Span.SetOperationName)
+	OperationName string
+
+	// Parent may specify Span instance that caused the new (child) Span to be
+	// created. May be nil.
```
yurishkuro: `// if nil, start a new trace`  ?

span.go
tschottdorf: Assumptions about timestamps are somewhat natural, which could trip up the changes between implementations. Let's be clear about what can be assumed on LogData and the various other timestamps. I'm ok with saying that there are no guarantees whatsoever, but there should be some shared interpretation of a span that starts tomorrow, ends yesterday and has event timestamps from 23BC to 2142.

Also, unfortunately (at least with Go) ownership needs clarification: Is the slice off-limits for the caller after having been passed in?

tracer.go
yurishkuro: not really zipkin, more like a case of any tracer that uses upfront sampling. If it decides not to sample the trace and marks the span as such, it might also short-circuit saving other init data to the span, even the timestamp, since it assumes the span will never be reported. But if we pass ~~debug~~ `force_sampled=true` tag to the constructor method, tracer won't have any data loss.

tracer.go
```diff
@@ -14,13 +16,36 @@ type Tracer interface {
...
 	//
-	StartTrace(operationName string) Span
+	StartSpan(operationName string) Span
+	StartSpanWithOptions(opts *SpanOptions) Span
+}
+
+// StartSpanOptions allows Tracer.StartSpanWithOptions callers to override the
+// start timestamp, specify a parent Span, and make sure that Tags are
+// available at Span initialization time.
+type StartSpanOptions struct {
```
yurishkuro: maybe move options to `options.go`?

tracer.go
tschottdorf: actually it seems better to pass this (and any other options) by value in l. 30 below to avoid an allocation (and the question of what `nil`) means.

tracer.go
tschottdorf: Have you considered not having a start- and end time for the trace and implicitly defining these two as the lowest and highest timestamp of the timestamps of the log events in a trace? Maybe there are compelling reasons against it, but it would remove these issues completely, without any interface changes.

tracer.go
tschottdorf: curious to hear more.

Again, at least in Go, ownership question here (making maps isn't free, so if I have a fixed set of tags I'm using, I would want to use the same map in different traces, but of course that means the span must take a copy).

yurishkuro:
lgtm!

It even address the original intent of #22, which was "have some way of appending data to *existing* span", due to the name not referring to child: `StartSpanWithOptions`. 

The scenario was client->haproxy->server where haproxy emits logs somewhere, but does not have a capability to create a child span inline and modify the header on the fly. Most Zipkin backends historically allowed multi-part spans, i.e. same span ID coming from different processes and getting merged into one in the UI, along with their logs/events. In that model haproxy did not have to start a child span, it could simply attach timed logs like `haproxy-in` / `haproxy-out` to the existing span ID (and do it via a process that tails haproxy logs and uses historical API).

tracer.go
tschottdorf: Hmm, I guess there's a use case for a Span that doesn't actually contain a log but only exists to measure the duration of its lifetime.

tschottdorf:
had some comments, but generally LGTM as well.

tracer.go
bensigelman: @tschottdorf that may even be the most-common type of span for many codebases... so yeah, I think we should continue to support it. (Though it is certainly possible to think of start/finish as "special" loggable events... but I want them to be semantically required and thus feel good about this arrangement)

tracer.go
bensigelman: @tschottdorf I think you're right... I'll leave this comment as-is and change the code to expect a non-pointer.

tracer.go
bensigelman: I think I prefer to have them near their respective `...WithOptions` methods... no strong feelings but will stick with that.

tracer.go
yurishkuro: maybe we should profile it, but I suspect on-stack struct will be indeed faster, even with copying by value, than heap alloc.

bensigelman:
@tschottdorf @yurishkuro big news: this COMPILES now. And the tests and example are working, too.

PTAL...

bensigelman:
(and I've addressed the comments from the previous round)

span.go
```diff
@@ -34,6 +31,9 @@ type Span interface {
 	// Finish() should be the last call made to any span instance, and to do
 	// otherwise leads to undefined behavior.
 	Finish()
+	// FinishWithOptions is like Finish() but with explicit control over
+	// timestamps and log data.
+	FinishWithOptions(opts FinishOptions)
```
yurishkuro: for consistency with StartSpan, should the type be called FinishSpanOptions ?

yurishkuro:
one minor comment on naming, otherwise lgtm

span.go
bensigelman: I was naming the struct after the method, and `FinishSpan` seems redundant... thoughts?

yurishkuro: ah, ok.

PR #51, opened 2016-02-11
add a OT-Go semver

bensigelman:
An RFC per https://github.com/opentracing/opentracing.github.io/issues/33 (@bcronin)

### DIFF
--- a/examples/dapperish/dapper.go
+++ b/examples/dapperish/dapper.go
@@ -7,10 +7,5 @@ import (
 
 // NewTracer returns a new dapperish Tracer instance.
 func NewTracer(processName string) opentracing.Tracer {
-	return standardtracer.New(
-		NewTrivialRecorder(processName),
-		&opentracing.ImplementationID{
-			Name:    "dapperish",
-			Version: "0.1.0",
-		})
+	return standardtracer.New(NewTrivialRecorder(processName))
 }
--- a/examples/dapperish/dapper.go
+++ b/examples/dapperish/dapper.go
@@ -7,5 +7,10 @@ import (
 
 // NewTracer returns a new dapperish Tracer instance.
 func NewTracer(processName string) opentracing.Tracer {
-	return standardtracer.New(NewTrivialRecorder(processName))
+	return standardtracer.New(
+		NewTrivialRecorder(processName),
+		&opentracing.ImplementationID{
+			Name:    "dapperish",
+			Version: "0.1.0",
+		})
 }
--- a/tracer.go
+++ b/tracer.go
@@ -41,3 +41,9 @@ type ImplementationID struct {
 	// encouraged.
 	Version string
 }
+
+const (
+	// OpenTracingSemVer identifies the semantic version (http://semver.org/)
+	// of the OpenTracing Go library itself.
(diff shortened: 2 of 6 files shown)


tracer.go
```diff
@@ -23,4 +23,21 @@ type Tracer interface {
...
+
+	// ImplementationID returns information about the OpenTracing
+	// implementation backing the Tracer. The return value should not change
+	// over the life of a particular Tracer instance.
+	ImplementationID() *ImplementationID
+}
+
+// ImplementationID is a simple, extensible struct that describes an
+// OpenTracing implementation.
+type ImplementationID struct {
```
bensigelman: (the only interesting part of this change)

bensigelman:
This intentionally does not handle any notion of `Capabilities`... Upon reflection I'd like to see the APIs used more in the wild first.

tracer.go
```diff
@@ -23,4 +23,21 @@ type Tracer interface {
...
+	// The (stable) name of the implementation. E.g., "Dapper" or "Zipkin". The
+	// Name should not reflect the host language or platform.
+	Name string
+
+	// Version may take any form, but SemVer (http://semver.org/) is strongly
+	// encouraged.
+	Version string
```
bcronin: @bensigelman , this looks pretty good, but Go seems a bit like the exceptional case with regards to the utility of a version string (since everything in Go is basically statically bound). I'm curious what this might look like for Python. I'm specifically considering the case where the *supported version of the OpenTracing standard* is what's of interest rather than the underlying *implementation version*. E.g. if an arbitrary piece of Python code is handed a `tracer` object, how can it (cleanly) check the supported version of OpenTracing on that object?

(By the way I believe I'm safe in assuming that in the above `Version` is intended to represent the implementation version given the struct name and the "may take any form" comment.)

yurishkuro: I agree with @bcronin, it's the OpenTracing version that the original ticket was asking for. We said it would be best done via Capabilities API rather than a version. At the moment, however, there doesn't seem to be any aspects of the API that are optional and require introspection by the client code.

bensigelman:
@bcronin @yurishkuro roger that... makes sense. I added a `const` here for the Go OpenTracing version, and I sent out https://github.com/opentracing/opentracing-python/pull/16 for python...

examples/dapperish/dapper.go
```diff
@@ -7,5 +7,10 @@ import (
 
 // NewTracer returns a new dapperish Tracer instance.
 func NewTracer(processName string) opentracing.Tracer {
-	return standardtracer.New(NewTrivialRecorder(processName))
+	return standardtracer.New(
+		NewTrivialRecorder(processName),
+		&opentracing.ImplementationID{
```
tschottdorf: you're explicitly letting tracers return `nil` here? Doesn't seem like a bad idea to force tracers to pick a name and version.

bensigelman:
Per https://github.com/opentracing/opentracing-python/pull/16, I'm pulling this back to be a semver for the Go API and nothing else.

bensigelman:
(withdrawing this per opentracing/opentracing.github.io#33)

PR #52, opened 2016-02-14
Import opentracing-go as opentracing

tschottdorf:
The implicit import name makes resolution easier for
tools like `godef`, which at the time of writing do
not have the capability to resolve nonstandard
import paths, for example see

    https://github.com/rogpeppe/godef/issues/18

I looked into adding it for `godef`, but it did not
seem trivial.


bensigelman:
shame on `godef`, but LGTM.

PR #53, opened 2016-02-14
Avoid allocations, fix CI

tschottdorf:
Embed StandardContext (not *StandardContext) into
RawSpan and initialize the attribute and tags map
only on the first write (it is valid to read from
a nil map).

Move the trace attributes from StandardContext onto
spanImpl; they were not exposed to the recorder, and
the lock on StandardContext required that object to
always be heap allocated. If they need to be exposed
to the user, they can be moved onto RawSpan (or even
back onto StandardContext, but without the lock).
Locking is instead synchronized through the Span's
log.

Work towards sane allocation behavior and add some
benchmarks to that effect: Span creation and destruction
with calls to LogEvent, SetTag oder SetTraceAttribute
now don't allocate.

Changed various other pointers to be passed by-value.

Fix various linter errors and the invocation in
the Makefile (which did not exit nonzero on issues).

Use a sync.Pool instead of new allocations …
This reduces the allocs for a common Span to zero.

```
name               old time/op    new time/op    delta
Span_Empty-8          411ns ± 0%     337ns ± 0%   ~     (p=1.000 n=1+1)
Span_100Events-8     18.2µs ± 0%    18.6µs ± 0%   ~     (p=1.000 n=1+1)
Span_1000Events-8    18.2µs ± 0%    19.3µs ± 0%   ~     (p=1.000 n=1+1)
Span_100Tags-8       25.1µs ± 0%    26.2µs ± 0%   ~     (p=1.000 n=1+1)
Span_1000Tags-8      24.7µs ± 0%    24.9µs ± 0%   ~     (p=1.000 n=1+1)

name               old alloc/op   new alloc/op   delta
Span_Empty-8           144B ± 0%       0B ±NaN%   ~     (p=1.000 n=1+1)
Span_100Events-8     16.2kB ± 0%    16.1kB ± 0%   ~     (p=1.000 n=1+1)
Span_1000Events-8    16.2kB ± 0%    16.1kB ± 0%   ~     (p=1.000 n=1+1)
Span_100Tags-8       10.7kB ± 0%    10.5kB ± 0%   ~     (p=1.000 n=1+1)
Span_1000Tags-8      10.7kB ± 0%    10.5kB ± 0%   ~     (p=1.000 n=1+1)

name               old allocs/op  new allocs/op  delta
Span_Empty-8           1.00 ± 0%     0.00 ±NaN%   ~     (p=1.000 n=1+1)
Span_100Events-8       9.00 ± 0%      8.00 ± 0%   ~     (p=1.000 n=1+1)
Span_1000Events-8      9.00 ± 0%      8.00 ± 0%   ~     (p=1.000 n=1+1)
Span_100Tags-8         12.0 ± 0%      11.0 ± 0%   ~     (p=1.000 n=1+1)
Span_1000Tags-8        12.0 ± 0%      11.0 ± 0%   ~     (p=1.000 n=1+1)
```

FYI, I'm piling more commits on top of this, but would be good to get this
in/discuss it before it gets more refined.

### DIFF
--- a/propagation.go
+++ b/propagation.go
@@ -7,19 +7,19 @@ import "errors"
 ///////////////////////////////////////////////////////////////////////////////
 
 var (
-	// TraceNotFound occurs when the `carrier` passed to Extractor.JoinTrace()
+	// ErrTraceNotFound occurs when the `carrier` passed to Extractor.JoinTrace()
 	// is valid and uncorrupted but has insufficient information to join or
 	// resume a trace.
-	TraceNotFound = errors.New("opentracing: Trace not found in Extraction carrier")
+	ErrTraceNotFound = errors.New("opentracing: Trace not found in Extraction carrier")
 
-	// InvalidCarrier errors occur when Injector.InjectSpan() or
+	// ErrInvalidCarrier errors occur when Injector.InjectSpan() or
 	// Extractor.JoinTrace() implementations expect a different type of
 	// `carrier` than they are given.
-	InvalidCarrier = errors.New("opentracing: Invalid Injection/Extraction carrier")
+	ErrInvalidCarrier = errors.New("opentracing: Invalid Injection/Extraction carrier")
 
-	// TraceCorrupted occurs when the `carrier` passed to Extractor.JoinTrace()
+	// ErrTraceCorrupted occurs when the `carrier` passed to Extractor.JoinTrace()
 	// is of the expected type but is corrupted.
-	TraceCorrupted = errors.New("opentracing: Trace data corrupted in Extraction carrier")
+	ErrTraceCorrupted = errors.New("opentracing: Trace data corrupted in Extraction carrier")
 )
 
 // Injector is responsible for injecting Span instances in a manner suitable
@@ -102,6 +102,7 @@ type SplitTextCarrier struct {
 	TraceAttributes map[string]string
 }
 
+// NewSplitTextCarrier creates a new SplitTextCarrier.
 func NewSplitTextCarrier() *SplitTextCarrier {
 	return &SplitTextCarrier{
 		TracerState:     make(map[string]string),
@@ -125,6 +126,7 @@ type SplitBinaryCarrier struct {
 	TraceAttributes []byte
 }
 
+// NewSplitBinaryCarrier creates a new SplitTextCarrier.
 func NewSplitBinaryCarrier() *SplitBinaryCarrier {
 	return &SplitBinaryCarrier{
 		TracerState:     []byte{},
--- a/span.go
+++ b/span.go
@@ -146,7 +146,7 @@ type FinishOptions struct {
 	//
 	// If specified, the caller hands off ownership of BulkLogData at
 	// FinishWithOptions() invocation time.
-	BulkLogData []*LogData
+	BulkLogData []LogData
 }
 
 // Tags are a generic map from an arbitrary string key to an opaque value type.
--- a/standardtracer/bench_test.go
+++ b/standardtracer/bench_test.go
@@ -1,0 +1,65 @@
+package standardtracer
+
+import (
+	"fmt"
+	"sync/atomic"
+	"testing"
+)
+
+var tags []string
+
+func init() {
+	tags = make([]string, 1000)
+	for j := 0; j < len(tags); j++ {
+		tags[j] = fmt.Sprintf("%d", randomID())
+	}
+}
+
+type countingRecorder int32
+
+func (c *countingRecorder) RecordSpan(r RawSpan) {
+	atomic.AddInt32((*int32)(c), 1)
+}
+
+func benchmarkWithOps(b *testing.B, numEvent, numTag, numAttr int) {
+	var r countingRecorder
+	t := New(&r)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		sp := t.StartSpan("test")
+		for j := 0; j < numEvent; j++ {
+			sp.LogEvent("event")
+		}
+		for j := 0; j < numTag; j++ {
+			sp.SetTag(tags[j], nil)
+		}
+		for j := 0; j < numAttr; j++ {
+			sp.SetTraceAttribute(tags[j], tags[j])
+		}
+		sp.Finish()
+	}
+	b.StopTimer()
+	if int(r) != b.N {
+		b.Fatalf("missing traces: expected %d, got %d", b.N, r)
+	}
+}
+
+func BenchmarkSpan_Empty(b *testing.B) {
+	benchmarkWithOps(b, 0, 0, 0)
+}
+
+func BenchmarkSpan_100Events(b *testing.B) {
+	benchmarkWithOps(b, 100, 0, 0)
+}
+
+func BenchmarkSpan_1000Events(b *testing.B) {
+	benchmarkWithOps(b, 100, 0, 0)
+}
+
+func BenchmarkSpan_100Tags(b *testing.B) {
+	benchmarkWithOps(b, 0, 100, 0)
+}
+
+func BenchmarkSpan_1000Tags(b *testing.B) {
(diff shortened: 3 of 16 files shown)


propagation.go
```diff
@@ -7,19 +7,19 @@ import "errors"
 ///////////////////////////////////////////////////////////////////////////////
 
 var (
-	// TraceNotFound occurs when the `carrier` passed to Extractor.JoinTrace()
+	// ErrTraceNotFound occurs when the `carrier` passed to Extractor.JoinTrace()
 	// is valid and uncorrupted but has insufficient information to join or
 	// resume a trace.
-	TraceNotFound = errors.New("opentracing: Trace not found in Extraction carrier")
+	ErrTraceNotFound = errors.New("opentracing: Trace not found in Extraction carrier")
```
bensigelman: thanks for the renames

standardtracer/span.go
```diff
@@ -11,22 +11,33 @@ import (
 // Implements the `Span` interface. Created via tracerImpl (see
 // `standardtracer.New()`).
 type spanImpl struct {
-	lock     sync.Mutex
-	tracer   *tracerImpl
-	recorder SpanRecorder
-	raw      RawSpan
+	tracer     *tracerImpl
+	sync.Mutex // protects the fields below
```
bensigelman: is there a reason you went for Mutex rather than RWMutex as before?

standardtracer/span.go
```diff
@@ -11,22 +11,33 @@ import (
...
+	raw        RawSpan
+	// TODO(tschottdorf): should this be available to the Recorder
```
bensigelman: the recorder should have access to the trace attrs

span.go
```diff
@@ -146,7 +146,7 @@ type FinishOptions struct {
 	//
 	// If specified, the caller hands off ownership of BulkLogData at
 	// FinishWithOptions() invocation time.
-	BulkLogData []*LogData
+	BulkLogData []LogData
```
yurishkuro: @bensigelman ^^^

standardtracer/span.go
tschottdorf: `spanImpl` had a `Mutex` even before, and generally I would only use an `RWMutex` when it's indicated by a specific workload (since it has generally higher overhead). I would naively expect a trace to be "mostly" written to, so that a `Mutex` seems the better fit.

standardtracer/span.go
tschottdorf: Ok. Don't think that was the case before, but I can fix it. I'll put it on RawSpan if that's ok.

propagation.go
tschottdorf: thank `golint`, not me :-)

standardtracer/span.go
```diff
@@ -45,14 +56,14 @@ func (s *spanImpl) LogEventWithPayload(event string, payload interface{}) {
...
-	defer s.lock.Unlock()
+	s.Lock()
+	defer s.Unlock()
 
 	if ld.Timestamp.IsZero() {
 		ld.Timestamp = time.Now()
 	}
 
-	s.raw.Logs = append(s.raw.Logs, &ld)
+	s.raw.Logs = append(s.raw.Logs, ld)
```
bensigelman: for large LogData instances I'm surprised that this is a perf improvement! but that's what benchmarks are for, I guess.

standardtracer/bench_test.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+}
+
+type countingRecorder int32
+
+func (c *countingRecorder) RecordSpan(r RawSpan) {
+	atomic.AddInt32((*int32)(c), 1)
+}
+
+func benchmarkWithOps(b *testing.B, numEvent, numTag, numAttr int) {
```
bensigelman: thanks for adding this

standardtracer/span.go
bensigelman: I'm fine with the decision, I was just curious if it came from a specific benchmark or more of an ad hoc hypothesis (sounds like the latter, and that's fine with me at this point). The RWMutex was there for the Trace Attributes specifically since I thought there might be more reads than writes there, but absent a contention profile I'm fine keeping it simple.

standardtracer/span.go
tschottdorf: `LogData` is always small: a timestamp and two pointers.

standardtracer/bench_test.go
tschottdorf: I'm adding corresponding ones for the serializers right now.

standardtracer/span.go
yurishkuro: I suppose because LogData is a timestamp and 2 pointers, faster to copy it than to allocate on heap. I.e. the speed improvement is probably coming from the points where LogData is created.

span.go
bensigelman: Yes, I saw this... it's the only "interesting" (non-lint) change to the public API. I don't know whether it's reasonable to assume that the caller will always know how many `LogData`s they want to add at once. Nor do I know whether the above actually matters... It looks like @tschottdorf's benchmarks suggest the allocations are worse than the copying in the `RawSpan` case at least.

bensigelman:
@tschottdorf the standardtracer stuff is not optimized (or at least it was not optimized) and I appreciate you taking the time to profile and fix it up. https://github.com/opentracing/opentracing-go/pull/53#discussion-diff-52850024 is the only concern I have... I wish we had a more realistic real-world benchmark. I'm okay proceeding with the non-pointer `[]LogData`, though, unless @yurishkuro or others (@slimsag?) with appreciable Go experience have words of warning to share on this front.

Basically LGTM, though, and thanks again!

standardtracer/span.go
bensigelman: @tschottdorf @yurishkuro per https://github.com/opentracing/opentracing.github.io/issues/30, the LogData is going to include more things in the future (standard debug-logging fileds, namely).

standardtracer/span.go
bensigelman: fine by me

standardtracer/bench_test.go
bensigelman: Thanks in advance for those, then.

I didn't want to do profiling and optimization until the API had stabilized... I guess we're there (or quite nearly there), so hopefully this isn't a waste of your time!

standardtracer/span.go
tschottdorf: Yeah, no specific benchmark. I had two mutexes here and needed to decide for one.

standardtracer/span.go
tschottdorf: like what? The only thing that would ruin it are many large fields. If we're adding more pointers (strings, ...) then the actual struct should remain fairly small.

span.go
tschottdorf: > I don't know whether it's reasonable to assume that the caller will always know how many LogDatas they want to add at once.

Might be missing something, but how does that change change the game here?

tschottdorf:
> I'm okay proceeding with the non-pointer []LogData, though, unless @yurishkuro or others (@slimsag?) with appreciable Go experience have words of warning to share on this front.

I'd be surprised if the pointer were better (we could always change it back). It's definitely worse in that unless you're tricky, each `LogEntry` is a single allocation in the previous version.

standardtracer/span.go
tschottdorf: done.

standardtracer/bench_test.go
tschottdorf: I realize we're not quite there yet, but hopefully close to the final version. Benchmarking can expose flaws in the API as well, and I figured you'd be busy enough with all the fallout from stabilization, so I went to give Go a head start.

yurishkuro:
lgtm

I would suggest merging incrementally than piling more commits onto this pr.

tschottdorf:
Yep, I'm already holding off.

PR #54, opened 2016-02-15
Add benchmarks for serialization, and make some improvements

tschottdorf:
Pick some (but not all) low hanging fruit. For example, the encoding
in `github.com/cockroachdb/cockroach/util/encoding` has primitives
better suited to providing an efficient wire encoding.

```
name                                 old time/op    new time/op    delta
Inject_SplitBinary_Empty-8              792ns ± 1%     825ns ± 4%   +4.26%        (p=0.006 n=10+10)
Inject_SplitBinary_100Attributes-8     98.6µs ± 2%    38.0µs ± 2%  -61.48%        (p=0.000 n=10+10)
Extract_SplitBinary_Empty-8             882ns ± 1%     885ns ± 1%     ~            (p=0.287 n=9+10)
Extract_SplitBinary_100Attributes-8     155µs ± 1%      67µs ± 4%  -56.87%        (p=0.000 n=10+10)

name                                 old alloc/op   new alloc/op   delta
Inject_SplitBinary_Empty-8               352B ± 0%      352B ± 0%     ~     (all samples are equal)
Inject_SplitBinary_100Attributes-8     39.9kB ± 0%     6.8kB ± 0%  -83.09%        (p=0.000 n=10+10)
Extract_SplitBinary_Empty-8              224B ± 0%      224B ± 0%     ~     (all samples are equal)
Extract_SplitBinary_100Attributes-8    37.7kB ± 0%    22.5kB ± 0%  -40.38%        (p=0.000 n=10+10)

name                                 old allocs/op  new allocs/op  delta
Inject_SplitBinary_Empty-8               10.0 ± 0%      10.0 ± 0%     ~     (all samples are equal)
Inject_SplitBinary_100Attributes-8      1.22k ± 0%     0.41k ± 0%  -66.28%        (p=0.000 n=10+10)
Extract_SplitBinary_Empty-8              10.0 ± 0%      10.0 ± 0%     ~     (all samples are equal)
Extract_SplitBinary_100Attributes-8     1.21k ± 0%     0.62k ± 0%  -49.09%        (p=0.000 n=10+10)
```

### DIFF
--- a/propagation.go
+++ b/propagation.go
@@ -91,6 +91,10 @@ const (
 // The Span is separated in this way for a variety of reasons; the most
 // important is to give OpenTracing users a portable way to opt out of Trace
 // Attribute propagation entirely if they deem it a stability risk.
+//
+// It is legal to provide one or both maps as `nil`; they will be created
+// as needed. If non-nil maps are provided, they will be used without
+// clearing them out on injection.
 type SplitTextCarrier struct {
 	// TracerState is Tracer-specific context that must cross process
 	// boundaries. For example, in Dapper this would include a trace_id, a
@@ -104,10 +108,7 @@ type SplitTextCarrier struct {
 
 // NewSplitTextCarrier creates a new SplitTextCarrier.
 func NewSplitTextCarrier() *SplitTextCarrier {
-	return &SplitTextCarrier{
-		TracerState:     make(map[string]string),
-		TraceAttributes: make(map[string]string),
-	}
+	return &SplitTextCarrier{}
 }
 
 // SplitBinaryCarrier breaks a propagated Span into two pieces.
@@ -115,6 +116,9 @@ func NewSplitTextCarrier() *SplitTextCarrier {
 // The Span is separated in this way for a variety of reasons; the most
 // important is to give OpenTracing users a portable way to opt out of Trace
 // Attribute propagation entirely if they deem it a stability risk.
+//
+// Both byte slices may be nil; on injection, what is provided will be cleared
+// and the resulting capacity used.
 type SplitBinaryCarrier struct {
 	// TracerState is Tracer-specific context that must cross process
 	// boundaries. For example, in Dapper this would include a trace_id, a
@@ -128,8 +132,5 @@ type SplitBinaryCarrier struct {
 
 // NewSplitBinaryCarrier creates a new SplitTextCarrier.
 func NewSplitBinaryCarrier() *SplitBinaryCarrier {
-	return &SplitBinaryCarrier{
-		TracerState:     []byte{},
-		TraceAttributes: []byte{},
-	}
+	return &SplitBinaryCarrier{}
 }
--- a/propagation.go
+++ b/propagation.go
@@ -104,7 +104,6 @@ type SplitTextCarrier struct {
 
 // NewSplitTextCarrier creates a new SplitTextCarrier.
 func NewSplitTextCarrier() *SplitTextCarrier {
-	// TODO(tschottdorf): a global sync.Pool can help reduce allocations.
 	return &SplitTextCarrier{
 		TracerState:     make(map[string]string),
 		TraceAttributes: make(map[string]string),
--- a/propagation.go
+++ b/propagation.go
@@ -104,6 +104,7 @@ type SplitTextCarrier struct {
 
 // NewSplitTextCarrier creates a new SplitTextCarrier.
 func NewSplitTextCarrier() *SplitTextCarrier {
+	// TODO(tschottdorf): a global sync.Pool can help reduce allocations.
 	return &SplitTextCarrier{
 		TracerState:     make(map[string]string),
 		TraceAttributes: make(map[string]string),
--- a/standardtracer/propagation.go
+++ b/standardtracer/propagation.go
@@ -43,9 +43,11 @@ func (p *splitTextPropagator) InjectSpan(
 		fieldNameSampled: strconv.FormatBool(sc.raw.Sampled),
 	}
 	sc.Lock()
-	splitTextCarrier.TraceAttributes = make(map[string]string, len(sc.raw.Attributes))
-	for k, v := range sc.raw.Attributes {
-		splitTextCarrier.TraceAttributes[k] = v
+	if l := len(sc.raw.Attributes); l > 0 {
+		splitTextCarrier.TraceAttributes = make(map[string]string, l)
+		for k, v := range sc.raw.Attributes {
+			splitTextCarrier.TraceAttributes[k] = v
+		}
 	}
 	sc.Unlock()
 	return nil
@@ -201,31 +203,35 @@ func (p *splitBinaryPropagator) JoinTrace(
 		return nil, opentracing.ErrTraceCorrupted
 	}
 	iNumAttrs := int(numAttrs)
-	attrMap := make(map[string]string, iNumAttrs)
-	for i := 0; i < iNumAttrs; i++ {
-		var keyLen int32
-		err = binary.Read(attrsReader, binary.BigEndian, &keyLen)
-		if err != nil {
-			return nil, opentracing.ErrTraceCorrupted
-		}
-		keyBytes := make([]byte, keyLen)
-		err = binary.Read(attrsReader, binary.BigEndian, &keyBytes)
-		if err != nil {
-			return nil, opentracing.ErrTraceCorrupted
-		}
+	var attrMap map[string]string
+	if iNumAttrs > 0 {
+		attrMap = make(map[string]string, iNumAttrs)
+		for i := 0; i < iNumAttrs; i++ {
+			var keyLen int32
+			err = binary.Read(attrsReader, binary.BigEndian, &keyLen)
+			if err != nil {
+				return nil, opentracing.ErrTraceCorrupted
+			}
+			keyBytes := make([]byte, keyLen)
+			err = binary.Read(attrsReader, binary.BigEndian, &keyBytes)
+			if err != nil {
+				return nil, opentracing.ErrTraceCorrupted
+			}
 
-		var valLen int32
-		err = binary.Read(attrsReader, binary.BigEndian, &valLen)
-		if err != nil {
-			return nil, opentracing.ErrTraceCorrupted
+			var valLen int32
+			err = binary.Read(attrsReader, binary.BigEndian, &valLen)
+			if err != nil {
+				return nil, opentracing.ErrTraceCorrupted
+			}
+			valBytes := make([]byte, valLen)
+			err = binary.Read(attrsReader, binary.BigEndian, &valBytes)
+			if err != nil {
+				return nil, opentracing.ErrTraceCorrupted
+			}
+			// TODO(tschottdorf): should be able to convert the byte slice
+			// to string via unsafe.
+			attrMap[string(keyBytes)] = string(valBytes)
 		}
-		valBytes := make([]byte, valLen)
-		err = binary.Read(attrsReader, binary.BigEndian, &valBytes)
-		if err != nil {
-			return nil, opentracing.ErrTraceCorrupted
-		}
-
-		attrMap[string(keyBytes)] = string(valBytes)
 	}
 
 	sp := p.tracer.getSpan()
--- a/standardtracer/propagation.go
+++ b/standardtracer/propagation.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/base64"
 	"encoding/binary"
+	"io"
 	"net/http"
 	"strconv"
 	"strings"
@@ -129,7 +130,7 @@ func (p *splitBinaryPropagator) InjectSpan(
 	}
 
 	// Handle the trace and span ids, and sampled status.
-	contextBuf := new(bytes.Buffer)
+	contextBuf := bytes.NewBuffer(splitBinaryCarrier.TracerState[:0])
 	err = binary.Write(contextBuf, binary.BigEndian, sc.raw.TraceID)
 	if err != nil {
 		return err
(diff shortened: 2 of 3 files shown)


standardtracer/propagation.go
```diff
@@ -44,9 +45,11 @@ func (p *splitTextPropagator) InjectSpan(
 		fieldNameSampled: strconv.FormatBool(sc.raw.Sampled),
```
yurishkuro: on line 37 (github won't let me comment there):
```go
sc := sp.(*spanImpl)
```
Maybe there is another type of std error the injectors could return, i.e.
```go
sc, ok := sp.(*spanImpl)
if !ok {
  return ErrWrongSpanType
}
```

standardtracer/propagation.go
```diff
@@ -44,9 +45,11 @@ func (p *splitTextPropagator) InjectSpan(
...
 	}
 	sc.Lock()
-	splitTextCarrier.TraceAttributes = make(map[string]string, len(sc.raw.Attributes))
-	for k, v := range sc.raw.Attributes {
-		splitTextCarrier.TraceAttributes[k] = v
+	if l := len(sc.raw.Attributes); l > 0 {
+		splitTextCarrier.TraceAttributes = make(map[string]string, l)
```
yurishkuro: This doesn't look right - what if that map has already been initialized with a reference to request.Header, for example?

tschottdorf: Not sure I understand. The only issue here is maybe passing in a nonempty carrier (so that old data survives a new injection). Is that what you mean?

tschottdorf: I think `NewSplitTextCarrier` should return a struct with `nil` maps so that they can be initialized on first write. Then we can set them to `nil` here if `l == 0`.

standardtracer/propagation.go
tschottdorf: SGTM, but that doesn't belong in this PR.

standardtracer/propagation.go
yurishkuro: I was thinking of this scenario:
```go
carrier := &SplitTextCarrier{TracerState: request.Header, TraceAttributes: requestHeader}
injector.inject(span, carrier)
```
I.e. it's not the job of the injector to allocate memory for target maps, and in this case I am not allocating any memory at all, only using already available `request.Header` map.

tschottdorf: I think we have to decide what the semantics should be. I want to avoid a mandatory allocation if not using attributes at all, so I want the `nil` map to be valid for the carrier, which means that memory must be allocated sometimes (or the attributes dropped, which we probably don't want). The sane thing seems to be:

* if `nil`, create one
* if not `nil`, loop through and delete everything, then use it.

Or is there an explicit use case for adding to a nonempty map? I seems like that would invite surprises.

yurishkuro: I think
* if `nil`, create one (if needed, may not actually have any trace attributes)
* if not `nil`, append to provided map

The scenario with `http.Header` is exactly where you need the 2nd case, the tracer only needs to add its data to the headers, not to blank out the existing headers.

tschottdorf: There's no existing code that does this, at least not in the Go code, but I think that's because it's not possible (request headers aren't exposed as a map). It makes some sense for languages like Python, so it seems ok to define it that way.

standardtracer/propagation.go
```diff
@@ -177,52 +181,48 @@ func (p *splitBinaryPropagator) JoinTrace(
...
 	var sampledByte byte
 
-	err = binary.Read(contextReader, binary.BigEndian, &traceID)
-	if err != nil {
+	if err := binary.Read(contextReader, binary.BigEndian, &traceID); err != nil {
 		return nil, opentracing.ErrTraceCorrupted
 	}
-	err = binary.Read(contextReader, binary.BigEndian, &propagatedSpanID)
-	if err != nil {
+	if err := binary.Read(contextReader, binary.BigEndian, &propagatedSpanID); err != nil {
```
yurishkuro: if we decide to keep standardtracer around (and I am tempted, seeing the great work @tschottdorf is doing), this propagation logic should be in some other package, because it makes explicit assumption about the shape of tracer context being `(trace-id, span-id, flags)`, which is Dapper-ish, but not Zipkin-like.

standardtracer/propagation.go
yurishkuro: Header is a map, but you're right it's the wrong map, with multiple values per key:
```
type Header map[string][]string
```
But Go native http is not the only protocol, other protocols may have headers as plain map, so I think the principle holds.

I actually wanted one of the standard injector carriers to be an interface like
```
type SplitTextCarrier interface {
    SetTracerContextField(key, value string)
    SetTraceAttributeField(key, value string)
}
```

standardtracer/propagation.go
tschottdorf: What's the original plan for `standardtracer`? I thought it was going to be moved out of this package into its own (which is fine).

I don't understand your suggestion - surely `standardtracer` would make some assumptions about itself (looks like it's being dapperish), and so its serialization logic would as well). Are you suggesting pulling out some of the encoding logic into a more reusable form for other types of tracers?

standardtracer/propagation.go
tschottdorf: Seems easy enough to do, but wouldn't wanna muddle this up in this PR. I've added a commit that defines and implements the behavior you suggested (only for `SplitText`; for binary, I don't think it makes as much sense and I'm reusing the buffers).

standardtracer/propagation.go
yurishkuro: the original plan for `standardtracer` was to serve as an abstract base for real tracers, which would be responsible for providing their own trace context serialization and reporter. As you can see, there is a lot of logic in the standardtracer that deals with the main API implementation without knowing how the context is serialized or spans are reported. Essentially, it was a Service Provider Interface.

But considering how many disagreements we went through about the main OpenTracing API, maintaining a shared, reusable base implementation was a bit too much at that point.

I think at this time we just want to treat it as a reference implementation, and as such it does need to make assumption about trace context serialized format. Although it doesn't mean it cannot be made easily swappable should people decide to base custom implementations on this reference implementation.

tschottdorf: I see. Keeping an opinionated and performant reference implementation sounds good to me, and I think if we make the encoding routines available in a nested package, a lot is gained for other implementations as well.

yurishkuro: Yep. Ben used to have the concept of a TraceContextSource that was abstracting encoding from standard tracer. So it's not so much a nested package we need, as just an intermediate provider of injectors/extractors that the standard tracer would delegate to, and then derived implementations can just replace the provider (and reporter).

bensigelman: Very briefly, the idea with `standardtracer` was to try and provide a reasonable / generic in-memory representation of an active / extant Span while coupling as little as possible to a particular tracing data model. When we removed TraceContext/TraceContextSource from OT I caved and made `standardtracer` way less general. In so doing it became rather pointless from a practical perspective (if you ask me).

As @yurishkuro said, @tschottdorf is clearly doing great work here so it seems a shame to throw `standardtracer` out! One possibility would be to re-introduce the TraceContext-like abstractions and make it more general while keeping most of the performance improvements here. Another possibility would be to lay the groundwork for a lightweight, OpenTracing-native multilingual tracing *system* that integrates with all OT platforms and provides a basic but useful in-memory-only backend. A third possibility would be to leave it more or less as-is.

tschottdorf:
Rebased.

```
name                                           old time/op    new time/op    delta
Span_Empty-8                                      283ns ± 6%     276ns ± 6%     ~            (p=0.217 n=3+10)
Span_100Events-8                                 18.9µs ± 2%    18.4µs ± 1%     ~             (p=0.085 n=3+8)
Span_1000Events-8                                20.5µs ± 6%    18.3µs ± 1%  -10.91%          (p=0.009 n=3+9)
Span_100Tags-8                                   26.0µs ± 3%    26.0µs ± 4%     ~            (p=0.776 n=3+10)
Span_1000Tags-8                                  25.8µs ± 0%    26.0µs ± 5%     ~            (p=0.811 n=3+10)
Span_100Attributes-8                              120µs ± 2%     119µs ± 5%     ~            (p=0.573 n=3+10)
TrimmedSpan_100Events_100Tags_100Attributes-8     125µs ± 7%     122µs ± 5%     ~            (p=0.692 n=3+10)
Inject_SplitText_Empty-8                          590ns ± 4%     328ns ± 6%  -44.51%         (p=0.007 n=3+10)
Inject_SplitText_100Attributes-8                 10.9µs ± 5%     7.7µs ± 5%  -29.41%         (p=0.007 n=3+10)
Inject_SplitBinary_Empty-8                        835ns ± 3%     839ns ± 1%     ~            (p=0.434 n=3+10)
Inject_SplitBinary_100Attributes-8                103µs ± 1%      40µs ± 5%  -61.18%         (p=0.007 n=3+10)
Extract_SplitText_Empty-8                         736ns ± 1%     697ns ± 7%     ~            (p=0.077 n=3+10)
Extract_SplitText_100Attributes-8                 750ns ± 2%     712ns ± 7%     ~            (p=0.084 n=3+10)
Extract_SplitBinary_Empty-8                       852ns ± 2%     876ns ± 2%     ~            (p=0.077 n=3+10)
Extract_SplitBinary_100Attributes-8               160µs ± 1%      71µs ± 2%  -55.59%         (p=0.007 n=3+10)

name                                           old alloc/op   new alloc/op   delta
Span_Empty-8                                     0.00B ±NaN%    0.00B ±NaN%     ~     (all samples are equal)
Span_100Events-8                                 16.1kB ± 0%    16.1kB ± 0%     ~     (all samples are equal)
Span_1000Events-8                                16.1kB ± 0%    16.1kB ± 0%     ~     (all samples are equal)
Span_100Tags-8                                   10.5kB ± 0%    10.5kB ± 0%     ~            (p=0.455 n=3+10)
Span_1000Tags-8                                  10.5kB ± 0%    10.5kB ± 0%     ~            (p=1.021 n=3+10)
Span_100Attributes-8                             10.5kB ± 0%    10.5kB ± 0%   -0.06%         (p=0.042 n=3+10)
TrimmedSpan_100Events_100Tags_100Attributes-8    0.00B ±NaN%    0.00B ±NaN%     ~     (all samples are equal)
Inject_SplitText_Empty-8                           400B ± 0%       64B ± 0%  -84.00%         (p=0.000 n=3+10)
Inject_SplitText_100Attributes-8                 5.85kB ± 0%    0.06kB ± 0%  -98.91%         (p=0.000 n=3+10)
Inject_SplitBinary_Empty-8                         352B ± 0%      352B ± 0%     ~     (all samples are equal)
Inject_SplitBinary_100Attributes-8               39.9kB ± 0%     6.8kB ± 0%  -83.09%         (p=0.000 n=3+10)
Extract_SplitText_Empty-8                        0.00B ±NaN%    0.00B ±NaN%     ~     (all samples are equal)
Extract_SplitText_100Attributes-8                0.00B ±NaN%    0.00B ±NaN%     ~     (all samples are equal)
Extract_SplitBinary_Empty-8                        224B ± 0%      224B ± 0%     ~     (all samples are equal)
Extract_SplitBinary_100Attributes-8              37.7kB ± 0%    22.4kB ± 0%  -40.47%         (p=0.007 n=3+10)

name                                           old allocs/op  new allocs/op  delta
Span_Empty-8                                      0.00 ±NaN%     0.00 ±NaN%     ~     (all samples are equal)
Span_100Events-8                                   8.00 ± 0%      8.00 ± 0%     ~     (all samples are equal)
Span_1000Events-8                                  8.00 ± 0%      8.00 ± 0%     ~     (all samples are equal)
Span_100Tags-8                                     11.0 ± 0%      11.0 ± 0%     ~     (all samples are equal)
Span_1000Tags-8                                    11.0 ± 0%      11.0 ± 0%     ~     (all samples are equal)
Span_100Attributes-8                               11.0 ± 0%      11.0 ± 0%     ~     (all samples are equal)
TrimmedSpan_100Events_100Tags_100Attributes-8     0.00 ±NaN%     0.00 ±NaN%     ~     (all samples are equal)
Inject_SplitText_Empty-8                           4.00 ± 0%      2.00 ± 0%  -50.00%         (p=0.000 n=3+10)
Inject_SplitText_100Attributes-8                   8.00 ± 0%      2.00 ± 0%  -75.00%         (p=0.000 n=3+10)
Inject_SplitBinary_Empty-8                         10.0 ± 0%      10.0 ± 0%     ~     (all samples are equal)
Inject_SplitBinary_100Attributes-8                1.22k ± 0%     0.41k ± 0%  -66.28%         (p=0.000 n=3+10)
Extract_SplitText_Empty-8                         0.00 ±NaN%     0.00 ±NaN%     ~     (all samples are equal)
Extract_SplitText_100Attributes-8                 0.00 ±NaN%     0.00 ±NaN%     ~     (all samples are equal)
Extract_SplitBinary_Empty-8                        10.0 ± 0%      10.0 ± 0%     ~     (all samples are equal)
Extract_SplitBinary_100Attributes-8               1.21k ± 0%     0.62k ± 0%  -49.09%         (p=0.000 n=3+10)
```

propagation.go
```diff
@@ -115,6 +116,9 @@ func NewSplitTextCarrier() *SplitTextCarrier {
 // The Span is separated in this way for a variety of reasons; the most
 // important is to give OpenTracing users a portable way to opt out of Trace
 // Attribute propagation entirely if they deem it a stability risk.
+//
+// Both byte slices may be nil; what is provided will be cleared and the
```
bensigelman: > what is provided will be cleared

The carriers are used for both injection and extraction... what does the above mean in the injection case? (Naively it sounds like the contents will be cleared/ignored, though of course that's not what we mean)

standardtracer/propagation.go
```diff
@@ -127,7 +133,7 @@ func (p *splitBinaryPropagator) InjectSpan(
 	}
 
 	// Handle the trace and span ids, and sampled status.
-	contextBuf := new(bytes.Buffer)
+	contextBuf := bytes.NewBuffer(splitBinaryCarrier.TracerState[:0])
```
bensigelman: for my information: the parameter is just to take a guess at the size/capacity, right?

bensigelman:
LGTM aside from the nit about the comment!

standardtracer/propagation.go
tschottdorf: this creates a buffer which uses the (truncated) slice from the binary carrier. That is, if the carrier has a 1000 character slice, we'll use the capacity of 1000 (but set the length to zero).

propagation.go
tschottdorf: Updated.

PR #55, opened 2016-02-15
Add dynamic configuration

tschottdorf:
Introduce the Options type and corresponding Tracer
constructor NewWithOptions.
Options currently include a user-provided sample
function (i.e. a mapping from TraceID to boolean),
the SpanRecorder, and the option to ignore logs,
attributes and tags (except ext.SamplingPriority)
on non-sampled Spans for performance.

Support ext.SamplingPriority in standardtracer.

This needs a few unit tests, but I wanted to get into
the discussion first. The motivation for this change is
wanting to use `standardtracer` in [cockroachdb/cockroach](https://github.com/cockroachdb/cockroach);
these are some basic changes which make relevant
functionality available.

Added benchmark which demonstrates that a `trimmed` Span behaves
effectively like a no-op span, except that there will still be serialization overhead.
This can be avoided a) by better optimization of serialization and b)
by checking for the SamplingPriority tag before calls to the injector
(if serialization is to be avoided entirely).

```
BenchmarkTrimmedSpan_100Events_100Tags_100Attributes-8	   50000	     33144 ns/op	       0 B/op	       0 allocs/op
```

### DIFF
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -22,6 +22,9 @@ var (
 
 	// PeerPort records port number of the peer
 	PeerPort = uint16Tag("peer.port")
+
+	// SamplingPriority determines the priority of sampling this Span.
+	SamplingPriority = uint32Tag("sampling.priority")
 )
 
 type stringTag string
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -1,8 +1,6 @@
 package ext
 
-import (
-	"github.com/opentracing/opentracing-go"
-)
+import opentracing "github.com/opentracing/opentracing-go"
 
 var (
 	// PeerXXX tags can be emitted by either client-side of server-side to describe
@@ -24,7 +22,7 @@ var (
 	PeerPort = uint16Tag("peer.port")
 
 	// SamplingPriority determines the priority of sampling this Span.
-	SamplingPriority = uint32Tag("sampling.priority")
+	SamplingPriority = uint16Tag("sampling.priority")
 )
 
 type stringTag string
--- a/standardtracer/tracer.go
+++ b/standardtracer/tracer.go
@@ -2,16 +2,67 @@ package standardtracer
 
 import (
 	"sync"
+	"sync/atomic"
 	"time"
 
 	opentracing "github.com/opentracing/opentracing-go"
 )
 
-// New creates and returns a standard Tracer which defers to `recorder` after
-// RawSpans have been assembled.
-func New(recorder SpanRecorder) opentracing.Tracer {
+type sampleFunc func(int64) bool
+
+// Options allows creating a customized Tracer via NewWithOptions. The object
+// is thread safe and can be updated freely to influence the behavior of the
+// Tracer.
+type Options struct {
+	shouldSample       atomic.Value // contains a sampleFunc
+	trimUnsampledSpans int32        // updated atomically
+	recorder           atomic.Value // contains a SpanRecorder
+}
+
+// DefaultOptions returns an Options object with a 1 in 64 sampling rate and
+// all options disabled. A Recorder needs to be set manually before using the
+// returned object with a Tracer.
+func DefaultOptions() *Options {
+	opts := &Options{}
+	opts.ShouldSample(func(traceID int64) bool { return traceID%64 == 0 })
+	return opts
+}
+
+// ShouldSample takes a function which is called when creating a new Span and
+// determines whether that Span is sampled. The randomized TraceID is supplied
+// to allow deterministic sampling decisions to be made across different nodes.
+// For example,
+//
+//   func(traceID int64) { return traceID % 64 == 0 }
+//
+// samples every 64th trace on average.
+func (o *Options) ShouldSample(f func(int64) bool) {
+	o.shouldSample.Store(sampleFunc(f))
+}
+
+// Recorder receives Spans which have been finished.
+func (o *Options) Recorder(recorder SpanRecorder) {
+	o.recorder.Store(recorder)
+}
+
+// TrimUnsampledSpans turns potentially expensive operations on unsampled
+// Spans into no-ops. More precisely, tags, attributes and log events
+// are silently discarded.
+func (o *Options) TrimUnsampledSpans(b bool) {
+	var i int32
+	if b {
+		i = 1
+	}
+	atomic.StoreInt32(&o.trimUnsampledSpans, i)
+}
+
+// NewWithOptions creates a customized Tracer.
+func NewWithOptions(opts *Options) opentracing.Tracer {
+	if opts == nil {
+		panic("nil Options passed to NewWithOptions")
+	}
 	rval := &tracerImpl{
-		recorder: recorder,
+		Options: opts,
 		spanPool: sync.Pool{New: func() interface{} {
 			return &spanImpl{}
 		}},
@@ -22,10 +73,21 @@ func New(recorder SpanRecorder) opentracing.Tracer {
 	return rval
 }
 
+// New creates and returns a standard Tracer which defers completed Spans to
+// `recorder`.
+// Spans created by this Tracer support the ext.SamplingPriority tag: Calling
+// SetTag(ext.SamplingPriority, nil) causes the Span to be Sampled from that
+// point on.
+func New(recorder SpanRecorder) opentracing.Tracer {
+	opts := DefaultOptions()
+	opts.Recorder(recorder)
+	return NewWithOptions(opts)
+}
+
 // Implements the `Tracer` interface.
 type tracerImpl struct {
+	*Options
 	spanPool         sync.Pool
-	recorder         SpanRecorder
 	textPropagator   *splitTextPropagator
 	binaryPropagator *splitBinaryPropagator
 	goHTTPPropagator *goHTTPPropagator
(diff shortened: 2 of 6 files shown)


ext/tags.go
```diff
@@ -22,6 +22,9 @@ var (
 
 	// PeerPort records port number of the peer
 	PeerPort = uint16Tag("peer.port")
+
+	// SamplingPriority determines the priority of sampling this Span.
+	SamplingPriority = uint32Tag("sampling.priority")
```
yurishkuro: curious, why 32bit?

standardtracer/tracer.go
```diff
@@ -2,16 +2,67 @@ package standardtracer
...
+	"sync/atomic"
 	"time"
 
 	opentracing "github.com/opentracing/opentracing-go"
 )
 
-// New creates and returns a standard Tracer which defers to `recorder` after
-// RawSpans have been assembled.
-func New(recorder SpanRecorder) opentracing.Tracer {
+type sampleFunc func(int64) bool
```
yurishkuro: in our implementation Sampler is an interface, similar to the Reporter (or Recorder), because we want them to have a `Close()` method. One of our Sampler implementations is called "remotely controlled", because it establishes a connection to the mothership, which needs to be closed for clean shutdown.

standardtracer/tracer.go
```diff
@@ -2,16 +2,67 @@ package standardtracer
...
+
+// Options allows creating a customized Tracer via NewWithOptions. The object
+// is thread safe and can be updated freely to influence the behavior of the
+// Tracer.
+type Options struct {
+	shouldSample       atomic.Value // contains a sampleFunc
+	trimUnsampledSpans int32        // updated atomically
+	recorder           atomic.Value // contains a SpanRecorder
```
yurishkuro: why not have these immutable?

tschottdorf: I want to be able to change them at runtime (probably not `trimUnsampledSpans`, but nothing is lost - if you want immutable, simply don't hold on to the Options object after creating the tracer).

standardtracer/tracer.go
tschottdorf: This tracer doesn't have moving parts, so it doesn't need to be involved in shutting down the system.

ext/tags.go
tschottdorf: No reason, I just took what's here - I don't fully understand this code. What's the point of these internal tag types which are all strings? Using this tag from `standardtracer` is now an awkward `SetTag(string(ext.SamplingPriority), ...)`.

yurishkuro: The point here is that the tags are _typed_, the way you set them is via:
```go
ext.SamplingPriority.Add(span, value)
```
and the compiler enforces the correct type of `value`

NB: we should change the syntax to use `Set` method instead of `Add`, it's a left-over from `AddTag` API

standardtracer/tracer.go
```diff
@@ -22,10 +73,21 @@ func New(recorder SpanRecorder) opentracing.Tracer {
...
+// point on.
+func New(recorder SpanRecorder) opentracing.Tracer {
+	opts := DefaultOptions()
+	opts.Recorder(recorder)
+	return NewWithOptions(opts)
+}
+
 // Implements the `Tracer` interface.
 type tracerImpl struct {
+	*Options
```
yurishkuro: @tschottdorf  curious, why did you go for a pointer this time, instead of embedding the object?

ext/tags.go
tschottdorf: but if you look in the code below, they're immediately converted to string, so there's really no typing going on here at all. It won't work that way until `SetTag` takes an `interface{}` key.

standardtracer/tracer.go
tschottdorf: The `Options` struct contains various non-copyable fields, so it's either making all of these pointers or having a top-level pointer. I prefer the latter though there's no good reason for it.

tschottdorf: The other way around is probably better, come to think of it. I'll make the change (and expose Options only through an interface, so nobody can create an uninitialized one).

tschottdorf: Ok, enough flip-flopping: Introducing the interface makes the *Option the best choice after all.

ext/tags.go
yurishkuro: You mean values are converted to string by the Span?  First, it's the implementation detail of this particular tracer, some tracers may want to store values in the native format without conversion. But more importantly, the point here is to provide type safety, for example you won't be able to set a port tag with a string garbage 
```
PeerPort.Add(span, "garbage")
```
the compiler will require `uint16` argument to this `Add` method.

standardtracer/tracer.go
yurishkuro: what I meant was: by treating them immutable you can avoid synchronization via `atomic.Value`. What is the use case when you want to change these at runtime (vs. in tests where you can still do it)? And is that use case worth the performance overhead?

tschottdorf: Adjusting the sampling rate in a running system, and we have a lot of testing situations in which it is very inconvenient to keep the system from moving until the test wants to set its tracing options (often, that would be a benchmark not keen on tracing). I'm not too worried about the performance overhead. Most of it is off the critical path (`trimUnsampledSpans` probably gets the most traffic), and atomic updates are pretty fast.
What pattern were you suggesting regarding immutability - I didn't quite follow.

ext/tags.go
tschottdorf: I see. Would make more sense (in `Go`) to not expose the constants directly but via an interface:

```go
type Uint16Tag interface{
  Set(opentracing.Span, value uint16)
}
var PeerPort Uint64Tag = uint64Tag("peer.port")
```

standardtracer/tracer.go
bensigelman: @tschottdorf why can't the sampleFunc handle any realtime adjustments internally?

My concern here is not performance, but just plain-old complexity. We have taken what would be a straightforward Options struct and turned it into something with subtlety and opacity.

Also, FWIW, it's important to record the sampling rate somewhere if statistical statements are to be made about the recorded traces later. That's why "dapperish" uses more than a bool to record the sampling state (dapper recorded the sampling *decision* as a bit, but also packed the reciprocal sampling rate in there as well). Finally, a truly smart sampler would want more than just the trace_id to make the decision, but that seems out of scope.

However, stepping way back from the minutia of this change: it worries me that we're debating things this impl-specific in any `opentracing-*` repository. It seems like `standardtracer` is quickly becoming `cockroachtracer` :)  Which is not necessarily a bad thing, as it was just a toy before. But a lot of the details of the "standard" API are making assumptions about implementation priorities that may not generalize well.

bensigelman: To clarify my intent in the last paragraph there: I don't object to this PR in particular, but we should probably talk about the general goals/plan... today will be tricky, but tomorrow is fine (or we could use the wednesday standing google hangout meeting if it's ok to wait that long).

yurishkuro: @bensigelman @tschottdorf why don't we fork it now into its own repo, like `reference-tracer-go`?

It will also help with vendoring, I am working on adding tracing instrumentation to our internal libs, which requires me to vendor opentracing, and I don't really need a standardtracer in there.

tschottdorf: Don't worry, I appreciate the feedback and I'm really mostly testing the water here. Obviously I want a tracer that we can actually use at `cockroachdb`. We had some forking experiments when we implemented logging, and I value getting "grounded" in something that doesn't spiral off the common code base into an abyss of weirdness.

Stepping back, I think I can probably hide the dynamic options inside of the recorder and sampleFunc - I'll set `trimUnsampledSpans` unconditionally anyway. Getting rid of the atomic updates definitely doesn't make me feel bad.

I kept the `sampleFunc` simple because well - there simply isn't much information and I'm decidedly doing upfront sampling here since I don't want to pay the price for storing information that a smart sampler would later need. It's a start.

Regarding the forking, I thought that was going to happen anyway. As far as I'm concerned, now's as good as anytime. Maybe after the currently open PRs close.

bensigelman: SGTM re :fork_and_knife:!

bensigelman: (and thanks for being flexible, and thanks as well for all of the great contributions)

tschottdorf: Sure thing. I added a commit which makes the options static.

ext/tags.go
yurishkuro: that's how it sort of was in the initial version of this file, then it got changed to current form that was deemed more streamlines / Go-like. I do like the ability to hide the string value to avoid downcasting abuse.

tschottdorf: Ok. Maybe worth switching back since there's also the change of accidental abuse here when overlooking the package-level methods. Alas, different PR.

tschottdorf:
From what I can tell, all open discussions resolved (or at least attempted to). PTAL

ext/tags.go
yurishkuro: Note that the final const must expose both a Set method and a string key. The Set method is used by client code, but the key is used by the tracer to build special handling of these tags. Maybe we should just add a comment at the top of the file showing the intended usage pattern (yes, different PR).

yurishkuro: back to the original question, do we need 32bit precision for sampling priority field? I would start with uint16.

yurishkuro:
LGTM

ext/tags.go
tschottdorf: oh, didn't know that was the original question. I'm fine with `uint16`.

bensigelman: (I'd be fine with `byte` for priority, FWIW)

Issue #56, opened 2016-02-19
Add LICENSE to opentracing-go


bensigelman:
(seems like this didn't happen!)

yurishkuro:
Adrian was also suggesting CONTRIBUTING.md

PR #57, opened 2016-02-19
add the apparently-missing MIT license file

bensigelman:
Per https://github.com/opentracing/opentracing-go/issues/56

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016 The OpenTracing Authors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


yurishkuro:
:shipit:

bensigelman:
I'll give people 48h to comment on this choice of license... I know there are some strong feelings out there. I am a when-in-Rome person with licenses, and I think MIT is the most prevalent for Go.

@yurishkuro @adriancole @slimsag have all said stuff about these topics before

adriancole:
If go cannot be used without MIT (as core dependencies are licensed that
way), than another license (such as Apache) doesn't add value. So, sgtm

tschottdorf:
:shipit:

bensigelman:
48h :hourglass:

PR #58, opened 2016-02-20
Delete `standardtracer` and clean up

tschottdorf:


bensigelman:
:ship:!

ext/tags_test.go
```diff
@@ -1,20 +1,27 @@
 package ext_test
 
 import (
+	"reflect"
 	"testing"
 
+	"github.com/opentracing/basictracer-go"
+	"github.com/opentracing/basictracer-go/testutils"
```
bensigelman: this is a little awkward from a deps standpoint, but for now, "whatever."

tschottdorf: Should be ok, it's only a test dependency. If it becomes an issue, someone will have to put a trivial tracer impl that supports attributes here.

yurishkuro: why not just use noop tracer? I think the only point of these tests (at least in python) was to get test coverage to 100%. In Go they have additional benefit of validating signatures, but we don't need asserts to read back the values imo.

examples/dapperish/dapper.go
```diff
@@ -1,11 +1,11 @@
 package dapperish
 
 import (
+	"github.com/opentracing/basictracer-go"
```
yurishkuro: can we move daperish into basic tracer too?  these recursive imports look strange.

ext/tags_test.go
tschottdorf: SGTM

PR #62, opened 2016-02-23
Remove basictracer-go dependency

bg451:
There is currently a test dependency loop between `basictracer-go` and `opentracing-go`. This removes the `basictracer-go` dependency by moving the `dapperish` example to basictracer. This PR also adds a noop tracer that has span tags implemented so `ext/tags_test.go` can work without `basictracer-go`.

Per https://github.com/opentracing/opentracing.github.io/issues/67#issuecomment-187444001


bg451:
ping @bensigelman

ext/tags_test.go
```diff
@@ -29,13 +27,59 @@ func TestPeerTags(t *testing.T) {
...
+func (n noopTracer) Injector(format interface{}) opentracing.Injector {
+	return noopInjectorExtractor{}
+}
+
+func (n noopInjectorExtractor) InjectSpan(span opentracing.Span, carrier interface{}) error {
+	return nil
+}
+
+func (n noopInjectorExtractor) JoinTrace(operationName string, carrier interface{}) (opentracing.Span, error) {
+	return nil, nil
```
yurishkuro: I think this, strictly speaking, is not a valid implementation of the method, I would rather `panic("not implemented")`

yurishkuro:
lgtm

PR #64, opened 2016-02-27
Change common tags to use Set instead of Add method. Add Span Kind.

yurishkuro:
Brings Go in parity with Python, where span kind tags have been defined for a while.

### DIFF
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -2,6 +2,17 @@ package ext
 
 import opentracing "github.com/opentracing/opentracing-go"
 
+// These constants define common tag names recommended for better portability across
+// tracing systems and languages/platforms.
+//
+// The tag names are defined as typed strings, so that in addition to the usual use
+//
+//     span.setTag(TagName, value)
+//
+// they also support value type validation via this additional syntax:
+//
+//    TagName.Set(span, value)
+//
 var (
 	// PeerXXX tags can be emitted by either client-side of server-side to describe
 	// the other side/service in a peer-to-peer communications, like an RPC call.
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -23,25 +23,54 @@ var (
 
 	// SamplingPriority determines the priority of sampling this Span.
 	SamplingPriority = uint16Tag("sampling.priority")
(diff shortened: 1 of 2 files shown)


ext/tags.go
```diff
@@ -23,25 +23,54 @@ var (
...
+
+	// SpanKindRPCServer marks a span representing the server-side of an RPC
+	// or other remote call
+	SpanKindRPCServer = SpanKindEnum("s")
 )
 
+// ---
+
+// SpanKindEnum represents common span types
+type SpanKindEnum string
```
tschottdorf: The plain types here make this pretty confusing to use - why not something like

```go
type SpanKind interface {
  spanKindMarker()
  Set(span opentracing.Span, value string)
}

type spanKindImpl struct {
  name string
}
func (sk *spanKindImpl) spanKindMarker() {}
func (sk *...) Set(...) { ... }

var SpanKindRPCClient SpanKind = (&spanKindImpl{name:"s"})(nil)
```
func (

yurishkuro: I've added new values using the same pattern used by the other ones. That pattern has been agreed upon back in Dec https://github.com/opentracing/opentracing-go/commit/3e2f0d7c16a7aa88576ef114a638cbaf3471f8c9#diff-0c56d0f1cdc8e3b8ba3be800367de5ee

If you think it's worth changing, e.g. from performance considerations, I think it should be a different PR.

bensigelman:
@yurishkuro forgive me for being lazy and not just checking myself, but is this PR compliant with @dkuebric's pending https://github.com/opentracing/opentracing.github.io/pull/61 ?

yurishkuro:
@bensigelman yes, it is (by the nature of just following what's already done in Python). This is form Dan's pr:

```
* `span.kind` - string
    - One of `c` or `s`, indicating if this span represents a client or server
```

bensigelman:
Well, this LGTM once you and @tschottdorf are on the same page...

ext/tags.go
tschottdorf: SGTM

PR #65, opened 2016-02-27
Simplify propagation by removing an API layer

bensigelman:
Per https://github.com/opentracing/opentracing.github.io/issues/70 . The Tracer.Injector() calls were going to be 1:1 with Injector.InjectSpan() calls, and the Tracer.Extractor() calls were going to be 1:1 with Extractor.JoinTrace() calls; this change cuts out the middleman.

@tschottdorf @bg451 @yurishkuro what do we want to do about `basictracer-go`? I can do this change in a way that breaks nothing there (i.e., temporarily support both APIs in OT-go, remove all deps on the old one, remove old API, etc, etc) or I can stage a change in `basictracer-go` and merge the two in close succession. No strong feelings either way but wanted to check whether anyone is depending on `basictracer-go` in a way where that sort of short-term breakage is unacceptable. (If so, I will obviously do the cleaner/laborious thing)

### DIFF
--- a/ext/tags_test.go
+++ b/ext/tags_test.go
@@ -43,8 +43,6 @@ type noopSpan struct {
 	Tags opentracing.Tags
 }
 
-type noopInjectorExtractor struct{}
-
 func (n noopSpan) SetTag(key string, value interface{}) opentracing.Span {
 	n.Tags[key] = value
 	return n
@@ -68,18 +66,10 @@ func (n noopTracer) StartSpanWithOptions(opts opentracing.StartSpanOptions) open
 	return noopSpan{Tags: make(opentracing.Tags)}
 }
 
-func (n noopTracer) Extractor(format interface{}) opentracing.Extractor {
-	return noopInjectorExtractor{}
+func (n noopTracer) Inject(toInject opentracing.Span, format interface{}, carrier interface{}) error {
+	panic("not implemented")
 }
 
-func (n noopTracer) Injector(format interface{}) opentracing.Injector {
-	return noopInjectorExtractor{}
-}
-
-func (n noopInjectorExtractor) InjectSpan(span opentracing.Span, carrier interface{}) error {
-	return nil
-}
-
-func (n noopInjectorExtractor) JoinTrace(operationName string, carrier interface{}) (opentracing.Span, error) {
+func (n noopTracer) Join(operationName string, format interface{}, carrier interface{}) (opentracing.Span, error) {
 	panic("not implemented")
 }
--- a/ext/tags_test.go
+++ b/ext/tags_test.go
@@ -66,7 +66,7 @@ func (n noopTracer) StartSpanWithOptions(opts opentracing.StartSpanOptions) open
 	return noopSpan{Tags: make(opentracing.Tags)}
 }
 
-func (n noopTracer) Inject(toInject opentracing.Span, format interface{}, carrier interface{}) error {
+func (n noopTracer) Inject(sp opentracing.Span, format interface{}, carrier interface{}) error {
 	panic("not implemented")
 }
 
--- a/propagation.go
+++ b/propagation.go
@@ -9,12 +9,12 @@ import "errors"
 var (
 	// ErrUnsupportedFormat occurs when the `format` passed to Tracer.Inject() or
 	// Tracer.Join() is not recognized by the Tracer implementation.
-	ErrUnsupportedFormat = errors.New("opentracing: Unknown or unsupported inject/join format")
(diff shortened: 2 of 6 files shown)


propagation.go
```diff
@@ -7,66 +7,36 @@ import "errors"
 ///////////////////////////////////////////////////////////////////////////////
 
 var (
-	// ErrTraceNotFound occurs when the `carrier` passed to Extractor.JoinTrace()
-	// is valid and uncorrupted but has insufficient information to join or
-	// resume a trace.
+	// ErrUnsupportedFormat occurs when the `format` passed to Tracer.Inject() or
+	// Tracer.Join() is not recognized by the Tracer implementation.
+	ErrUnsupportedFormat = errors.New("opentracing: Unknown or unsupported inject/join format")
```
bensigelman: (this was previously represented by a `nil` return from `Tracer.Injector(...)`

tschottdorf:
I'm fine with merging in close succession.

bensigelman:
@tschottdorf cool... once I have an LGTM here as well as https://github.com/opentracing/basictracer-go/pull/6 I will merge this PR followed by that PR.

span.go
```diff
@@ -184,16 +182,7 @@ func StartChildSpan(parent Span, operationName string) Span {
 	})
 }
 
-// InjectSpan is a helper that injects a Span instance into a carrier.
-//
-// See Tracer.Injector() and Injector.InjectSpan().
+// InjectSpan is a trivial helper that injects a Span instance into a carrier.
 func InjectSpan(sp Span, format interface{}, carrier interface{}) error {
```
tschottdorf: Looks ripe for removal.

tracer.go
```diff
@@ -5,7 +5,7 @@ import "time"
 // Tracer is a simple, thin interface for Span creation.
 //
 // A straightforward implementation is available via the
-// `opentracing/basictracer` package's `standardtracer.New()'.
+// `opentracing/basictracer-go` package's `standardtracer.New()'.
```
tschottdorf: the package is actually called `basictracer`, though.

span.go
bensigelman: It is borderline... the `InjectSpan` helper does remove the slight stuttering of specifying the Span twice. But I will kill it -- easier to add later.

tracer.go
bensigelman: In the sense of `package basictracer`, yes... in the sense of the pathname that we have here (or the github repo), no. No strong feelings but wanted to explain the reasoning.

bensigelman:
(minor changes pushed)

yurishkuro:
LGTM

bg451:
lgtm

ext/tags_test.go
```diff
@@ -68,18 +66,10 @@ func (n noopTracer) StartSpanWithOptions(opts opentracing.StartSpanOptions) open
...
-}
-
-func (n noopTracer) Injector(format interface{}) opentracing.Injector {
-	return noopInjectorExtractor{}
-}
-
-func (n noopInjectorExtractor) InjectSpan(span opentracing.Span, carrier interface{}) error {
-	return nil
+func (n noopTracer) Inject(toInject opentracing.Span, format interface{}, carrier interface{}) error {
+	panic("not implemented")
```
tschottdorf: shouldn't that succeed always (and same for unmarshaling)?

tracer.go
tschottdorf: sgtm

tracer.go
```diff
@@ -27,11 +27,12 @@ type Tracer interface {
 	StartSpan(operationName string) Span
 	StartSpanWithOptions(opts StartSpanOptions) Span
 
-	// Return an Injector for the given `format` value, or nil if the Tracer
-	// does not support such a format.
+	// Inject() takes the `toInject` Span instance and represents it for
```
tschottdorf: I find `toInject` confusing because it sounds like something is to be injected into the span. Just `sp` is better imo.

tschottdorf:
LGTM

ext/tags_test.go
bensigelman: I think the idea here was that `tags_test` should never be calling these methods so the panic's help reinforce that understanding. In any case, I was just trying to make Inject symmetrical with https://github.com/opentracing/opentracing-go/pull/65/files#diff-9f1aa8a53149ad3fc609b2518bf25139L84

tracer.go
bensigelman: huh, interesting... well, I have no strong feelings so will change this to `sp`.

PR #69, opened 2016-03-04
Change Go-Context methods not to return the input span

yurishkuro:
Per gitter conversation, returning the input span is only useful in very specialized cases, like this:

```go
ctx, span := opentracing.ContextWithSpan(ctx, tracer.StartSpan(…))
```


bensigelman:
LGTM

:crying_cat_face:

PR #70, opened 2016-03-05
Fix a nil map assignment in JoinFromHeader

bg451:
`JoinFromHeader` now creates the appropriate map if the `SplitTextCarrier`'s map are nil, since `NewSplitTextCarrier` doesn't create any maps. This fixes a nil map assignment bug for when a `Tracer` doesn't implement `Join` for `GoHTTPHeader` and tries to use `Join` for `SplitText`.

Here's the panic in action.

```
2016/03/05 09:46:59 http: panic serving [::1]:51661: assignment to entry in nil map
goroutine 11 [running]:
net/http.(*conn).serve.func1(0xc8200c7b80, 0x86edd0, 0xc820032168)
	/usr/local/go/src/net/http/server.go:1287 +0xb5
github.com/opentracing/opentracing-go.JoinFromHeader(0x4ecba0, 0xa, 0xc82019a000, 0x86ec00, 0xc820118e10, 0x0, 0x0, 0x0, 0x0)
	/Users/bg/go/src/github.com/opentracing/opentracing-go/utils.go:76 +0x87d
main.server.func1(0xba0058, 0xc8201a0000, 0xc82011a700)
	/Users/bg/go/src/sourcegraph.com/sourcegraph/appdash/opentracing/example/main.go:65 +0x6d
net/http.HandlerFunc.ServeHTTP(0x5a5a00, 0xba0058, 0xc8201a0000, 0xc82011a700)
	/usr/local/go/src/net/http/server.go:1422 +0x3a
net/http.(*ServeMux).ServeHTTP(0xc820016960, 0xba0058, 0xc8201a0000, 0xc82011a700)
	/usr/local/go/src/net/http/server.go:1699 +0x17d
net/http.serverHandler.ServeHTTP(0xc82015a000, 0xba0058, 0xc8201a0000, 0xc82011a700)
	/usr/local/go/src/net/http/server.go:1862 +0x19e
net/http.(*conn).serve(0xc8200c7b80)
	/usr/local/go/src/net/http/server.go:1361 +0xbee
created by net/http.(*Server).Serve
	/usr/local/go/src/net/http/server.go:1910 +0x3f6
```

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -66,12 +66,18 @@ func JoinFromHeader(
 			if err != nil {
 				return nil, err
 			}
+			if carrier.TracerState == nil {
+				carrier.TracerState = map[string]string{}
+			}
 			carrier.TracerState[strings.TrimPrefix(key, ContextIDHTTPHeaderPrefix)] = unescaped
 		} else if strings.HasPrefix(key, TagsHTTPHeaderPrefix) {
 			// We don't know what to do with anything beyond slice item v[0]:
 			unescaped, err := url.QueryUnescape(val[0])
 			if err != nil {
 				return nil, err
+			}
+			if carrier.Baggage == nil {
+				carrier.Baggage = map[string]string{}
 			}
 			carrier.Baggage[strings.TrimPrefix(key, TagsHTTPHeaderPrefix)] = unescaped
 		}


bensigelman:
The fix looks good. The function in general should move out of `opentracing-go` proper (since it's not part of the core API), but I'm not sure where, exactly... `basictracer` is a possibility, as is a sub-package. But that's not the purpose of this PR so happy to deal with it separately.

https://github.com/opentracing/opentracing-go/issues/71

bensigelman:
Actually, the function itself is pretty broken... the prefix stuff should be handled by the SplitTextCarrier, not by this layer. (Which reminds me also of https://github.com/opentracing/opentracing.github.io/issues/73, but that's a larger topic!)

We can merge this if it's blocking you, but there's more to do for sure.

bg451:
It's a bug I found on accident, so it's not blocking anything. Just figured I should open a PR for it.

PR #72, opened 2016-03-11
Add StartSpanFromContext (and testTracer)

bensigelman:
I've found myself wanting this function over and over again in my own Go instrumentation code... thoughts?

@beberlei @yurishkuro @tschottdorf @bg451 have you done something else for (safely/concisely) creating a `Span` from a `context.Context`?

### DIFF
--- a/testtracer_test.go
+++ b/testtracer_test.go
@@ -1,0 +1,48 @@
+package opentracing
+
+// testTracer is a most-noop Tracer implementation that makes it possible for
+// unittests to verify whether certain methods were / were not called.
+type testTracer struct{}
+
+type testSpan struct {
+	OperationName string
+	HasParent     bool
+}
+
+// testSpan:
+func (n testSpan) SetTag(key string, value interface{}) Span             { return n }
+func (n testSpan) Finish()                                               {}
+func (n testSpan) FinishWithOptions(opts FinishOptions)                  {}
+func (n testSpan) SetBaggageItem(key, val string) Span                   { return n }
+func (n testSpan) BaggageItem(key string) string                         { return "" }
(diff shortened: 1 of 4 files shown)


testtracer_test.go
```diff
@@ -0,0 +1,48 @@
+package opentracing
+
+// testTracer is a most-noop Tracer implementation that makes it possible for
+// unittests to verify whether certain methods were / were not called.
+type testTracer struct{}
```
bensigelman: a noopTracer fork, basically... obviously could do a lot more here, all I really needed at the moment was `testSpan.HasParent`

yurishkuro:
+1

testtracer_test.go
tschottdorf: Why do you need all of this? Just use `basictracer` with an `InMemoryRecorder`, and finish the child span which gives you access to `RawSpan.Context.ParentID`.

tschottdorf:
We're not using this, but it looks useful.

testtracer_test.go
yurishkuro: I'd rather not have circular dependency, or in fact any extra dependencies in opentracing-go

tschottdorf: Oh, didn't even think about that. Good catch.

bensigelman: (@yurishkuro's reason was precisely why I didn't just reuse the far superior `basictracer-go` machinery!)

bensigelman:
Looks like people think this is worthwhile... FWIW, I can see an argument for moving this out of the main `opentracing` package (maybe a utils subdir or something), but I'm going to err on the side of simplicity and leave it here for the time being.

PR #73, opened 2016-03-16
Move to the simpler Binary and TextMap carriers

bensigelman:
Per opentracing/opentracing.github.io#73

### DIFF
--- a/propagation.go
+++ b/propagation.go
@@ -41,71 +41,42 @@ var (
 type BuiltinFormat byte
 
 const (
-	// SplitBinary encodes the Span in a SplitBinaryCarrier instance.
+	// Binary encodes the Span in a BinaryCarrier instance.
 	//
-	// The `carrier` for injection and extraction must be a
-	// `*SplitBinaryCarrier` instance.
-	SplitBinary BuiltinFormat = iota
+	// The `carrier` for injection and extraction must be a `BinaryCarrier`
+	// instance (not a pointer to a BinaryCarrier instance).
+	Binary BuiltinFormat = iota
 
-	// SplitText encodes the Span in a SplitTextCarrier instance.
+	// TextMap encodes the Span in a TextMapCarrier instance.
 	//
-	// The `carrier` for injection and extraction must be a `*SplitTextCarrier`
-	// instance.
-	SplitText
+	// The `carrier` for injection and extraction must be a `TextMapCarrier`
+	// instance (not a pointer to a TextMapCarrier instance).
+	TextMap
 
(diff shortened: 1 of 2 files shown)


propagation.go
```diff
@@ -41,71 +41,69 @@ var (
...
-	// instance.
+	// instance (not a pointer to an http.Header instance).
+	//
+	// If there are entries in the http.Header map prior to a call to Inject(),
+	// they are left alone (i.e., the map is not cleared). Similarly, in calls
+	// to Join() it is fine (and expected in some cases) for the http.Header
+	// map to contain other unrelated data (i.e., non-OpenTracing headers).
 	GoHTTPHeader
+
+	// SplitBinary is DEPRECATED
```
bensigelman: I will remove these as soon as https://github.com/opentracing/basictracer-go/pull/14 is merged.

propagation.go
```diff
@@ -41,71 +41,69 @@ var (
...
+type TextMapCarrier map[string]string
+
+// BinaryCarrier represents a Span for propagation as an opaque byte array.
 //
-// It is legal to provide one or both maps as `nil`; they will be created
-// as needed. If non-nil maps are provided, they will be used without
-// clearing them out on injection.
+// It is fine to pass `nil` to Inject(); in that case, it will allocate a
+// []byte for the caller.
+type BinaryCarrier *[]byte
```
bensigelman: Curious what people think about this... it seems a little silly to wrap such basic types in a struct but I could be convinced. I also considered a `bytes.Buffer` here, though this seemed simpler and more general.

yurishkuro:
lgtm

propagation.go
bg451: I'm fine with this as is. The two main arguments I would have in favor of using `bytes.Buffer` is 1) they're reusable, so user can use a pool of buffers if they want, and 2) `byte.Buffers` are fairly common in Go, so it would remove a couple of steps of turning the `BinaryCarrier` into a buffer post injection in a few cases. Those two arguments are probably premature optimizations though.

yurishkuro: or we could go with interface approach and use io.Writer and io.Reader for injector/extractor.

bensigelman: I'm going to create a separate PR illustrating the interface approach... that makes certain optimizations feasible without specifying a concrete type like `bytes.Buffer`.

bensigelman: https://github.com/opentracing/opentracing-go/pull/74

bensigelman:
(went with interfaces instead)

PR #74, opened 2016-03-17
Represent the simplified carriers as interfaces

bensigelman:
See https://github.com/opentracing/opentracing-go/pull/73 for comparison (and https://github.com/opentracing/opentracing.github.io/issues/73 for context).


propagation.go
```diff
@@ -41,71 +44,83 @@ var (
...
-// The Span is separated in this way for a variety of reasons; the most
-// important is to give OpenTracing users a portable way to opt out of
-// Baggage propagation entirely if they deem it a stability risk.
+// TextMapCarrier represents a Span for propagation within a key:value map of
+// unicode strings.
+type TextMapCarrier interface {
+	// Add a key:value pair to the carrier.`key` should be prefixed in a way
+	// that protects against collisions with things like HTTP headers (which
+	// may share space with Tracer data in the TextMapCarrier).
+	Add(key, val string)
```
bensigelman: the comment for this feels kind of awkward, but so it goes!

propagation.go
```diff
@@ -41,71 +44,83 @@ var (
...
+
+	// GetAll returns all contents of the text map via repeated calls to the
+	// `handler` function. If any call to `handler` returns a non-nil error,
+	// GetAll terminates and returns that error.
+	//
+	// NOTE: `handler` may be invoked for key:value combinations that were
```
bensigelman: (decided to embrace the multimap-ness)

utils.go
```diff
@@ -57,30 +40,29 @@ func JoinFromHeader(
...
-			carrier.TracerState[strings.TrimPrefix(key, ContextIDHTTPHeaderPrefix)] = unescaped
-		} else if strings.HasPrefix(key, TagsHTTPHeaderPrefix) {
-			// We don't know what to do with anything beyond slice item v[0]:
-			unescaped, err := url.QueryUnescape(val[0])
+	// Else, fall back on TextMap.
+	carrier := HTTPHeaderTextMapCarrier{h}
+	return tracer.Join(operationName, TextMap, carrier)
+}
+
+type HTTPHeaderTextMapCarrier struct {
```
bensigelman: not sure where this belongs.

propagation.go
yurishkuro: it does. Whose responsibility is it to prefix, and why?  E.g. I see Tracer calling `Add` with keys like `ARRRPC_TRACE_ID` and `ARRRPC_BAGGAGE_{actual_key}`, but the implementor of the carrier does not need to do anything.

propagation.go
```diff
@@ -41,71 +44,83 @@ var (
...
+	// *not* added via `Set` (in this process or otherwise). As such,
+	// implementations MUST check that `key` is something they care about.
+	//
+	// NOTE: A single `key` may appear in multiple calls to `handler` for a
+	// single `GetAll` invocation.
+	GetAll(handler func(key, val string) error) error
```
yurishkuro: how about splitting read/write interfaces? They are never needed together in the same carrier instance, and in some cases may be impossible to implement without resorting to duds.

propagation.go
```diff
@@ -41,71 +44,83 @@ var (
...
+}
+
+// BinaryCarrier represents a Span for propagation as an opaque byte stream.
 //
-// It is legal to provide one or both maps as `nil`; they will be created
-// as needed. If non-nil maps are provided, they will be used without
-// clearing them out on injection.
+// The io.Writer is intended for use with Tracer.Inject() and the io.Reader is
+// intended for use with Tracer.Join().
+type BinaryCarrier interface {
```
yurishkuro: I suggest (a) split read and write, (b) avoid defining this altogether, because we can use `io.Reader`/Writer directly.

utils.go
yurishkuro: seems ok here, or utils. In theory we could also provide a wrapper for `map[string]string` to satisfy text map carrier.

utils.go
```diff
@@ -57,30 +40,29 @@ func JoinFromHeader(
...
+	http.Header
```
yurishkuro: I suggest instead (assuming we're not splitting read/write, which I think we should):
```go
type HTTPHeaderTextMapCarrier http.Header
```

propagation.go
yurishkuro: +1

propagation.go
yurishkuro: i.e. `TextMapReader` and `TextMapWriter` (not sure we need the word "carrier" in the type)

propagation.go
bg451: +1

propagation.go
```diff
@@ -41,71 +44,83 @@ var (
...
 	//
 	// The `carrier` for both injection and extraction must be an http.Header
-	// instance.
+	// instance (not a pointer to an http.Header instance).
+	//
+	// If there are entries in the http.Header map prior to a call to Inject(),
+	// they are left alone (i.e., the map is not cleared). Similarly, in calls
+	// to Join() it is fine (and expected in some cases) for the http.Header
+	// map to contain other unrelated data (i.e., non-OpenTracing headers).
 	GoHTTPHeader
```
bmizerany: Are you planning on adding `JavaHTTPHeader`, `JavascriptHTTPHeader`, `RubyHTTPHeader`, etc? If not, maybe consider removing the `Go` prefix.

bensigelman: @bmizerany I considered this the other day... the concern I had was a non-`http.Header` representation of http headers, but – upon reflection – the package scope on such a format should be enough to distinguish it. I'll change this.

propagation.go
bensigelman: @yurishkuro fair point... I guess there's no need to tell the interface implementor about this sort of thing. (Another benefit to using an `interface` rather than the raw underlying storage type)

propagation.go
bensigelman: I see what you're getting at for sure. This is a deviation from the "concept" of a carrier, though, which was originally conceived as a data structure rather than an interface. If nothing else, it almost feels like the wrong word, now.

Per some of the Gitter traffic yesterday, I wonder if we should have `format`s that correspond to particular <injector, extractor> interface pairs. Thoughts?

I will try that out in this PR later in the afternoon.

yurishkuro: I definitely prefer interfaces to data structures, can always convert the first to the second, but not otherwise.

We could use the term "protocol", the way it's used in Thrift: 

> The Protocol abstraction defines a mechanism to map in-memory data structures to a wire-format. In other words, a protocol specifies how datatypes use the underlying Transport to encode/decode themselves. Thus the protocol implementation governs the encoding scheme and is responsible for (de)serialization. Some examples of protocols in this sense include JSON, XML, plain text, compact binary etc.

bensigelman:
@yurishkuro @bg451 @bmizerany @tschottdorf 

PTAL... Probably worth thinking this one over a bit. IMO it's an improvement and addresses a lot of problems at the expense of extra indirection. I'll add a few comments here and there.

propagation.go
```diff
@@ -41,71 +46,124 @@ var (
...
+	//
+	// For Tracer.Inject(): the carrier must be a `TextMapWriter`.
+	//
+	// For Tracer.Join(): the carrier must be a `TextMapReader`.
+	//
+	// See HTTPHeaderTextMapCarrier for an implementation of both TextMapWriter
+	// and TextMapReader that defers to an http.Header instance for storage.
+	// For example, Inject():
+	//
+	//    carrier := HTTPHeaderTextMapCarrier{
```
bensigelman: I got rid of the `util.go` stuff because (a) it really didn't belong in the base library, and (b) after this change it was nothing more than these five lines.

propagation.go
```diff
@@ -41,71 +46,124 @@ var (
...
+// the caller can decode a propagated Span as entries in a multimap of unicode
+// strings.
+type TextMapReader interface {
+	// GetAll returns TextMap contents via repeated calls to the `handler`
+	// function. If any call to `handler` returns a non-nil error, GetAll
+	// terminates and returns that error.
 	//
-	// The `carrier` for both injection and extraction must be an http.Header
-	// instance.
-	GoHTTPHeader
```
bensigelman: this is no longer needed (though see my caveat about character casing)

propagation.go
```diff
@@ -41,71 +46,124 @@ var (
...
-//
-// The Span is separated in this way for a variety of reasons; the most
-// important is to give OpenTracing users a portable way to opt out of
-// Baggage propagation entirely if they deem it a stability risk.
+// HTTPHeaderTextMapCarrier satisfies both TextMapWriter and TextMapReader.
 //
-// It is legal to provide one or both maps as `nil`; they will be created
-// as needed. If non-nil maps are provided, they will be used without
-// clearing them out on injection.
+// NOTE: All `key` parameters to Add() and the ReadAllEntries() handler func
```
bensigelman: ... I'm having flashbacks to the debate about baggage keys. Not sure what alternative we have here.

propagation.go
```diff
@@ -41,71 +46,124 @@ var (
...
+	//    	Header:       h,
+	//    }
+	//    err := span.Tracer().Inject(span, TextMap, carrier)
+	//
+	// Or Join():
 	//
-	// The `carrier` for injection and extraction must be a `*SplitTextCarrier`
-	// instance.
+	//    carrier := HTTPHeaderTextMapCarrier{
+	//    	HeaderPrefix: "opentracing-",
```
yurishkuro: this example looks wrong to me, it implies that it's the instrumented library that decides which prefix to add to keys on the wire, while I think it's the tracer who should decide, as otherwise the library of the RPC peer service must use the exact same prefix. I think we only need 
```go
type HTTPHeaderTextMapCarrier http.Header```
```
and add read/write methods that delegate to the Header struct.

propagation.go
```diff
@@ -41,71 +46,124 @@ var (
...
+type TextMapWriter interface {
+	// Add a key:value pair to the carrier.
+	Add(key, val string)
+}
+
+// TextMapWriter is the Join() carrier for the TextMap builtin format. With it,
+// the caller can decode a propagated Span as entries in a multimap of unicode
+// strings.
+type TextMapReader interface {
+	// GetAll returns TextMap contents via repeated calls to the `handler`
```
yurishkuro: mismatch with the method name GetAll vs. ReadAllEntries

propagation.go
```diff
@@ -41,71 +46,124 @@ var (
...
+	// function. If any call to `handler` returns a non-nil error, GetAll
+	// terminates and returns that error.
 	//
-	// The `carrier` for both injection and extraction must be an http.Header
-	// instance.
-	GoHTTPHeader
-)
+	// NOTE: A single `key` may appear in multiple calls to `handler` for a
+	// single `GetAll` invocation.
+	ReadAllEntries(handler func(key, val string) error) error
```
yurishkuro: strictly speaking, this is a visitor pattern, maybe use that terminology? But no strong feeling.

bensigelman: I'll see what others think (if anything)... I also don't have any strong feelings.

propagation.go
bensigelman: @yurishkuro fair point -- this code got moved around a few times but it belongs in the Tracers. I will do that tomorrow, need to :zzz: now.

yurishkuro:
I like the new interface approach, we just need to decide on the responsibilities for prefixing and lower-casing. 

I think prefixing is the tracer's job, because it's implementation detail of how tracer decides to encode baggage into "headers" - it can choose to collapse them all into a base64 string and store in a single header for example, making prefixing unnecessary.

Lower-casing does not seem necessary to me either, as long as we maintain the restriction that baggage keys are case insensitive. I think it's the job of the concrete carrier to encode keys and values to respect restrictions of its transport, like Camel-Case HTTP header name, and url-encode HTTP header value while a different protocol, like TChannel or maybe some thrift-based RPC, which does not impose any restrictions on the alphabet of keys and values, does not need to perform any encoding. I think case insensitivity is the only nod we're giving to HTTP as being the most widely used transport, all other alphabet restriction do not need to be affecting Tracer and baggage.

propagation.go
bensigelman: (this is addressed in the latest revision)

propagation.go
```diff
@@ -41,71 +45,112 @@ var (
...
+// HTTPHeaderTextMapCarrier satisfies both TextMapWriter and TextMapReader.
 //
-// The Span is separated in this way for a variety of reasons; the most
-// important is to give OpenTracing users a portable way to opt out of
-// Baggage propagation entirely if they deem it a stability risk.
-//
-// It is legal to provide one or both maps as `nil`; they will be created
-// as needed. If non-nil maps are provided, they will be used without
-// clearing them out on injection.
+// NOTE: All `key` parameters to Add() and the ReadAllEntries() handler func
```
bensigelman: I can't decide whether it's fair to say that `HTTPHeaderTextMapCarrier` actually satisfies `TextMap(Writer|Reader)` given the string casing issues here. Oh well.

testtracer_test.go
```diff
@@ -39,10 +46,39 @@ func (n testTracer) StartSpanWithOptions(opts StartSpanOptions) Span {
...
 
 // Join belongs to the Tracer interface.
 func (n testTracer) Join(operationName string, format interface{}, carrier interface{}) (Span, error) {
+	switch format {
+	case TextMap:
+		// Just for testing purposes... generally not a worthwhile thing to
+		// propagate.
+		rval := testSpan{}
+		err := carrier.(TextMapReader).ReadAll(func(key, val string) error {
+			switch strings.ToLower(key) {
```
bensigelman: This is the sort of thing that's going to be a gotcha for anyone who naively uses `TextMapCarrier` with an http header implementation... Oh well.

bensigelman:
Any thoughts about this PR? Having slept on it a bit I am positive about the direction (of moving towards carrier interfaces).

propagation.go
```diff
@@ -41,71 +45,115 @@ var (
...
+	// SplitText is DEPRECATED
 	SplitText
+)
+
+// TextMapWriter is the Inject() carrier for the TextMap builtin format. With
+// it, the caller can encode a Span for propagation as entries in a multimap of
+// unicode strings.
+type TextMapWriter interface {
+	// Add a key:value pair to the carrier. Multiple values may be added for a
+	// single (repeated) key.
```
yurishkuro: I was in favor of multimap semantics in span tags, but I am not in favor of it here. Firstly, I don't know why we need is, since the baggage is not a multimap, and trace ID does not require it. Secondly, making multimap support a requirement puts quite a burden on RPC framework writers, especially if their protocol does not allow it. Even in http headers the multimap support is implemented with an ugly semicolon-separated encoding.

propagation.go
```diff
@@ -41,71 +45,115 @@ var (
...
+	// ReadAll returns TextMap contents via repeated calls to the `handler`
+	// function. If any call to `handler` returns a non-nil error, ReadAll
+	// terminates and returns that error.
 	//
-	// The `carrier` for both injection and extraction must be an http.Header
-	// instance.
-	GoHTTPHeader
-)
+	// NOTE: A single `key` may appear in multiple calls to `handler` for a
+	// single `ReadAllEntries` invocation.
```
yurishkuro: name mismatch: ReadAll.

Here I do see the need for multimap semantics. We may want to clarify that the reason we introduce it is for cases where RPC framework implements application headers as a multimap.

propagation.go
bensigelman: The thinking here basically comes back to the HTTP header spec: https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

In particular this tidbit:

> Multiple message-header fields with the same field-name MAY be present in a message ...

(though the `...` above has a bunch of caveats about comma-separated values, alternate encodings, etc, etc... so maybe best to stick with `TextMapWriter.Set()` and call it a day)

propagation.go
bensigelman: It is pretty confusing for the reader and writer interfaces to be asymmetrical... I think I'd argue that we start with the plain (non-multi) map semantics which are, at least, clear; if there are other non-HTTP text maps to consider in the future it may be burdensome to force implementations to handle concatenation and escaping.

propagation.go
yurishkuro: I prefer if we put case-insensitive comment on the TextMapReader, since it's a requirement there by extension of lowest common denominator (http headers).

bensigelman: What about the symmetry argument, though? Saying that `TextMapReader` is case insensitive makes perfect sense to me, but what's the point of a case-sensitive `TextMapWriter` if `"key"` and `"kEy"` are collapsed into the same callback invocation in the reader?

propagation.go
```diff
@@ -41,71 +45,115 @@ var (
...
+	ReadAll(handler func(key, val string) error) error
```
yurishkuro: nit: maybe `ForEach` will be more conventional name for this pattern

propagation.go
```diff
@@ -41,71 +45,115 @@ var (
...
+// respect character casing for keys. Also note that the underlying http.Header
+// implementation will change the casing on keys per
+// http.CanonicalMIMEHeaderKey().
+type HTTPHeaderTextMapCarrier http.Header
+
+// Add conforms to the TextMapWriter interface.
+func (c HTTPHeaderTextMapCarrier) Add(key, val string) {
+	// We need to convert to a proper http.Header or `c.Add(...)` recurses
+	// infinitely.
+	h := (*http.Header)(&c)
```
yurishkuro: I think `(http.Header)(c)` works the same, since ultimately both types are just a map reference.

https://play.golang.org/p/uzTAE5cl2m

testtracer_test.go
```diff
@@ -39,10 +46,39 @@ func (n testTracer) StartSpanWithOptions(opts StartSpanOptions) Span {
...
 // Inject belongs to the Tracer interface.
 func (n testTracer) Inject(sp Span, format interface{}, carrier interface{}) error {
-	return nil
+	span := sp.(testSpan)
+	switch format {
+	case TextMap:
+		// Just for testing purposes... generally not a worthwhile thing to
+		// propagate.
+		carrier.(TextMapWriter).Add(testHTTPHeaderPrefix+"opname", span.OperationName)
+		carrier.(TextMapWriter).Add(testHTTPHeaderPrefix+"hasparent", strconv.FormatBool(span.HasParent))
```
yurishkuro: it feels like we're setting a bad example by adding opname and hasparent as baggage.

tracer.go
```diff
@@ -40,9 +40,13 @@ type Tracer interface {
 	//
 	// Example usage (sans error handling):
 	//
+	//     carrier := opentracing.HTTPHeaderTextMapCarrier{
+	//         HeaderPrefix: "opentracing-",
+	//         Header: httpReq.Header,
+	//     }
```
yurishkuro: wrong usage example

tracer.go
```diff
@@ -40,9 +40,13 @@ type Tracer interface {
...
 	//     tracer.Inject(
 	//         span,
-	//         opentracing.GoHTTPHeader,
+	//         opentracing.TextMap,
 	//         httpReq.Header)
```
yurishkuro: should be passing carrier, not header

tracer.go
```diff
@@ -76,10 +80,14 @@ type Tracer interface {
 	//
 	// Example usage (sans error handling):
 	//
+	//     carrier := opentracing.HTTPHeaderTextMapCarrier{
+	//         HeaderPrefix: "opentracing-",
+	//         Header: httpReq.Header,
```
yurishkuro: wrong usage

propagation_test.go
```diff
@@ -0,0 +1,57 @@
+package opentracing
+
+import (
+	"net/http"
+	"testing"
+)
+
+const testHeaderPrefix = "testprefix-"
+
+func TestHTTPHeaderInject(t *testing.T) {
```
yurishkuro: this test doesn't feel right, its logic relies mostly on the details of the test tracer implementation.  It looks more like a testable example (https://blog.golang.org/examples). For the actual code in the API, I think we should simply test using the HTTP carrier to put stuff and read stuff back, without involving the tracer.

propagation.go
yurishkuro: I think we also need to add a disclaimer here and in reader that the carrier does not guarantee that when reading values back the tracer will only be given keys it has written out.

yurishkuro: +1 for `Set`

propagation.go
bmizerany: Yes. Although `Foreach` or `ForeachKey` may be more conventional. See: https://godoc.org/golang.org/x/net/http2#SettingsFrame.ForeachSetting.

Is this using the callback for garbage reasons? I.e. Is it reusing memory underneath or trying to avoid allocating a large array/map? Maybe reading from the underlying value requires a lock? If so, some documentation explaining so would be helpful to understand why it isn't giving the user a convenient slice of strings or a map instead. It may also give me useful information about how/when to use it or how to implement it.

propagation.go
bmizerany: Just `http.Header(c)` should work.  https://play.golang.org/p/9mOR0iS5ex

propagation.go
yurishkuro: My concern is to simplify the implementation of the carrier by RPC owners. If RPC supports application headers as a plain map, then multi-map semantics of the Writer makes it difficult to implement. And conversely, if RPC supports multi-map headers, then requiring plain map in the Reader makes it difficult to implement. So rather than going with additional prescriptions like "concatenate values with semicolon" or "call handler only on the first value", I'd prefer an API that imposes minimal restrictions in each case, even if it makes it asymmetrical.

What do you think?

propagation.go
yurishkuro: Oh, fair enough, I think we should say both are case-insensitive. We should probably also use the same disclaimer that multiple calls to Set() with equivalent keys lead to undefined behavior - some carrier might append to underlying multi-map, other may override in the underlying plain map.

propagation.go
yurishkuro: @bmizerany I think there are a couple of reasons. One is that we don't know if the underlying RPC request represents application headers as a plain map or a multi-map; using a visitor avoids addressing that issue. Another is that a visitor API is more streaming-friendly, which incidentally addresses memory allocation concerns.

yurishkuro: @bensigelman I think @bmizerany has a good point about documenting the reasons for the API choice, to avoid revisiting them later.

propagation.go
bensigelman: @yurishkuro good point re the disclaimer... I'll add something. (This got lost during the revision where the carrier was handling the prefixing)

propagation.go
bensigelman: I think simple key:value associations are sufficiently expressive for Tracer implementors, and I am worried about novella-length comments scaring off OT "users" doing RPC over HTTP.

One other idea would be to call a spade a spade and basically `s/TextMap/HTTPHeaders`/ across languages. With all of the bizarre restrictions on keys, etc, we are paying the cost of that loss of generality already but not taking advantage of the existing mental model people have.

propagation.go
bensigelman: (this was me having flashbacks to C/C++... moot point in any case since we are doing `s/Add/Set/`)

testtracer_test.go
bensigelman: I guess so... it's "just a test" and all of that. I can add some other junk to the spans for this purpose, though.

propagation.go
bensigelman: On this note, earlier on @yurishkuro wrote:

> I think case insensitivity is the only nod we're giving to HTTP as being the most widely used transport, all other alphabet restriction do not need to be affecting Tracer and baggage.

There is also the translation of `_` and `-`, the poorly-defined treatment of extended ASCII chars (not to mention unicode proper), etc, etc, etc. I.e., HTTP header keys are a mess of uncertainty unless you stick with simple alphanumeric chars and hyphens.

So do we want the HTTP header implementation to do some funky escaping to make all of this work?

propagation.go
bensigelman: @bmizerany I like `ForeachKey`, thank you for the suggestion.

As for the reasoning: this PR is an attempt to move away from concrete structured types (maps, slices, etc) and towards interfaces that allow for any number of possible implementations. We could of course have the implementor build a `map[string]string` up from scratch (or whatever else), but being able to use their intrinsic data structure in place is more general. Thoughts? I will add a comment in any case.

propagation_test.go
bensigelman: Meh... I see what you're saying but I'd like to leave this as-is since – if nothing else – it makes it easy to quickly see the implications for various API changes. If you feel strongly I can make this change, though.

propagation.go
yurishkuro: I am not sure how much help adding `HTTPHeaders` format would provide. The key restrictions imposed by http actually spill out into baggage API, not so much the carriers. And if we did special case HTTP headers, we'd still need a text map for other RPCs (grpc, finagle, etc.)

Strictly speaking, we can even remove the case insensitive restriction and allow any strings for baggage keys, we'd just need tracers to encode the keys before injecting into http. We can provide a guide to implementers how to do that, e.g. assume that Canonical-Mixed-Case-Keys go in unchanged, and all other case combinations are url-encoded. The worst case is people won't be able to read the headers in tcpdump if they don't use Mixed-Case (and that's assuming tracer even keeps baggage in individual headers instead of a single base64 header).

bensigelman: Yeah, I hear you. I just (like 20 seconds ago) pushed a revision on this PR that attempts to find a middleground. I suspect that we'll need to revisit all of the text-map key stuff (esp for the baggage API) again before too long, but I can live with what we have at this layer for now.

tracer.go
```diff
@@ -40,10 +40,11 @@ type Tracer interface {
 	//
 	// Example usage (sans error handling):
 	//
+	//     carrier := opentracing.HTTPHeaderTextMapCarrier(Header: httpReq.Header)
```
yurishkuro: don't need `Header: `

yurishkuro:
LGTM (one correction to a comment).

One of the build was failing, I restarted.

yurishkuro:
Odd, the build fails with `Daily bandwidth rate limit exceeded for IP 52.22.60.255.`

```
# cd .; git clone https://go.googlesource.com/tools /home/travis/gopath/src/golang.org/x/tools
Cloning into '/home/travis/gopath/src/golang.org/x/tools'...
remote: Daily bandwidth rate limit exceeded for IP 52.22.60.255.
fatal: protocol error: bad pack header
package github.com/golang/lint
	imports golang.org/x/tools/go/gcimporter15: exit status 128
The command "go get -u github.com/golang/lint/..." failed and exited with 1 during .
Your build has been stopped.
```

bg451:
@yurishkuro Seems that `go.googlesource.com` is rate limited. https://github.com/golang/go/issues/12933

PR #75, opened 2016-03-17
keep func signatures on a single line

bmizerany:
Easier for `grep`, visual scanning, etc.


bensigelman:
@bmizerany I hear you on grep, don't-hear-you on visual scanning, but... I appreciate that you care and will merge for that reason alone :)

PR #77, opened 2016-04-12
Fix example in README

kyleconroy:
Span is the first return argument, not context


yurishkuro:
@kyleconroy - Thanks, good spot.

bensigelman:
:+1:, thanks, @kyleconroy!

PR #78, opened 2016-05-05
Add a trivial MockTracer to test OT client code

bensigelman:
See https://github.com/go-kit/kit/pull/247 where I need to use this.


yurishkuro:
lgtm. For completeness, it would be good to propagate baggage too

bensigelman:
Good idea, I'll do that before merging.

PR #79, opened 2016-05-06
Add HTTP tags to ext/

bensigelman:
... also for the benefit of https://github.com/go-kit/kit/pull/247

### DIFF
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -14,8 +14,33 @@ import opentracing "github.com/opentracing/opentracing-go"
 //    TagName.Set(span, value)
 //
 var (
-	// PeerXXX tags can be emitted by either client-side of server-side to describe
-	// the other side/service in a peer-to-peer communications, like an RPC call.
+	//////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


ext/tags.go
```diff
@@ -14,8 +14,33 @@ import opentracing "github.com/opentracing/opentracing-go"
...
+	//////////////////////////////////////////////////////////////////////
+	// SpanKind (client/server)
+	//////////////////////////////////////////////////////////////////////
+
+	// SpanKind hints at relationship between spans, e.g. client/server
+	SpanKind = spanKindTag("span.kind")
+
+	// SpanKindRPCClient marks a span representing the client-side of an RPC
+	// or other remote call
+	SpanKindRPCClient = SpanKindEnum("c")
```
yurishkuro: do you want to change them to client/server per Data Semantics, while at it?

yurishkuro:
lgtm

ext/tags.go
bensigelman: yes, good catch!

bg451:
lgtm

PR #80, opened 2016-05-21
Add standard tag "component", per http://opentracing.io/data-semantics/

yurishkuro:
Also change tests to validate against actual strings


bg451:
lgtm

PR #82, opened 2016-06-11
introduce and take advantage of SpanContext

bensigelman:


gocontext.go
```diff
@@ -7,23 +7,17 @@ type contextKey struct{}
 var activeSpanKey = contextKey{}
 
 // ContextWithSpan returns a new `context.Context` that holds a reference to
-// the given `Span`.
+// the given `Span`'s `SpanContext`.
 func ContextWithSpan(ctx context.Context, span Span) context.Context {
```
bensigelman: Since the proposed semantics allow for `Span.SpanContext()` to be called after `Span.Finish()`, I think it's fine to store the `Span` in the `context.Context

gocontext.go
```diff
@@ -7,23 +7,17 @@ type contextKey struct{}
...
 	return context.WithValue(ctx, activeSpanKey, span)
 }
 
-// BackgroundContextWithSpan is a convenience wrapper around
-// `ContextWithSpan(context.BackgroundContext(), ...)`.
-func BackgroundContextWithSpan(span Span) context.Context {
```
bensigelman: I got rid of this just because I regret it :)

Not related to the change here.

span.go
```diff
@@ -6,10 +6,52 @@ import (
...
+	//
+	// As such, `restrictedKey` MUST match the regular expression
+	// `(?i:[a-z0-9][-a-z0-9]*)` and is case-insensitive. That is, it must
+	// start with a letter or number, and the remaining characters must be
+	// letters, numbers, or hyphens. See CanonicalizeBaggageKey(). If
+	// `restrictedKey` does not meet these criteria, SetBaggageItem()
+	// results in undefined behavior.
+	//
+	// Returns a reference to this Span for chaining, etc.
+	SetBaggageItem(restrictedKey, value string) SpanContext
```
bensigelman: I really like how the baggage stuff moves to `SpanContext`... much clearer, conceptually speaking.

I didn't change the comments at all above... obv it looks like we'll move to remove these restrictions, but that's orthogonal.

span.go
```diff
@@ -28,7 +70,8 @@ type Span interface {
 	// Sets the end timestamp and calls the `Recorder`s RecordSpan()
 	// internally.
 	//
-	// Finish() should be the last call made to any span instance, and to do
+	// With the exception of calls to SpanContext() (which are always allowed),
```
bensigelman: this "exception" is the most important tangible benefit of this change.

tracer.go
```diff
@@ -57,7 +57,7 @@ type Tracer interface {
 	// fails anyway.
 	//
 	// See Tracer.Join().
-	Inject(sp Span, format interface{}, carrier interface{}) error
+	Inject(sc SpanContext, format interface{}, carrier interface{}) error
```
bensigelman: very important to note that Inject and Join are not symmetrical, type-wise... I think the API and semantics are much easier to reason about if `Join` returns a `Span` and `Inject` takes a `SpanContext`.

tracer.go
```diff
@@ -107,11 +107,11 @@ type StartSpanOptions struct {
...
 	OperationName string
 
-	// Parent may specify Span instance that caused the new (child) Span to be
-	// created.
+	// Parent may specify the SpanContext that caused the new (child) Span to
+	// be created.
 	//
 	// If nil, start a "root" span (i.e., start a new trace).
-	Parent Span
+	Parent SpanContext
```
bensigelman: (also an important change)

yurishkuro: nit: if we're going to change the Options struct, maybe we should go with the [functional options](https://commandcenter.blogspot.com.au/2014/01/self-referential-functions-and-design.html) instead. Doesn't need to be in the same PR, but I would vote for such change before bumping the API version. Only applies to Go, of course.

yurishkuro:
Overall, I like the change. However, one of the motivations for re-introducing SpanContext, among many, was that it would provide the datatype we could use to solve https://github.com/opentracing/opentracing.github.io/issues/28. I don't think the current change does that, because there is no API to deserialize just the SpanContext, without starting a new span.

lookfwd:
I like everything that weakens the `Finish()` semantics. I think I would like to see some form of diagram that demonstrates this change. The idea is that you have part of the `Span` API that outlives a `Span`? This part of the `Span` API includes baggage and parent-ship.

image

tracer.go
bensigelman: there are some nice things about that pattern, agreed.

bensigelman:
@yurishkuro re:

> one of the motivations for re-introducing SpanContext, among many, was that it would provide the datatype we could use to solve opentracing/opentracing.github.io#28. I don't think the current change does that, because there is no API to deserialize just the SpanContext, without starting a new span.

This change leaves the door open to some simplifications for the API. The data in a `SpanContext` is 1:1 with the state represented by `Inject` carriers. As such, we could do something like this:

```
type SpanContext interface {
	SetBaggageItem(key, val string)
	BaggageItem(key string) string
}

type Tracer interface {
	StartSpan(operationName string, opts ...StartSpanOption) Span
	Inject(sc SpanContext, format interface{}, carrier interface{}) error
}

type BuiltinFormat byte

const (
	// NOTE: an in-memory SpanContext interface descending from the appropriate
	// Tracer type becomes a format, much like the InMemoryCarrier we've
	// discussed previously.
	LocalSpanContext BuiltinFormat = iota
	Binary
	TextMap
)

type SpanReferenceType int

const (
	ParentRef SpanReferenceType = iota
	StartedBeforeRef
	EndedBeforeRef
	// NOTE: other sorts of span refererences go here per
	// https://github.com/opentracing/opentracing.github.io/issues/28
)

type SpanReference struct {
	RefType SpanReferenceType
	Format  interface{}
	Carrier interface{}
}

type StartSpanOptions struct {
	StartTime     time.Time
	InitialTags   map[string]interface{}
	References    []SpanReference
}

type StartSpanOption func(*StartSpanOptions)

func ReferenceParentSpanContext(sc SpanContext) StartSpanOption {
	return ReferenceParent(LocalSpanContext, sc)
}

// TODO: etc, etc, for other SpanReferenceTypes
func ReferenceParent(format, carrier interface{}) StartSpanOption {
	return func(o *StartSpanOptions) {
		o.References = append(o.References, SpanReference{
			RefType: ParentRef,
			Format:  format,
			Carrier: carrier,
		})
	}
}

// TODO: etc, etc, for the other StartSpanOption functions.
```

I.e., there's no more `Join`! Just `StartSpan(opName, ReferenceParent(format, carrier))`.

I would like to avoid an explicit `carrier -> SpanContext` API, and doing something like the above would accomplish same.

yurishkuro:
@bensigelman span context as a carrier format - ingenious!

One problem I see with the above is that by losing a distinction between Start and Join we'd be breaking Zipkin's "one span per RPC" model. Unless we say that a combination of `ParentRef` and non-`LocalSpanContext` can be an indication to reuse span ID, which is quite confusing given that the struct refers to "parent". I wish we could drop "one span per RPC" as an option, but I don't see Zipkin changing that any time soon.

FWIW, adding `F(format, carrier) -> SpanContext` API to tracer might be a simpler option. The above is clever, but not as easy to grasp as an explicit "deserialize" method, imho. And it actually increases the overall API surface by adding so many new types.

bensigelman:
@yurishkuro 

> Unless we say that a combination of ParentRef and non-LocalSpanContext can be an indication to reuse span ID, which is quite confusing given that the struct refers to "parent". I wish we could drop "one span per RPC" as an option, but I don't see Zipkin changing that any time soon.

Well, maybe we shouldn't use the term "Parent" here. `RPCClientRef` would be specific and keep the Zipkin code from reading poorly. I don't mind that the method is called `StartSpan`, btw (even if the span exists in two processes with different logical start times).

Anyway, a slightly revised mini-proposal along these lines:

```
import "time"

type SpanContext interface {
	SetBaggageItem(key, val string)
	BaggageItem(key string) string
}

type Tracer interface {
	StartSpan(operationName string, opts ...StartSpanOption) Span

	Inject(sc SpanContext, format interface{}, carrier interface{}) error
	Extract(format, carrier interface{}) (SpanContext, error)
}

type BuiltinFormat byte

const (
	// NOTE: an in-memory SpanContext interface descending from the appropriate
	// Tracer type becomes a format, much like the InMemoryCarrier we've
	// discussed previously.
	LocalSpanContext BuiltinFormat = iota
	Binary
	TextMap
)

type SpanReferenceType int

const (
	BlockedParentRef SpanReferenceType = iota
	RPCClientRef
	StartedBeforeRef
	EndedBeforeRef
	// NOTE: other sorts of span refererences go here per
	// https://github.com/opentracing/opentracing.github.io/issues/28
)

type SpanReference struct {
	RefType SpanReferenceType
	SpanContext
}

type StartSpanOptions struct {
	OperationName string
	StartTime     time.Time
	InitialTags   map[string]interface{}
	References    []SpanReference
}

type StartSpanOption func(*StartSpanOptions)

func ReferenceSpanContext(refType SpanReferenceType, sc SpanContext) StartSpanOption {
	return func(o *StartSpanOptions) {
		o.References = append(o.References, SpanReference{refType, sc})
	}
}

// TODO: etc, etc, for the other StartSpanOption functions.
func OperationName(n string) StartSpanOption {
	return func(o *StartSpanOptions) {
		o.OperationName = n
	}
}
```

If I have time in the next 36 hours I will try to flesh out the above.

yurishkuro:
I like this approach better. Looks like it doesn't need `LocalSpanContext BuiltinFormat` anymore. Inject/extract are symmetric again.

Minor nit: when using the functional option pattern, if we define `StartSpanOption` as interface, e.g.

```go
type StartSpanOption interface {
    Apply(* StartSpanOptions)
}
```

and implement various options explicitly with structs instead of lambdas, then the call to `StartSpan` will require 0 memory allocations. Otherwise, when a lambda captures surrounding variable(s), it requires an alloc (afaik).

bensigelman:
Well, I have jury duty next week (:us: :us: :us: :us:) ... hopefully that means I will just be sitting in a courthouse waiting to be called and hacking a little further on this. In which case I will work out the implications for `basictracer` and some example calling code. My sense, though, is that the approach here is a net improvement. And even though it introduces one new concept (`SpanContext`), it also *removes* a bunch of asymmetry and opportunity for confusion in the current set of concepts, esp around `Join`. And there will only be one way to create a Span: another sign of progress IMO.

I'm going to cc some people here since I realized they may not get Gitter notifications: @dkuebric @bg451 @adriancole @michaelsembwever @dawallin @bcronin @jmacd this is an important change and I'd appreciate if you took a look. If it's a problem that this RFC is in Go, I can try and translate to something analogous in Java or Python or whatever.

dawallin:
I like the spanContext. For me it makes it more explicit in Inject that only the context part is passed on to the carrier.  It is also good if Inject and Join/Extract could stay symmetric. 

I like the introduction of multiple parents, but in the solution above, it looks to me that all parents are needed at the start of the span. Is this always true? Shouldn't it be possible to add parents later or that introduces other problems ?

What is the parent-after-finish problem I saw mentioned somewhere? Is it not ok to let a new span be created after the parent is finished? This would be the standard scenario in "fire and forget" eventdriven architecture.

What are the SpanReferenceTypes (BlockedParentRef, RPCClientRef, StartedBeforeRef, 
EndedBeforeRef)? They seem specialized to solve a specific problem?

bg451:
I spent time thinking about this yesterday and I'm a really big fan of this change. It solves a lot of the problems around the semantics surrounding `Finish`, and will make async type programs so much easier to instrument. I really like the idea of `Join` being removed per one of the above proposals as well, and `Extract` is a lot easier to understand imo. I'm going to spend some more time thinking about this today and see if there's anything I disagree with, but overall 👍 

> What is the parent-after-finish problem I saw mentioned somewhere?

@dawallin Since the finish semantics means a) this is the last call on the span and b) I won't be using the span anymore, a few opentracing implementations recycle the underlying span objects after `Finish()` is called. As a result, there were some use after finish cases where a child span using an recycled parent span, leading to a variety of issues. Check out https://github.com/opentracing/basictracer-go/issues/23 for an example.

bensigelman:
@yurishkuro @bg451 @dawallin @lookfwd @jmacd 

Since you've all commented on this, I pushed another change to make symmetric Inject+Extract.

I think this is a big improvement. The actual `CausalReference` enum values are lousy and inconsistent here, but the basic concept seems strong to me.

@dawallin asked:

> I like the introduction of multiple parents, but in the solution above, it looks to me that all parents are needed at the start of the span. Is this always true? Shouldn't it be possible to add parents later or that introduces other problems ?

This should be possible, yes. I think we need to decide whether those refs should happen via a new method on `Span` or as an extension of the `LogKV` sort of mechanism described in https://github.com/opentracing/opentracing.github.io/issues/96 . I would propose we leave things as-is until we have clearer use cases for references that arrive after Span start time.

bensigelman:
(and I updated https://github.com/opentracing/basictracer-go/pull/29 to illustrate the ramifications for calling code and implementations)

gocontext.go
```diff
@@ -48,10 +42,12 @@ func StartSpanFromContext(ctx context.Context, operationName string) (Span, cont
...
-		OperationName: operationName,
-		Parent:        parent,
-	})
+	opts := []StartSpanOption{}
+	if span := SpanFromContext(ctx); span != nil {
+		opts = append(opts, Reference(RefBlockedParent, span.SpanContext()))
+	}
+	span := tracer.StartSpan(
+		operationName,
+		opts...)
```
yurishkuro: I'd prefer

```go
var span Span
if parentSpan := := SpanFromContext(ctx); parentSpan != nil {
    span = tracer.StartSpan(operationName, Reference(RefBlockedParent, parentSpan.SpanContext()))
} else {
    span = tracer.StartSpan(operationName)
}
```

avoids allocation/append of a slice

span.go
```diff
@@ -6,10 +6,54 @@ import (
...
+}
+
 // Span represents an active, un-finished span in the OpenTracing system.
 //
 // Spans are created by the Tracer interface.
 type Span interface {
+	// Returns the SpanContext for this Span. Note that the return value of
+	// SpanContext() is still valid after a call to Span.Finish(), as is a call
+	// to Span.SpanContext() after a call to Span.Finish().
+	SpanContext() SpanContext
```
yurishkuro: maybe just `Context()`? Since it's always used with `span.Context()`

tracer.go
yurishkuro: after the last change the are symmterical

tracer.go
```diff
@@ -121,6 +118,74 @@ type StartSpanOptions struct {
...
+
+// CausalReference pairs a reference type and a referent SpanContext. See the
+// CausalReferenceType documentation.
+type CausalReference struct {
+	RefType CausalReferenceType
+	SpanContext
+}
+
+// StartSpanOption instances (zero or more) may be passed to Tracer.StartSpan.
+type StartSpanOption func(*StartSpanOptions)
```
yurishkuro: I would like this to be an interface, not a function. And the options below to be structs, not lambdas.

span.go
djspoons: +1

bensigelman:
Okay, per today's weekly hangout, we are going to move forward with this. I will clean things up and ping the PR when it's ready for a detailed look.

span.go
jmacd: I wonder why not embed SpanContext in Span, so that a Span can be passed as an argument to functions accepting SpanContext?

span.go
```diff
@@ -174,10 +184,12 @@ func CanonicalizeBaggageKey(key string) (string, bool) {
...
+// parent (per StartSpanOptions.Parent) and an operation name per
+// Span.SetOperationName.
 func StartChildSpan(parent Span, operationName string) Span {
-	return parent.Tracer().StartSpanWithOptions(StartSpanOptions{
-		OperationName: operationName,
-		Parent:        parent,
-	})
+	return parent.Tracer().StartSpan(
+		operationName,
+		Reference(RefBlockedParent, parent.SpanContext()),
```
jmacd: The comment for RefBlockedParent implies this is a change of semantics. StartChildSpan says nothing about blocking its parent.

tracer.go
```diff
@@ -121,6 +118,74 @@ type StartSpanOptions struct {
...
+
+// Reference returns a StartSpan() option that adds a reference from the
+// newly-started span to a referent Span (parent or otherwise).
+func Reference(t CausalReferenceType, sc SpanContext) StartSpanOption {
```
jmacd: I would prefer if this were

func (rt CausalReferenceType) To(sc SpanContext) {
...
}

and then you'd write

tracer.StartSpan(operationName, ParentRef.To(parent))

yurishkuro:
> Do an %s/SpanContext/SpanMetadata/g

@bensigelman iirc you've always resisted the term "metadata" :-)

bensigelman:
@yurishkuro 

> @bensigelman iirc you've always resisted the term "metadata" :-)

I did some soul-searching... my options, as I perceived them:

- SpanContext (too verbose, too redundant)
- Context (too general, terribly confusing in Go)
- Baggage (too weird, doesn't feel right for `References` which is important)
- Baton (also too weird, also doesn't feel right for `References`)
- Coordinates (also too weird, doesn't feel right for `SetBaggageItem` (though great for `References`)
- Metadata (I don't like it but doesn't have the problems above)

Other suggestions??

bensigelman:
Anyway, this is now cleaned up and ready for a nitpicky review... @yurishkuro @jmacd @bg451 @dawallin @lookfwd

span.go
```diff
@@ -52,47 +80,12 @@ type Span interface {
...
-	// system), and with it some powerful costs: use this feature with care.
-	//
-	// IMPORTANT NOTE #1: SetBaggageItem() will only propagate trace
-	// baggage items to *future* children of the Span.
-	//
-	// IMPORTANT NOTE #2: Use this thoughtfully and with care. Every key and
-	// value is copied into every local *and remote* child of this Span, and
-	// that can add up to a lot of network and cpu overhead.
-	//
-	// IMPORTANT NOTE #3: Baggage item keys have a restricted format:
```
bensigelman: FYI, I got rid of the baggage key restrictions in this PR.

README.md
```diff
@@ -105,18 +110,25 @@ It's always possible to create a "root" (parentless) `Span`.
...
+            // Create a span referring to the RPC client.
+            serverSpan = opentracing.StartSpan(
+                appSpecificOperationName,
+                opentracing.RefRPCClient.Point(wireMetadata))
+        } else {
+            // Create a root span.
+            serverSpan = opentracing.StartSpan(appSpecificOperationName)
         }
+
         var goCtx context.Context = ...
```
yurishkuro: nit: s/goCtx/ctx/

README.md
```diff
@@ -105,18 +110,25 @@ It's always possible to create a "root" (parentless) `Span`.
...
         goCtx, _ = opentracing.ContextWithSpan(goCtx, serverSpan)
```
yurishkuro: simplify: `ctx := opentracing.ContextWithSpan(ctx, serverSpan)` (it no longer returns 2 values)

README.md
```diff
@@ -105,18 +110,25 @@ It's always possible to create a "root" (parentless) `Span`.
...
-        defer serverSpan.Finish()
         ...
```
yurishkuro: why did we lose the call to Finish()?

propagation.go
```diff
@@ -65,17 +65,17 @@ const (
...
+	//    span, err := tracer.Extract("opName", TextMap, carrier)
 	//
 	TextMap
 )
 
 // TextMapWriter is the Inject() carrier for the TextMap builtin format. With
-// it, the caller can encode a Span for propagation as entries in a multimap of
-// unicode strings.
+// it, the caller can encode a SpanMetadata for propagation as entries in a
+// multimap of unicode strings.
```
yurishkuro: I missed this before - I think we must not require multi-map semantics from the TextMap carrier implementation, and tracers should not expect the underlying transport to support multi-map headers.

tracer.go
```diff
@@ -121,6 +126,110 @@ type StartSpanOptions struct {
...
+	//
+	RefBlockedParent ReferenceType = iota
+
+	// RefRPCClient is a special case of RefBlockedParent for the server side
+	// of RPCs to refer to the client side of RPCs. Timing diagram:
+	//
+	//     [-Remote Client Referent------]
+	//                [-New Server Span-]
+	//
+	RefRPCClient
```
yurishkuro: would it make sense to explain why we special-case this form of parent-child? E.g. one reason is Zipkin's model of one span ID per RPC.

tracer.go
```diff
@@ -121,6 +126,110 @@ type StartSpanOptions struct {
...
+
+	// RefStartedBefore refers to a Span that merely started before the
+	// newly-started Span. This is the weakest of the ReferenceType causality
+	// assertions. Timing diagram:
+	//
+	//     [-Referent-]
+	//          [-New Span-]
+	//
+	RefStartedBefore
```
yurishkuro: would be good to reference some example where capturing this relationship is useful.

tracer.go
```diff
@@ -121,6 +126,110 @@ type StartSpanOptions struct {
...
+type StartSpanOption interface {
+	Apply(*StartSpanOptions)
+}
+
+// ReferenceType is an enum type describing different categories of
+// relationships between two Spans. If Span-2 refers to Span-1, the
+// ReferenceType describes Span-1 from Span-2's perspective. For example,
+// RefBlockedParent means that Span-1 caused Span-2, and that it's blocked on
+// Span-2's finish.
+type ReferenceType int
```
yurishkuro: The references can also be used for _meta-trace_ linkage, e.g. an http request (trace-1) enqueuing a background job (trace-2). Can we clarify in the comments that span-1 and span-2 do not have to belong to the same trace?

tracer.go
```diff
@@ -121,6 +126,110 @@ type StartSpanOptions struct {
...
+	//
+	RefStartedBefore
+
+	// RefFinishedBefore refers to a Span that finished (and started) before
+	// the newly-created Span. Timing diagram:
+	//
+	//     [-Referent-]
+	//                    [-New Span-]
+	//
+	RefFinishedBefore
```
yurishkuro: similarly, maybe mention as example that this ref type can be used to model joins.

tracer.go
```diff
@@ -121,6 +126,110 @@ type StartSpanOptions struct {
...
+// IntraTrace returns true if the referent is in the same trace as the referee
+// (i.e., if the referee should adopt the referent's trace-scoped metadata,
+// like a `trace_id`).
+func (r ReferenceType) IntraTrace() bool {
+	return r <= RefFinishedBefore
+}
+
+// Point returns a StartSpanOption that describes a SpanReference (from the
+// Span that's about to be started to the `referent`).
+func (r ReferenceType) Point(referent SpanMetadata) SpanReference {
```
yurishkuro: had to read it a couple times to understand. I think something like `NewRef` name would be more intuitive.

tracer.go
```diff
@@ -121,6 +126,110 @@ type StartSpanOptions struct {
...
+func (r SpanReference) Apply(o *StartSpanOptions) {
+	o.References = append(o.References, r)
+}
+
+// IntraTrace returns true if the referent is in the same trace as the referee
+// (i.e., if the referee should adopt the referent's trace-scoped metadata,
+// like a `trace_id`).
+func (r ReferenceType) IntraTrace() bool {
+	return r <= RefFinishedBefore
+}
```
yurishkuro: not sure why we need this, or if it's even correct if we allow cross-trace references in `RefStartedBefore`

tracer.go
```diff
@@ -121,6 +126,110 @@ type StartSpanOptions struct {
...
+type Tags map[string]interface{}
+
+// Merge incorporates the keys and values from `other` into this `Tags`
+// instance, then returns same.
+func (t Tags) Merge(other Tags) Tags {
+	for k, v := range other {
+		t[k] = v
+	}
+	return t
+}
```
yurishkuro: this doesn't feel like it belongs in the API

yurishkuro:
@bensigelman Between these two:

>    SpanContext (too general, too verbose, too redundant)
>    SpanMetadata (I don't like it but doesn't have the problems above)

I have a preference for SpanContext. It doesn't feel to me that "too general, too verbose, too redundant" qualifications apply to "context" more than to "metadata". A context goes well with the whole concept of distribute context propagation. And while trace/span IDs are well described by "metadata", the baggage isn't.

README.md
bensigelman: I went with `ctx = opentracing.ContextWithSpan(context.Background(), serverSpan)`

README.md
bensigelman: oversight, fixed!

propagation.go
bensigelman: I agree. I'll make that change here.

tracer.go
bensigelman: yes, that is definitely the idea here... I will clarify.

tracer.go
bensigelman: I added this so that simple things like `basictracer` know when to call a reference a "parent id" vs when to ignore the reference entirely.

tracer.go
bensigelman: I think `PointTo` might be clearer... I'll try that. Naming for this stuff was subtle.

tracer.go
bensigelman: yeah, that was just a move. agree that it's not needed, I'll get rid of it.

tracer.go
```diff
@@ -121,6 +127,118 @@ type StartSpanOptions struct {
...
+// Apply satisfies the StartSpanOption interface.
+func (r SpanReference) Apply(o *StartSpanOptions) {
+	o.References = append(o.References, r)
+}
+
+// IntraTrace returns true if the referent is in the same trace as the referee
+// (i.e., if the referee should adopt the referent's trace-scoped metadata,
+// like a `trace_id`).
+func (r ReferenceType) IntraTrace() bool {
+	return r <= FinishedBefore
```
bg451: I might be missing something, but why not just `r != FinishedBefore` instead of relying on the enum order?

bensigelman: All of the `ReferenceType`s are intra-trace references (as opposed to some of the sorts of references described in https://github.com/opentracing/opentracing.github.io/issues/28)...

tracer.go
```diff
@@ -121,6 +127,118 @@ type StartSpanOptions struct {
...
+}
+
+// Apply satisfies the StartSpanOption interface.
+func (r SpanReference) Apply(o *StartSpanOptions) {
+	o.References = append(o.References, r)
+}
+
+// IntraTrace returns true if the referent is in the same trace as the referee
+// (i.e., if the referee should adopt the referent's trace-scoped metadata,
+// like a `trace_id`).
```
yurishkuro: A bit confused by the word choice of "referee" (m-w: "one to whom a thing is referred") and "referent" (m-w: "one that refers or is referred to").  Usually I see "referrer" and "referee".

tracer.go
yurishkuro: Is it the intention that this method will always be correct for any enum value?  Are we treating this enum as locked down or extensible?

Also, the semantics of having a "parent" is not clear, especially in the case of FinishedBefore. For example, if I want to model a join, i.e. fire N async requests/spans, wait for all of them to finish, and then start an "after-join" span, then that span is still intra-trace, whose parent is the higher level span that did the async forking, and the N spans will be FinishedBefore referees, intra-trace, but not parents. So I'm not sure what `IntraTrace` method tells us in that scenario. Is seems the real question is how to select the parent span in case of multiple referees provided.

bensigelman: Upon reflection, I think it would be better to get rid of `IntraTrace()` entirely. It's not needed for this PR and IMO it's too difficult to resolve threads like this one without more concrete use cases.

tracer.go
bensigelman: yeah, this was a documentation bug... I updated things (and did `s/SpanReference.Metadata/SpanReference.Referee/` while I was at it).

bcronin:
*(Jumping in a little late here, so apologies if I'm repeating any prior discussion of this...)*

Subjective comments  here, but the below code fragment seems a bit awkward to me:

```
tracer.StartSpan("my_span", opentracing.BlockedParent.PointTo(parentSpan.Metadata()))
```

I personally would think `LinkTo` or `LinksTo` would be more readily describe to me what's happening. Also, I did see the comment on postponing the decision on allowing multiple parents / `LogKV`, but regardless, a method on `Span` would best convey what's happening to me:

```
span = tracer.StartSpan("my_span")
span.LinkTo(opentracing.BlockedParent, parentSpan.Metadata())
```

My two cents :)

jmacd:
+1 for LinkTo
ᐧ

yurishkuro:
@bcronin I'd rather not separate span creation from setting a parent span reference. Some implementation may want to have span context to be immutable.

Re `LinkTo`, I think a simpler syntax would be
  * name enums with `Ref` as in the earlier commits, i.e. `BlockParentRef`
  * have a function `func BlockedParent(referee SpanContext)`

If overloading were supported, I would've also added `func BlockedParent(referee Span)`

tracer.go
```diff
@@ -121,6 +127,111 @@ type StartSpanOptions struct {
...
+// ReferenceType is an enum type describing different categories of
+// relationships between two Spans. If Span-2 refers to Span-1, the
+// ReferenceType describes Span-1 from Span-2's perspective. For example,
+// BlockedParent means that Span-1 caused Span-2, and that it's blocked on
+// Span-2's finish.
+//
+// NOTE: Span-1 and Span-2 do *not* necessarily depend on each other for
+// completion; e.g., Span-2 may be part of a background job enqueued by Span-1,
+// or Span-2 may be sitting in a distributed queue behind Span-1.
+type ReferenceType int
```
yurishkuro: nit: SpanReferenceType would be a more accurate type name

tracer.go
```diff
@@ -121,6 +127,111 @@ type StartSpanOptions struct {
...
+	// (where Referee represents stage N of same).
+	FinishedBefore
+)
+
+// SpanReference pairs a ReferenceType and a referee ("the one who is referred
+// to") SpanMetadata. See the ReferenceType documentation.
+type SpanReference struct {
+	Type    ReferenceType
+	Referee SpanMetadata
+}
```
yurishkuro: @bensigelman the main outstanding issue I see here is clarifying the semantics of "who is the parent" when multiple span references are given to `StartSpan()`. It seems that establishing some rules about "the parent" based on RefType priorities or SpanRef ordering in the call is going to be quite arbitrary and brittle. On the other hand, we can explicitly withdraw judgement about "the parent" and leave it to implementations (but we should explicitly say so). I like it because it removes the "tree" restriction from the trace model in favor of a more generic DAG.

I wonder if it would hurt interoperability / usability.

dawallin:
I also prefere the naming SpanContext more than SpanMetadata.

I have a problem with PointsTo, Reference and the enums BlockedParent, StartedBefore and StartedAfter. All feel to technical and are difficult to understand without the documentation and the examples. I like the proposed  “link” name instead of “ref”. I then think it could be clearer to say what the span is instead of what its parent is. All the suggested ReferenceTypes describe the parent. If we instead look at the span it could be something like: RPCServerSpan, RPCClientSpan, EventSpan and SequenceSpan.

adriancole:
We discussed links before with regards to disconnected, but associated
traces. Using the term "link" here for parent/child in the same trace might
cause interference if we broaden the topic of links.

bensigelman:
Thoughts:
- ok, fine, fine, `SpanContext` it is :)
- @yurishkuro I am ok not-specifying a way of determining the "parent" when there are multiple reference spans
- I am not enjoying the task of coming up with the perfect set of enum values: perhaps we should just choose the "minimum viable subset" that we *know* we need and leave it at that for now. As we gain experience we can add others. (E.g., perhaps it should be more of a bitmask and less of an enum... I can certainly see some benefits there)

I'll make another iteration tomorrow... other concrete proposals welcome!

bensigelman:
I've thought a bunch more about the semantics for the references (regardless of what we call them). A few conclusions:

- there's no point in specifying `StartedBeforeRef` or similar... all referenced Spans must have already started or we wouldn't have a context object for them.
- it's feasible to do all of this with bitmasking, but I think that's too complicated to explain... when I prototyped it I felt like the comments were getting jargony
- "parent" was a poor choice of terminology in dapper... but it's also something a lot of us think about and say, so I like that term more than `FinishesAfterRef` or similar (which is really the same idea as a blocking parent)
- per previous comments, I don't want to bite off the not-in-the-same-trace relationships at this stage

All of this has me thinking we should only support three sorts of references for now... (Per the above, all of them imply `StartedBefore` which we can thus leave out)

1. `Parent` (equivalent to `FinishesAfter`)
1. `RPCClient` (a special case, but an important one)
1. `FinishedBefore` (for distributed queues and so on)

... or, if we express things from the perspective of the Span that's about to start:

1. `Child` (equivalent to `FinishesBefore`)
1. `RPCServer`
1. `StartsAfter`

I'm going to sleep on this, but those are my thoughts at the moment.

bensigelman:
... well, I did sleep on this and I had a few more thoughts about it.

- The reason we're making `RPCClient` its own reference type is so that tracers like Zipkin can determine whether to create a new span_id or reuse an existing one... we could (and, I'd argue, should) just use the `Tags` mechanism for this.
- The only cases I feel comfortable distinguishing between at this point are the following:
  1. Fire-and-forget: all we know is that the new Span starts after the Span we're referencing
  1. Blocking children: the Span we're starting is supposed to finish before the (parent) Span we're referencing
  1. Pipelines: the Span we're referencing must have already finished

I'll try to codify this in Go later today.

yurishkuro:
> The reason we're making RPCClient its own reference type is so that tracers like Zipkin can determine whether to create a new span_id or reuse an existing one... we could (and, I'd argue, should) just use the Tags mechanism for this.

There is no guarantee that the tags will be provided at span construction time and not in a separate statement. In the latter case it's too late to assign span ID.

bensigelman:
@yurishkuro 

> > The reason we're making RPCClient its own reference type is so that tracers like Zipkin can determine whether to create a new span_id or reuse an existing one... we could (and, I'd argue, should) just use the Tags mechanism for this.
> There is no guarantee that the tags will be provided at span construction time and not in a separate statement. In the latter case it's too late to assign span ID.

If the caller is able to provide the reference at span start time, then they could equivalently provide the tag at span start time... and it seems likely to me that there will eventually be a way to add references *after* span start time, too. Furthermore, we already use span tags to differentiate the RPC vs non-RPC case.

I like the idea of the references being strictly about causality and not about how many process boundaries were crossed... seems like a cleaner separation of concerns.

dawallin:
I like this. The RPCClient was the ref type I had most problem with.

bensigelman:
I pushed some "minimum viable reference types" as well as a helper for the RPC server span case (in `ext/`). PTAL.

ext/tags.go
```diff
@@ -88,16 +88,31 @@ type SpanKindEnum string
...
 	span.SetTag(string(tag), value)
 }
 
+type rpcServerOption struct {
+	clientMetadata opentracing.SpanMetadata
+}
+
+func (r rpcServerOption) Apply(o *opentracing.StartSpanOptions) {
+	opentracing.Parent(r.clientMetadata).Apply(o)
+	(opentracing.Tags{SpanKind: SpanKindRPCServer}).Apply(o)
```
yurishkuro: probably need `string(SpanKind)`

+1

tracer.go
```diff
@@ -121,6 +125,110 @@ type StartSpanOptions struct {
...
+	// Span starts. For instance, one might use SequencePredecessorRefs to describe
+	// pipeline stages separated by queues.
+	//
+	// A timing diagram for a SequencePredecessorRef that must finish before the new
+	// Span can start:
+	//
+	//     [-SequencePredecessorRef Span-]  [-New Span-]
+	//
+	// See opentracing.SequencePredecessor()
+	SequencePredecessorRef
```
yurishkuro: how critical is it to keep the "Sequence" prefix?

tracer.go
```diff
@@ -121,6 +125,110 @@ type StartSpanOptions struct {
...
+	//     [-ParentRef Span------]
+	//          [-New Span------]
+	//
+	// A timing diagram for a ParentRef that's not blocked on the new Span:
+	//
+	//     [-ParentRef Span-]
+	//                  [-New Span-]
+	//
+	// See opentracing.Parent()
+	ParentRef ReferenceType = iota
```
yurishkuro: Eventually this would go to the Spec, but for now I think it's worth clarifying a few semantic points:

1. If there is no ParentRef in the StartSpanOptions, a new trace / root span is created
1. ParentRef may be accompanied by other Refs, e.g. below Span C can be started with Parent(A) and Predecessor(B)

```
[--Span A------------------------]
   [--Span B--]  [--Span C--]
```

tracer.go
yurishkuro: Also, can the predecessor be from a different trace?

yurishkuro:
I'm ok with the latest version, modulo minor renaming (meta->context, SpanReferenceType) and a few clarifications in the comments.

tracer.go
bensigelman: I thought about `Predecessor` (along with about a dozen other possibilities)... my concern was that it didn't make it sufficiently clear that the predecessor span *finished* before the new span started. And, per your second point, it didn't make it clear that it's from the same logical trace... e.g., there are plenty of "Predecessors" from a timing standpoint that don't prevent the new span from being created but are in different logical traces, and IMO those should be a different reference type.

Some alternative names I came up with were `PreviousStepRef`, `PreviousStageRef`, `MustBeFinishedRef`, and `Last(Step|Stage)Ref`. YMMV?

tracer.go
bensigelman: I don't agree with (1)... I've seen a pattern where there is no overarching root span, just a sequence like

```
[-Span A-]  [-Span B-]       [-Span C-] [-Span D-]
```

These are part of a single trace but don't have an explicit parent.

I am fine saying that a new Span with no references is the root of its own trace, though.

tracer.go
yurishkuro: ok, that clarifies your objection above. Definitely need a comment saying SequencePredecessorRef means it's the same trace.

yurishkuro:
LGTM

bensigelman:
I'm going to leave this unmerged until the analogous change is in for the docs site (which should be the "master copy").

bensigelman:
This has been updated per https://github.com/opentracing/opentracing.github.io/pull/100 and #85 + #86

bensigelman:
I will merge this tomorrow if there are no objections.

(cc @basvanbeek since I noticed you were kind enough to fix recent breakage in https://github.com/openzipkin/zipkin-go-opentracing/pull/5 ... Bas, I am happy to help with the downstream changes to support what's here if that's helpful.)

basvanbeek:
@bensigelman... that would be great as I'm quite busy at the moment and not sure how quickly I'd be able to look at it myself

bensigelman:
@basvanbeek I'll take a look and try to prepare a PR for you to review *before* I merge this `opentracing-go` change

README.md
```diff
@@ -105,18 +110,26 @@ It's always possible to create a "root" (parentless) `Span`.
...
         if err != nil {
-            // Create a root span if necessary
-            serverSpan = opentracing.StartTrace("serverSpan")
+            // Optionally record something about err here
+        }
+        if wireContext != nil {
+            // Create a span referring to the RPC client.
+            serverSpan = opentracing.StartSpan(
+                appSpecificOperationName,
+                opentracing.RPCServerOption(wireContext))
```
bg451: `s/opentracing/ext/` since RPCServerOption is in the ext package.

bensigelman: good catch, fixed.

bensigelman:
@basvanbeek see https://github.com/openzipkin/zipkin-go-opentracing/pull/9

I will plan to merge this pr #82 tomorrow unless you have concerns about the zipkin change.

PR #84, opened 2016-06-24
Add helper type TextMapCarrier; Add "error" tag

yurishkuro:
@bensigelman I think this is not controversial

### DIFF
--- a/propagation.go
+++ b/propagation.go
@@ -112,8 +112,27 @@ type TextMapReader interface {
 	ForeachKey(handler func(key, val string) error) error
 }
 
+// TextMapCarrier allows the use of regular map[string]string
+// as both TextMapWriter and TextMapReader.
+type TextMapCarrier map[string]string
+
+// ForeachKey conforms to the TextMapReader interface.
+func (c TextMapCarrier) ForeachKey(handler func(key, val string) error) error {
+	for k, v := range c {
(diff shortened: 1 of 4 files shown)


propagation.go
```diff
@@ -112,8 +112,27 @@ type TextMapReader interface {
 	ForeachKey(handler func(key, val string) error) error
 }
 
+// TextMapCarrier allows the use of regular map[string]string
+// as both TextMapWriter and TextMapReader.
```
bensigelman: technically this is in violation of the multimap spec, though the multimap spec is about to change... what to do??

bensigelman:
@yurishkuro this LGTM in the broader sense, though at the moment it does not fulfill the contract (per my comment). Thoughts?

propagation.go
yurishkuro: @bensigelman  I don't think we have multi-map *requirement* in the current API. The Writer doc says

```
	// Set a key:value pair to the carrier. Multiple calls to Set() for the
	// same key leads to undefined behavior.
```
so `map[string]string` satisfies that by overriding the key.

And the reader API says multiple keys *may* be returned, but map wouldn't do it.
```
	// NOTE: A single `key` may appear in multiple calls to `handler` for a
	// single `ForeachKey` invocation.
```

bensigelman:
:ship:

PR #85, opened 2016-06-30
Support ForeachBaggageItem

yurishkuro:
@bensigelman

### DIFF
--- a/ext/tags_test.go
+++ b/ext/tags_test.go
@@ -4,8 +4,8 @@ import (
 	"reflect"
 	"testing"
 
-	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
+	"github.com/opentracing/opentracing-go/mocktracer"
 )
 
 func assertEqual(t *testing.T, expected, actual interface{}) {
@@ -19,7 +19,7 @@ func TestPeerTags(t *testing.T) {
 	if ext.PeerService != "peer.service" {
 		t.Fatalf("Invalid PeerService %v", ext.PeerService)
 	}
-	tracer := noopTracer{}
+	tracer := mocktracer.New()
 	span := tracer.StartSpan("my-trace")
 	ext.PeerService.Set(span, "my-service")
 	ext.PeerHostname.Set(span, "my-hostname")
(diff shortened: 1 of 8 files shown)


ext/tags_test.go
```diff
@@ -40,68 +40,30 @@ func TestPeerTags(t *testing.T) {
...
 
-	rawSpan := span.(*noopSpan)
+	rawSpan := span.(*mocktracer.MockSpan)
 	assertEqual(t, "my-awesome-library", rawSpan.Tags["component"])
 	assertEqual(t, uint16(1), rawSpan.Tags["sampling.priority"])
 	assertEqual(t, true, rawSpan.Tags["error"])
 }
-
-// noopTracer and noopSpan with span tags implemented
-type noopTracer struct{}
```
bensigelman: oh good

span.go
```diff
@@ -87,6 +87,10 @@ type Span interface {
 	// See the `SetBaggageItem` notes about `restrictedKey`.
 	BaggageItem(restrictedKey string) string
 
+	// ForeachBaggageItem allows reading all baggage items stored in the span.
+	// The handler function will be called for each baggage key/value pair.
+	ForeachBaggageItem(handler func(k, v string))
```
bensigelman: sometimes these sorts of handler functions return a bool to indicate that the internal iterator can/should stop... thoughts?

bensigelman:
Basically LGTM. Would you be up for sending out the trivial change to support this in `basictracer-go` just to keep it up to date? If you don't have time I can do it.

span.go
```diff
@@ -87,6 +87,14 @@ type Span interface {
 	// See the `SetBaggageItem` notes about `restrictedKey`.
 	BaggageItem(restrictedKey string) string
 
+	// ForeachBaggageItem allows reading all baggage items stored in the span.
```
dominikh: Is it worth documenting whether ForeachBaggageItem has a stable iteration order, or if the order can differ between consecutive calls of ForeachBaggageItem?

yurishkuro: @dominikh https://github.com/opentracing/opentracing-go/commit/9e6d3e21876169b028e75016a19d75c4ad54cb11

bensigelman:
LGTM, thanks.

PR #86, opened 2016-06-30
Make MockTracer thread-safe

yurishkuro:
If people want to use MockTracer to test instrumentation of RPC frameworks, it needs to be thread-safe.

cc @bensigelman


yurishkuro:
@bensigelman now with tests and full coverage

Makefile
```diff
@@ -1,22 +1,32 @@
...
+test-and-lint: test lint
 
 .PHONY: test
 test:
 	go test -v -cover ./...
 
-.PHONY: fmt
-fmt:
-	go fmt ./...
+cover:
```
bensigelman: thank you!

bensigelman:
Looks great. Thanks again. :ship:

PR #87, opened 2016-07-06
allow RPCServerOption to be used without a parent client

basvanbeek:
it is very common for a server at the edge (ex. api gateway) to start an RPC server tagged span without a Client as the client might not have instrumentation. This change allows for clientContext to be nil

### DIFF
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -112,7 +112,9 @@ func (r rpcServerOption) Apply(o *opentracing.StartSpanOptions) {
 }
 
 // RPCServerOption returns a StartSpanOption appropriate for an RPC server span
-// with `client` representing the metadata for the remote peer Span.
+// with `client` representing the metadata for the remote peer Span if available.
+// In case client == nil, due to the client not being instrumented, this RPC
+// server span will be a root span.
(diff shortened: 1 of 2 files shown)


ext/tags.go
```diff
@@ -105,7 +105,9 @@ type rpcServerOption struct {
 }
 
 func (r rpcServerOption) Apply(o *opentracing.StartSpanOptions) {
-	opentracing.ChildOf(r.clientContext).Apply(o)
+	if r.clientContext != nil {
```
bensigelman: please update the docs for `RPCServerOption()` accordingly.

bensigelman:
LGTM after the docs change. Thanks!

ext/tags.go
```diff
@@ -112,7 +112,9 @@ func (r rpcServerOption) Apply(o *opentracing.StartSpanOptions) {
 }
 
 // RPCServerOption returns a StartSpanOption appropriate for an RPC server span
-// with `client` representing the metadata for the remote peer Span.
+// with `client` representing the metadata for the remote peer Span if available.
+// In case client == nil, due to the client not being instrumented, this RPC
+// server span will be a root span.
```
bensigelman: :ship:, thanks!

PR #88, opened 2016-07-06
fixed broken propagation of baggage in Mocktracer

basvanbeek:
after the SpanContext changes the baggage test broke in the Go kit tests using Mocktracer. This fixes it.


bensigelman:
Thanks.

PR #89, opened 2016-07-06
Improve comments per Gitter

bensigelman:
@dominikh care to take a quick look at this?

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -7,8 +7,9 @@ import "time"
 type Tracer interface {
 
 	// Create, start, and return a new Span with the given `operationName` and
-	// incorporate the given StartSpanOptions. (See StartSpanOption for all of
-	// the options on that front)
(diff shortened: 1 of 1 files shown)


tracer.go
```diff
@@ -7,8 +7,9 @@ import "time"
 type Tracer interface {
 
 	// Create, start, and return a new Span with the given `operationName` and
-	// incorporate the given StartSpanOptions. (See StartSpanOption for all of
-	// the options on that front)
+	// incorporate the given StartSpanOption `opts`. (Note that `opts` borrows
```
dominikh: Grammatical grey area. It's not just one StartSpanOption, but that's the type's name.

dominikh:
LGTM other than the one comment.

tracer.go
bensigelman: yeah, I thought I got the plural in there with `opts`. Agree about the ❔ area.

PR #90, opened 2016-07-08
Introduce Tag type for single-tag StartSpanOptions

jmacd:
Avoids an allocation in RPCServerOption.

### DIFF
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -23,11 +23,13 @@ var (
 
 	// SpanKindRPCClient marks a span representing the client-side of an RPC
 	// or other remote call
-	SpanKindRPCClient = SpanKindEnum("client")
+	SpanKindRPCClientEnum = SpanKindEnum("client")
+	SpanKindRPCClient     = opentracing.Tag{string(SpanKind), SpanKindRPCClientEnum}
 
 	// SpanKindRPCServer marks a span representing the server-side of an RPC
 	// or other remote call
-	SpanKindRPCServer = SpanKindEnum("server")
+	SpanKindRPCServerEnum = SpanKindEnum("server")
+	SpanKindRPCServer     = opentracing.Tag{string(SpanKind), SpanKindRPCServerEnum}
 
 	//////////////////////////////////////////////////////////////////////
 	// Component name
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -108,7 +108,7 @@ func (r rpcServerOption) Apply(o *opentracing.StartSpanOptions) {
 	if r.clientContext != nil {
 		opentracing.ChildOf(r.clientContext).Apply(o)
 	}
-	(opentracing.Tags{string(SpanKind): SpanKindRPCServer}).Apply(o)
+	opentracing.Tag{string(SpanKind), SpanKindRPCServer}.Apply(o)
 }
 
 // RPCServerOption returns a StartSpanOption appropriate for an RPC server span
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -19,7 +19,7 @@ var (
 	//////////////////////////////////////////////////////////////////////
 
 	// SpanKind hints at relationship between spans, e.g. client/server
-	SpanKind = spanKindTag("span.kind")
+	SpanKind = spanKindTagName("span.kind")
 
 	// SpanKindRPCClient marks a span representing the client-side of an RPC
 	// or other remote call
@@ -37,14 +37,14 @@ var (
 
 	// Component is a low-cardinality identifier of the module, library,
 	// or package that is generating a span.
-	Component = stringTag("component")
+	Component = stringTagName("component")
 
 	//////////////////////////////////////////////////////////////////////
 	// Sampling hint
 	//////////////////////////////////////////////////////////////////////
 
 	// SamplingPriority determines the priority of sampling this Span.
-	SamplingPriority = uint16Tag("sampling.priority")
+	SamplingPriority = uint16TagName("sampling.priority")
 
 	//////////////////////////////////////////////////////////////////////
 	// Peer tags. These tags can be emitted by either client-side of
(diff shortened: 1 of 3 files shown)


ext/tags.go
```diff
@@ -108,7 +108,7 @@ func (r rpcServerOption) Apply(o *opentracing.StartSpanOptions) {
 	if r.clientContext != nil {
 		opentracing.ChildOf(r.clientContext).Apply(o)
 	}
-	(opentracing.Tags{string(SpanKind): SpanKindRPCServer}).Apply(o)
+	opentracing.Tag{string(SpanKind), SpanKindRPCServer}.Apply(o)
```
bensigelman: thinking aloud, we could make `SpanKindRPCServer` a `Tag` proper and allows callers to deref `SpanKindRPCServer.Key` and `.Value` as they wish...

bensigelman:
Curious what @yurishkuro thinks of this, too.

ext/tags.go
jmacd: Done.

tracer.go
```diff
@@ -266,3 +264,17 @@ func (t Tags) Apply(o *StartSpanOptions) {
 		o.Tags[k] = v
 	}
 }
+
+// Tag may be passed as a StartSpanOption to add a tag to new spans.
+type Tag struct {
```
yurishkuro: can we not make Tag also behave like tags in `ext`?  Then instead of

```
ext.SpanKind.Set(span, ext.SpanKindRPCServerEnum)
```

we could do 
```
ext.SpanKindRPCServer.Set(span)
```

jmacd: I had a similar thought. I would say it can be done separately from this PR.

yurishkuro: well, the reason I think it belongs here is because you're breaking the API with this change by replacing previous tag values with `xxxEnum` versions, so best to provide the final replacement.

Pretty much it just needs this method

```go
func (t Tag) Set(span Span) {
    span.SetTag(t.Key, t.Value)
}
```

jmacd: I agree, although I think we should also rename the existing types xxxTag in tags.go to xxxTagName, since there are a bunch of `func (x xxxTag) Set(s Span, v xxx)` methods already, just for the sake of maintainability.

yurishkuro: I didn't quite follow that - *that* does sound like a separate PR 😄

yurishkuro:
LG :tm:

tracer.go
```diff
@@ -266,3 +264,30 @@ func (t Tags) Apply(o *StartSpanOptions) {
 		o.Tags[k] = v
 	}
 }
+
+// Tag may be passed as a StartSpanOption to add a tag to new spans,
+// or its Set method may be used to apply the tag to an existing Span,
+// for example:
+//
+// tracer.StartSpan("opName", Tag{"Key", value})
```
dominikh: This renders wrong in godoc. The example code should be indented, the "or" should not.

tracer.go
```diff
@@ -116,7 +114,7 @@ type Tracer interface {
 
 // StartSpanOptions allows Tracer.StartSpan() callers and implementors a
 // mechanism to override the start timestamp, specify Span References, and make
-// Tags available at Span start time.
+// a single Tag or multiple Tags available at Span start time.
```
dominikh: I know what you're getting at, but this prose reads redundantly, and isn't quite correct. One can pass multiple singular `Tag` to this function, which makes it multiple `Tag`'s, but not `Tags`. Do we really want both types to exist?

ext/tags.go
```diff
@@ -19,30 +19,32 @@ var (
 	//////////////////////////////////////////////////////////////////////
 
 	// SpanKind hints at relationship between spans, e.g. client/server
-	SpanKind = spanKindTag("span.kind")
+	SpanKind = spanKindTagName("span.kind")
 
 	// SpanKindRPCClient marks a span representing the client-side of an RPC
 	// or other remote call
-	SpanKindRPCClient = SpanKindEnum("client")
+	SpanKindRPCClientEnum = SpanKindEnum("client")
```
dominikh: This variable name feels very inconsistent with all the other names (`Component`, ...). This also conflates tags and options, which I think makes the ext package harder to grasp.

jmacd: Sorry I should wait longer before merging changes. I'll be happy to apply the fixes you recommend. I'm sympathetic to your statement, and I would prefer if we called what are currently "tag names" instead "labels".

I'm not sure what you mean about conflating tags and options, though there is already some confusion over Tags (a field in StartSpanOptions, also a map[string]interface{} type).  My understanding is that OT uses "Tag" to refer to a pair of key and value.

To the specific comment here, about consistency, I'm not sure. There are only two SpanKind*Enum values here, and they are different types than the other names (Component, ...--tag names or what I would call labels) so I don't see a need for consistency. I'm not sure we need them, since SpanKindRPC(Server|Client).Value is equivalent.

tracer.go
jmacd: I don't feel that we need Tags, but I don't feel strongly about removing it either. I'm generally not sure how we find consensus here (or how long to wait for comments)...

tracer.go
jmacd: I'll fix. Thanks.

ext/tags.go
dominikh: The conflation I am referring to is that all of the other variables specify just part of a tag, while these two are also options. And the consistency issue with `Enum` I am having is that it feels like a workaround, so that the label can coexist with the option. Every other string value in this package doesn't have the `Enum` prefix; and `Enum` as a prefix feels inherently unidiomatic.

tracer.go
dominikh: A quick prose fix would be something along the lines of `and make instances of Tag or a Tags available at Span start time`. I'm not 100% happy with that one, either, though.

jmacd: Staging these updates here:
https://github.com/opentracing/opentracing-go/pull/94
I went with "specify one or more tags" to keep it simple, and I think that's clear enough.

dominikh: SGTM, leaving the final decision to @bensigelman and folks obviously.

ext/tags.go
jmacd: OK, I vote to eliminate the Enum values entirely. I really think this confusion is over whether "tag" refers to a tag name (one string) or a tag+value (string, interface{}).

Do you mind that SpanKindRPCServer and SpanKindRPCClient are in this file, and are not tag names but tag+values?

dominikh: I think it's fine to have them as tag+values; as the tags provided are the only two that should be set.

tracer.go
```diff
@@ -266,3 +264,30 @@ func (t Tags) Apply(o *StartSpanOptions) {
...
+// or its Set method may be used to apply the tag to an existing Span,
+// for example:
+//
+// tracer.StartSpan("opName", Tag{"Key", value})
+//
+//   or
+//
+// Tag{"key", value}.Set(span)
+type Tag struct {
+	Key   string
```
bensigelman: nit: here you call it a `Key` (my pref, btw), elsewhere you call it a `Name`. Let's choose one and stick with it (key! key!).

tracer.go
bensigelman: I like `... and make zero or more Tags available at Span start time`. Zero is the common case, after all.

bensigelman: eh, I take it back given the context of this comment... so maybe `... and make one or more Tags available ...`

ext/tags_test.go
```diff
@@ -75,3 +77,30 @@ func TestMiscTags(t *testing.T) {
 		"error":             true,
 	}, rawSpan.GetTags())
 }
+
+func TestRPCServerOption(t *testing.T) {
```
bensigelman: thanks

bensigelman:
LGTM aside from my key vs name nit. Please give 24h for others to chime in before merging, though.

PR #91, opened 2016-07-08
Introduce CarriedChildOfOption and CarriedFollowsFromOption

jmacd:
These are StartSpanOption values that hide a call to tracer.Extract. In case of error, they annotate the span with an error instead of returning it to the caller.

### DIFF
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -35,6 +35,8 @@ var (
 	// Component name
 	//////////////////////////////////////////////////////////////////////
 
+	CarriedContextError = stringTag("carried.context.error")
+
 	// Component is a low-cardinality identifier of the module, library,
 	// or package that is generating a span.
 	Component = stringTag("component")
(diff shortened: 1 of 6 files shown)


jmacd:
FYI Ben is hesitant to move quickly with helper functions like this until the dust settles in other repositories, which I think means he likes it but thinks it's premature.

ext/tags.go
```diff
@@ -106,11 +108,11 @@ type rpcServerOption struct {
 	clientContext opentracing.SpanContext
 }
 
-func (r rpcServerOption) Apply(o *opentracing.StartSpanOptions) {
+func (r rpcServerOption) Apply(t opentracing.Tracer, o *opentracing.StartSpanOptions) {
```
yurishkuro: passing tracer to the option doesn't feel clean

jmacd: I'm curious what you think of the user-visible API change, since Apply is not user-visible. If you like the user-visible change, do you see another way to do this?

I considered passing the Tracer directly to the CarriedXXXOption, which also feels not clean, because that makes it possible for the Extract tracer to differ from the StartSpan tracer (and requires the user to pass the same thing twice).

I basically agree with you, but I think a little uncleanliness in an internal API is better than requiring users to call Extract directly when they don't have any special error handling requirements. This will make the common case of StartSpan from a carried context a one-liner, which I think is important (and I generally don't think purely-diagnostic APIs should require users to handle errors).

jmacd:
FYI this is very low-priority for us. It's fine with me to let this sit.

jmacd:
Arrr. I've screwed up by making this PR a request to merge into changes/add_tag_type, not master. I'm going to recreate this PR.

PR #93, opened 2016-07-08
Introduce CarriedChildOfOption and CarriedFollowsFromOption for starting spans with an extracted span ref

jmacd:


jmacd:
Comments lifted from the previous PR of this branch:

 @yurishkuro 

passing tracer to the option doesn't feel clean

 @jmacd

I'm curious what you think of the user-visible API change, since Apply is not user-visible. If you like the user-visible change, do you see another way to do this?

I considered passing the Tracer directly to the CarriedXXXOption, which also feels not clean, because that makes it possible for the Extract tracer to differ from the StartSpan tracer (and requires the user to pass the same thing twice).

I basically agree with you, but I think a little uncleanliness in an internal API is better than requiring users to call Extract directly when they don't have any special error handling requirements. This will make the common case of StartSpan from a carried context a one-liner, which I think is important (and I generally don't think purely-diagnostic APIs should require users to handle errors).

ext/options.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// TODO move RPCServerOption into this file? OR, remove it. Code such as:
+//
+// err := tracer.Extract(...)
+// if err != nil { ... }
+// tracer.StartSpan("...", RPCServerOption(...))
+//
+// can be rewritten as:
+// tracer.StartSpan("...", SpanKindRPCServer, ChildOfCarrierOption(...))
```
bensigelman: this comment is out-of-date naming wise, though I prefer `ChildOfCarrierOption` to `CarriedChildOfOption`... if you read it quickly, the latter sounds like the child is the thing in the carrier, not the parent (which is the case, of course).

ext/options.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	if err != nil {
+		opentracing.Tag{string(CarriedContextError), err.Error()}.Apply(tracer, o)
+	}
+}
+
+// CarriedSpanRefOption extracts a context from "carrier" (of type "format")
+// and returns a StartSpanOption that will the associated context as a span
+// reference of type "ref" to a newly started span.
+func CarriedSpanRefOption(ref opentracing.SpanReferenceType,
```
bensigelman: `s/CarriedSpanRefOption/ReferenceCarrierOption/`?

ext/options.go
```diff
@@ -0,0 +1,50 @@
...
+	format, carrier interface{}) opentracing.StartSpanOption {
+	return extractedOption{ref, format, carrier}
+}
+
+// CarriedChildOfOption constructs a opentracing.ChildOfRef SpanReference option
+// for use in StartSpan with a carried context.
+func CarriedChildOfOption(format, carrier interface{}) opentracing.StartSpanOption {
```
bensigelman: this would arguably flow better in calling code as

```
tracer.StartSpan("...", ext.ChildOfCarrier(format, carrier))
```

(i.e., get rid of the Option suffix... it's still obviously a StartSpanOption thanks to the return value if we're worried about greppability)

bensigelman:
This is what the older `Tracer.Join` should have been IMO. :)  A helper built on top of proper primitives.

I have some quibbles about naming and am a little perplexed about whether this should be in `ext/` (esp if it becomes `tags/`), but I like this in general.

ext/options.go
yurishkuro: "Carrier" in the name still reads redundant. Maybe `ExtractedChildOf`.

Also, the above again does not work for Zipkin's model without passing SpanKind tag separately, something that was done by another extension RPCServer

yurishkuro:
This does feel a bit premature to me. We just created a completely new API, I would like a few weeks to work it into the instrumentations before providing helpers with contentious naming & signatures. In all the instrumentation work I've done across languages shortening the code to inject/extract has never been at the top of the list, there are always various edge cases.

ext/options.go
jmacd: I like ExtractedChildOf.  I began with ExtractedChildOption, and Ben pointed out I needed "Of".

Re: the Zipkin model. Yes, I would recommend

  t.StartSpan("opName", ExtractedChildOf(format, carrier), TagSpanKindRPCServer)

(and I would remove RPCServerOption).

jmacd:
I'm not going to suggest that this is mature. Ben wants to wait until we've finished the new API in the other languagues, and I'm happy with that. I'm open to all naming suggestions / signatures and don't think this needs to be contentious 😄 

I went through our internal code base updating to the new SpanContext model and while there were only two places I needed to change Join into Extract, the existing code was confusingly dissimilar between the two code locations, because (a) one location used the gocontext helpers, one didn't, (b) they handled the errors from Join inconsistently. I could make the code look proper without these APIs, but it comes out much nicer this way. So, it's not about shortening the code, it's about hiding the error handling. If you are concerned with edge cases, of course you call Extract and handle the error, but my general feeling (as stated before) is that diagnostics APIs should not require handling of diagnostic errors.

yurishkuro:
Regarding the errors, I usually add a log when Extract() returns an error != SpanContextNotFound. In production these errors usually don't happen, so no performance impact, but when they do happen it indicates an incompatibility of the tracer implementation (we caught a couple of bugs like that). So I am not keen on providing methods in the API layer that encourage people to mask and ignore errors.

yurishkuro:
do we still need this PR?

bensigelman:
I'd say no... @jmacd yell if you disagree :)

jmacd:
I still like this PR.

PR #94, opened 2016-07-08
Update opentracing.Tag comments

jmacd:


jmacd:
@dominikh
See if you like these changes.

dominikh:
LGTM

ext/tags.go
```diff
@@ -15,39 +15,37 @@ import opentracing "github.com/opentracing/opentracing-go"
 //
 var (
 	//////////////////////////////////////////////////////////////////////
-	// SpanKind (client/server)
+	// SpanKind (client/server) tag names
```
bensigelman: same as that other PR -- `s/name/key/` for tags.

bensigelman:
(otherwise LGTM...)

ext/tags.go
```diff
@@ -83,22 +81,34 @@ var (
...
+	// Error Tag name
 	//////////////////////////////////////////////////////////////////////
 
 	// Error indicates that operation represented by the span resulted in an error.
 	Error = boolTagName("error")
 )
 
+var (
+	//////////////////////////////////////////////////////////////////////
+	// Conventional SpanKind Tags
```
yurishkuro: what does 'conventional' mean here?

ext/tags.go
```diff
@@ -83,22 +81,34 @@ var (
...
+	//////////////////////////////////////////////////////////////////////
+
+	// TagSpanKindRPCClient is a tag indicating the span is a RPC client.
+	TagSpanKindRPCClient = opentracing.Tag{Key: string(SpanKind), Value: SpanKindRPCClientTagValue}
+
+	// TagSpanKindRPCServer is a tag indicating the span is a RPC client.
+	TagSpanKindRPCServer = opentracing.Tag{Key: string(SpanKind), Value: SpanKindRPCServerTagValue}
```
yurishkuro: what do we lose if we just call it SpanKindRPCServer?  It matches with the other tags in this file, e.g.
```go
ext.SamplingPriority.Set(span, 1)
ext.SpanKindRPCClient.Set(span) // value implied
```

bensigelman:
@jmacd can you either update XOR close this PR? Thanks.

ext/tags.go
jmacd: To mean "conventional" means the same thing written above, "common [...] recommended for better portability [...]"

ext/tags.go
jmacd: OK

ext/tags.go
jmacd: I removed "Tag" as you suggested.

jmacd:
Hey all, sorry for the long delay. I'll submit this and have applied changes for all the recommendations above.

yurishkuro:
One way to keep this backwards compatible is to alias the old constant names to new constants.

jmacd:
I'm officially judging this not worth the trouble.

PR #96, opened 2016-07-09
Allow nil argument to ChildOf/FollowsFrom

yurishkuro:
Reason: to simplify span starting code by not forcing nil checks:

```go
sc, _ := tracer.Extract(someFormat, someCarrier)
span := tracer.StartSpan("operation", ChildOf(sc)) // should work even if sc == nil
```

cc @bensigelman

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -214,7 +214,7 @@ const (
 // it allows for a more concise syntax for starting spans:
 //
 //     sc, _ := tracer.Extract(someFormat, someCarrier)
-//     span := tracer.StartSpan("operation", ChildOf(sc))
+//     span := tracer.StartSpan("operation", opentracing.ChildOf(sc))
 //
 // The `ChildOf(sc)` option above will not panic if sc == nil, it will just
 // not add the parent span reference to the options.
@@ -231,8 +231,9 @@ func (r SpanReference) Apply(o *StartSpanOptions) {
 }
 
 // ChildOf returns a StartSpanOption pointing to a dependent parent span.
+// If sm == nil, the option has no effect.
 //
-// See ChildOfRef
+// See ChildOfRef, SpanReference
 func ChildOf(sm SpanContext) SpanReference {
 	return SpanReference{
 		Type:    ChildOfRef,
@@ -242,8 +243,9 @@ func ChildOf(sm SpanContext) SpanReference {
 
 // FollowsFrom returns a StartSpanOption pointing to a parent Span that caused
 // the child Span but does not directly depend on its result in any way.
+// If sm == nil, the option has no effect.
 //
-// See FollowsFromRef
+// See FollowsFromRef, SpanReference
 func FollowsFrom(sm SpanContext) SpanReference {
 	return SpanReference{
 		Type:    FollowsFromRef,
--- a/tracer.go
+++ b/tracer.go
@@ -209,7 +209,15 @@ const (
 
 // SpanReference is a StartSpanOption that pairs a SpanReferenceType and a
 // referee SpanContext ("referee" is "the one who is referred to"). See the
-// SpanReferenceType documentation.
+// SpanReferenceType documentation for supported relationships.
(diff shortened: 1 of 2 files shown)


dominikh:
Is it worth adding a comment stating that nil is an accepted value?

yurishkuro:
@dominikh good point, added.

tracer.go
```diff
@@ -209,15 +209,25 @@ const (
...
 
 // Apply satisfies the StartSpanOption interface.
 func (r SpanReference) Apply(o *StartSpanOptions) {
-	o.References = append(o.References, r)
+	if r.Referee != nil {
+		o.References = append(o.References, r)
+	}
 }
 
 // ChildOf returns a StartSpanOption pointing to a dependent parent span.
```
bensigelman: It may be redundant, but IMO we ought to repeat the comment about `nil` here and for `FollowsFrom()` as well

tracer.go
```diff
@@ -209,15 +209,25 @@ const (
 
 // SpanReference is a StartSpanOption that pairs a SpanReferenceType and a
 // referee SpanContext ("referee" is "the one who is referred to"). See the
-// SpanReferenceType documentation.
+// SpanReferenceType documentation for supported relationships.
+// If SpanReference is created with Referee == nil, it has no effect. Thus
+// it allows for a more concise syntax for starting spans:
+//
+//     sc, _ := tracer.Extract(someFormat, someCarrier)
+//     span := tracer.StartSpan("operation", ChildOf(sc))
```
bensigelman: nit: `opentracing.ChildOf(sc)`

bensigelman:
@yurishkuro I enthusiastically recognize that this PR is an improvement over what's in `master`, but AFAICT this is only really relevant when the `SpanContext` has been extracted from the wire (since `Span.Context()` will never return `nil`), and if we're talking about simplifying the `Extract` -> `StartSpan` code path it's perhaps better to do something like https://github.com/opentracing/opentracing-go/pull/93 instead (since it's even more concise).

yurishkuro:
@bensigelman afaict #93 is also _only_ about contexts extracted from the wire. And all it does is save just one line, the call to StartSpan, at the expense of quite awkward naming and limiting user's ability to handle errors. Re errors, I usually add a log when Extract() returns an error != SpanContextNotFound. In production these errors usually don't happen, so no performance impact, but when they do happen it indicates an incompatibility of the tracer implementation (we caught a couple of bugs like that). So I am not keen on providing methods in the API layer that encourage people to mask and ignore errors.

Re the other scenario of creating a child span from potential parent span, I wouldn't mind having a helper method to deal with potential SpanFromContext() returning nil, but it needs too many parameters (ctx, tracer, op name, other options, etc.) Instead, a composition approach would work better if we had a variant of ChildOf that could take a nullable Span:

```go
func ChildOfSpan(span Span) SpanReference {
    if span == nil {
        return ChildOf(nil)
    }
    return ChildOf(span.Context())
}
```

With it creating a plain child span is one liner
```go
span := tracer.StartSpan("operation", ChildOfSpan(SpanFromContext(ctx)))
```

tracer.go
yurishkuro: added

tracer.go
```diff
@@ -209,20 +209,31 @@ const (
...
+		o.References = append(o.References, r)
+	}
 }
 
 // ChildOf returns a StartSpanOption pointing to a dependent parent span.
+// If sm == nil, the option has no effect.
 //
-// See ChildOfRef
+// See ChildOfRef, SpanReference
 func ChildOf(sm SpanContext) SpanReference {
```
yurishkuro: @bensigelman  why do we can parameter `sm` here? `sc` would make more sense.

bensigelman: def should be `sc` -- that must have survived from the `SpanMetadata` days of the change.

bensigelman:
@yurishkuro thanks for the clarification re `context.Context` -- makes sense. LGTM.

PR #97, opened 2016-07-11
Support Trace ID and Sampled flag in MockTracer

yurishkuro:


mocktracer/mocktracer.go
```diff
@@ -212,10 +240,12 @@ func nextMockID() int {
...
 }
 
-func newMockSpanContext(spanID int) *MockSpanContext {
+func newMockSpanContext(traceID int, spanID int, sampled bool, baggage map[string]string) *MockSpanContext {
 	return &MockSpanContext{
+		TraceID: traceID,
 		SpanID:  spanID,
-		baggage: make(map[string]string),
+		Sampled: sampled,
+		baggage: baggage, // TODO make a copy
```
bensigelman: may as well do this now, I'd say

bensigelman:
:ship: after the map copy... thanks!

Issue #98, opened 2016-07-13
MockTracer GetFinishedSpans should be merely FinishedSpans


kriskowal:
Per the Go idiom of using bare names for accessors.

I’d suggest aliasing for compatibility.

bensigelman:
Would you be willing to send a PR? I agree that the Go-ish thing would be
`FinishedSpans`. (And we have just landed some much larger breaking
changes, so one more will be a drop in the bucket -- I think it's ok not to
alias)
ᐧ

kriskowal:
so it is written, so it shall be done. #99

bensigelman:
@kriskowal thanks :)

PR #99, opened 2016-07-13
Strip Get prefix from methods

kriskowal:
This change removes the Get prefix from mocktracer GetBaggage, GetTags, and GetFinishedSpans, per the Go idiom for accessor methods.

r @yurishkuro @bensigelman


kriskowal:
see #98

bensigelman:
LGTM.

@yurishkuro do you mind the breakage?

yurishkuro:
I'l live with it. :shipit:

PR #102, opened 2016-07-25
Add an HTTPHeaders format

bensigelman:
While in the neighborhood, improve some misc comments.

See https://github.com/opentracing/opentracing.github.io/issues/98

### DIFF
--- a/propagation.go
+++ b/propagation.go
@@ -45,32 +45,47 @@ var (
 type BuiltinFormat byte
 
 const (
-	// Binary encodes the SpanContext for propagation as opaque binary data.
+	// Binary represents SpanContexts as opaque binary data.
 	//
 	// For Tracer.Inject(): the carrier must be an `io.Writer`.
 	//
 	// For Tracer.Extract(): the carrier must be an `io.Reader`.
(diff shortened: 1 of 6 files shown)


propagation.go
```diff
@@ -45,32 +45,47 @@ var (
...
+	// For Tracer.Extract(): the carrier must be a `TextMapReader`.
+	//
+	// See HTTPHeaderCarrier for an implementation of both TextMapWriter
 	// and TextMapReader that defers to an http.Header instance for storage.
 	// For example, Inject():
 	//
-	//    carrier := HTTPHeaderTextMapCarrier(httpReq.Header)
-	//    err := span.Tracer().Inject(span, TextMap, carrier)
+	//    carrier := HTTPHeadersCarrier(httpReq.Header)
+	//    err := span.Tracer().Inject(span, HTTPHeaders, carrier)
```
yurishkuro: nit: probably should be using package name in the examples, `opentracing.HTTPHeaders`

yurishkuro:
LGTM

propagation.go
bensigelman: good point, done.

PR #108, opened 2016-08-29
Introduce key-value Span logging as an RFC

bensigelman:
This PR serves two purposes:
- to push the discussion about KV logging forward per https://github.com/opentracing/opentracing.github.io/issues/96
- to illustrate how KV logging might work in Go specifically (which is a little bit of a special case)

I haven't added all of the helpers that might be in here someday for various common cases, nor have I addressed the implications for implementations (MockTracer, BasicTracer, etc)... if we can come to a consensus about the basic direction I'll happily do so.

Thanks in advance for thoughts about any/all of this.


logfield.go
```diff
@@ -0,0 +1,136 @@
...
+	float64Type
+	errorType
+	objectType
+	deferredObjectType
+)
+
+// LogField instances are constructed via LogBool, LogString, and so on.
+// Tracing implementations may then handle them via the LogField.Process
+// method.
+type LogField struct {
```
bensigelman: "heavily influenced by" (i.e., partially stolen from) https://github.com/uber-go/zap

logfield.go
```diff
@@ -0,0 +1,136 @@
+package opentracing
```
yurishkuro: I was thinking we could put this into `opentracing/log` package, so that the fields could be created with `log.String()`, `log.Int64()`, etc. instead of much more verbose `opentracing.LogString()`

logfield.go
```diff
@@ -0,0 +1,136 @@
...
+	return LogField{
+		key:          key,
+		valType:      objectType,
+		interfaceVal: obj,
+	}
+}
+
+type DeferredObjectGenerator func() interface{}
+
+func LogDeferredObject(key string, generator DeferredObjectGenerator) LogField {
```
yurishkuro: name suggestion: simply `Deferred()` or `Lazy()`

span.go
```diff
@@ -49,22 +51,36 @@ type Span interface {
...
+	//        opentracing.LogInt("request_size", request.Size()))
 	//
-	// See LogData for semantic details.
-	Log(data LogData)
+	//    span.LogKV(
+	//        "request_path", request.Path(),
+	//        "request_size", request.Size())
+	//
+	// Also see Span.FinishWithOptions() and FinishOptions.BulkLogData.
+	LogFields(fields ...LogField)
```
yurishkuro: how about just `Log()`? It won't clash with the previous signature.

logfield.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+}
+
+// LogFieldProcessor allows access to the contents of a LogField (via a call to
+// LogField.Process).
+//
+// Tracer implementations typically provide an implementation of
+// LogFieldProcessor; OpenTracing callers should not need to concern themselves
+// with it.
+type LogFieldProcessor interface {
```
yurishkuro: A helper, I don't have a strong opinion about keeping or not keeping it, but if we keep it, I would suggest these naming changes, to reflect the pattern and make it more abstract:
* s/Processor/Visitor
* a/AddString/ForString (or OnString of VisitString)
* s/Process/RunVIsitor or ApplyVisitor

logfield.go
```diff
@@ -0,0 +1,136 @@
...
+	case int64Type:
+		processor.AddInt64(lf.key, lf.numericVal)
+	case float64Type:
+		processor.AddFloat64(lf.key, math.Float64frombits(uint64(lf.numericVal)))
+	case errorType:
+		processor.AddString(lf.key, lf.obj.(error).Error())
+	case objectType:
+		processor.AddObject(lf.key, lf.interfaceVal)
+	case deferredObjectType:
+		processor.AddObject(lf.key, lf.interfaceVal.(DeferredObjectGenerator)())
```
yurishkuro: why force a dereferencing here? A more general visitor behavior would be just to pass the value in, and the visitor can dereference if as needed

span.go
```diff
@@ -49,22 +51,36 @@ type Span interface {
 	// may ignore the tag, but shall not panic.
 	SetTag(key string, value interface{}) Span
 
-	// LogEvent() is equivalent to
+	// LogFields and LogKV are two ways to record logging data about a Span.
+	// Both allow for timestamped key-value logging of arbitrary data. Neither
+	// intrinsically supports message formatting; in fact, formatted log
+	// messages are discouraged (though not disallowed) in OpenTracing.
```
yurishkuro: we'd probably need a reference to somewhere, e.g. a blog post

span.go
```diff
@@ -49,22 +51,36 @@ type Span interface {
...
 	//
-	// See LogData for semantic details.
-	Log(data LogData)
+	//    span.LogKV(
+	//        "request_path", request.Path(),
+	//        "request_size", request.Size())
+	//
+	// Also see Span.FinishWithOptions() and FinishOptions.BulkLogData.
+	LogFields(fields ...LogField)
+	// For LogKV (as opposed to LogFields()), every even parameter must be a
```
yurishkuro: it seems even/odd terminology used implies 0-based indexing, which might be a bit confusing. Maybe rephrase without even/odd but as a "sequence of key, value, key, value, ... where each key is a string ..."?

span.go
```diff
@@ -49,22 +51,36 @@ type Span interface {
...
-	LogEventWithPayload(event string, payload interface{})
-
-	// Log() records `data` to this Span.
+	//    span.LogFields(
+	//        opentracing.LogString("request_path", request.Path()),
+	//        opentracing.LogInt("request_size", request.Size()))
 	//
-	// See LogData for semantic details.
-	Log(data LogData)
+	//    span.LogKV(
```
yurishkuro: I think this might be better places in the commends for LogKV method

span.go
```diff
@@ -49,22 +51,36 @@ type Span interface {
...
 	//
-	//   Log(LogData{Event: event, Payload: payload0})
+	// For example, the following are equivalent:
 	//
-	LogEventWithPayload(event string, payload interface{})
-
-	// Log() records `data` to this Span.
+	//    span.LogFields(
+	//        opentracing.LogString("request_path", request.Path()),
+	//        opentracing.LogInt("request_size", request.Size()))
```
yurishkuro: if we move this to `opentracing/log` package, we can also define a number of _standard_ fields, then the example might look more appealing, e.g. `span.Log(logs.Error("oh my"))`

yurishkuro:
Overall +1, all comments are stylistic.

logfield.go
yurishkuro: they might appreciate this comment being in the code :-)

logfield.go
bensigelman: yeah, I started down that road and moved it here because I thought I would eventually want `LogReference` and/or other things defined in the main OT package. Meh.

logfield.go
bensigelman: Sure, I've added it.

logfield.go
bensigelman: I was struggling with this a bit... why not allow for deferred things other than objects/interface{}? That's what led to the Object suffix. As for Deferred vs Lazy, no strong feelings here.

logfield.go
bensigelman: Actually not really "helper" at the moment, as there is no way to get (otherwise unexported) state out of LogFields without this.

I don't like the naming at the moment either. Not sure if I like Visitor much more, but if we agree on the basic shape of the API I'm happy to take a pass at better names throughout.

logfield.go
bensigelman: You mean for the generator specifically, right? No great reason.

span.go
bensigelman: yes, this was aspirational...

span.go
bensigelman: Well, it will *break* the previous signature, though at least it would do so obviously.

I am starting to feel as though the Go API has enough use that it's pretty shitty to break people's code without some grace period where the new and old are both supported.

bensigelman: Oh: unless you meant to make `LogField` an interface and have the old `LogData` implement it somehow?

yurishkuro: didn't mean it, but sounds interesting

logfield.go
yurishkuro: > as there is no way to get (otherwise unexported) state out of LogFields without this.

ah, I see. Not sure if we really need LogField's fields to be private as they were in zap. Private fields do prevent the possibility of creating an invalid instance where valType doesn't match the value.

logfield.go
yurishkuro: btw, the Marshaler type is interesting: https://github.com/uber-go/zap/blob/master/field.go#L164

yurishkuro: > why not allow for deferred things other than objects/interface{}

Actually, why shouldn't `deferred()` return a `LogField` as well? That seems more inline with the overall intention of type safety of this API, whereas if it returns an `interface{}` we're back into reflection based introspection.

logfield.go
bensigelman: I like the "enforced sanity" of the private fields, even if it meddles with extensibility... I am playing with a version of something like LogMarshaler that should help for exceptional situations.

bensigelman:
@yurishkuro PTAL... I'll add a few notes inline.

log/field.go
```diff
@@ -0,0 +1,204 @@
+package log
```
bensigelman: I think the `LazyLogger` will be sufficient for things like `Reference` logging (which we may not even special-case), so I moved this back into its own package.

log/field.go
```diff
@@ -0,0 +1,204 @@
...
+import "math"
+
+type fieldType int
+
+const (
+	stringType fieldType = iota
+	boolType
+	intType
+	int32Type
+	uint32Type
```
bensigelman: (added more numeric types)

log/field.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+}
+
+// FieldVisitor allows access to the contents of a Field (via a call to
+// Field.Visit).
+//
+// Tracer implementations typically provide an implementation of
+// FieldVisitor; OpenTracing callers should not need to concern themselves
+// with it.
+type FieldVisitor interface {
```
bensigelman: went with the visitor naming...

mocktracer/mocklogrecord.go
```diff
@@ -0,0 +1,105 @@
...
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/opentracing/opentracing-go/log"
+)
+
+// MockLogRecord represents data logged to a Span via Span.LogFields or
+// Span.LogKV.
+type MockLogRecord struct {
```
bensigelman: The `opentracing.LogRecord` is not very useful for tests, so I went with this... Note that `MockKeyValue` coerces all values to strings but does remember the `reflect.Kind`.

mocktracer/mockspan.go
```diff
@@ -183,8 +185,26 @@ func (s *MockSpan) Finish() {
...
 	s.Lock()
 	s.FinishTime = opts.FinishTime
-	s.logs = append(s.logs, opts.BulkLogData...)
 	s.Unlock()
+
+	// Handle any late-bound LogRecords.
+	for _, lr := range opts.LogRecords {
+		s.logFieldsWithTimestamp(lr.Timestamp, lr.Fields...)
+	}
+	// Handle (deprecated) BulkLogData.
```
bensigelman: this code gives a sense of what Tracer implementations might do to support the old API during a migration period

mocktracer/mockspan.go
```diff
@@ -196,26 +216,66 @@ func (s *MockSpan) String() string {
...
+	for i := 0; i*2 < len(keyValues); i++ {
+		key, ok := keyValues[i*2].(string)
+		if !ok {
+			s.LogFields(log.Error(
+				fmt.Errorf(
+					"Non-string key (pair=%v): %v",
+					i, reflect.TypeOf(keyValues[i*2]))))
+			return
+		}
+		valStr := fmt.Sprint(keyValues[i*2+1])
```
bensigelman: this `Sprint` is lame, but it's just the MockSpan (so who cares??).

bensigelman: (see https://github.com/opentracing/opentracing-go/pull/108/files#diff-c37ff0b131ce1be6dcd5b5ddc89c7f30R242)

span.go
```diff
@@ -90,39 +105,20 @@ type Span interface {
 
 	// Provides access to the Tracer that created this Span.
 	Tracer() Tracer
+
+	// DEPRECATED
+	LogEvent(event string)
```
bensigelman: note that I've added these back, so while all Tracer impls will break, existing calling code should continue to function for the time being as long as it doesn't use MockTracer.

log/field.go
```diff
@@ -0,0 +1,204 @@
...
+func Error(err error) Field {
+	return Field{
+		key:          "error",
+		fieldType:    errorType,
+		interfaceVal: err,
+	}
+}
+
+// Object adds an object-valued key:value pair to a Span.LogFields() record
+func Object(key string, obj interface{}) Field {
```
yurishkuro: can you clarify the intention/semantics? What is tracer expected to do (or NOT to do, like convert to String) with it?  Maybe a warning to the users that this is likely more expensive to log than primitive values.

log/field.go
yurishkuro: AddXyz methods look odd in a visitor... We could actually drop the prefix altogether.

log/field.go
```diff
@@ -0,0 +1,204 @@
...
+	return Field{
+		key:          key,
+		fieldType:    objectType,
+		interfaceVal: obj,
+	}
+}
+
+// LazyLogger allows for user-defined, late-bound logging of arbitrary data
+type LazyLogger func(fv FieldVisitor)
```
yurishkuro: `fv FieldVisitor` - neat!

mocktracer/mockspan.go
```diff
@@ -196,26 +216,66 @@ func (s *MockSpan) String() string {
...
+	s.Lock()
+	defer s.Unlock()
+	s.logs = append(s.logs, lr)
+}
+
+// LogKV belongs to the Span interface.
+//
+// This implementations coerces all "values" to strings, though that is not
+// something all implementations need to do. Indeed, a motivated person can and
+// probably should have this do a typed switch on the values.
```
yurishkuro: sounds like a candidate for helper function (to add later)

mocktracer/mocktracer_test.go
```diff
@@ -88,22 +91,115 @@ func TestMockTracer_FinishedSpans_and_Reset(t *testing.T) {
...
+		},
+	}, actual)
+}
+
+func TestMockSpan_LogFields(t *testing.T) {
+	tracer := New()
+	span := tracer.StartSpan("s")
+	span.LogFields(log.String("key0", "string0"))
+	span.LogFields(log.String("key1", "string1"), log.Uint32("key2", uint32(42)))
+	span.LogFields(log.Lazy("SHOULD_BE_IGNORED", func(fv log.FieldVisitor) {
```
yurishkuro: maybe `log.Lazy()` should not accept the key, only the lambda?

log/field.go
bensigelman: This is called `Marshaller` in zap, and I kinda understand why... think of it from the perspective of someone implementing a `LazyLogger` function. They're basically picking a type and setting a key-value pair.

We could drop the prefix altogether, but I don't love doing that sort of thing in Go since interfaces are implemented just by name overlaps and these base type names feel like they could be ripe for collisions.

yurishkuro:
overall, looks good.

log/field.go
```diff
@@ -0,0 +1,204 @@
...
+
+// Lazy adds a LazyLogger to a Span.LogFields() record; the tracing
+// implementation will call the LazyLogger function at an indefinite time in
+// the future (after Lazy() returns).
+//
+// Note that `ignoredKey` is ignored (as the LazyLogger can control the key).
```
jmacd: Why have this argument, then?

log/field.go
bensigelman: I honestly don't know what the semantics should be at the OT layer. We could at least own up to that... I can imagine impls that choose to ignore generic objects entirely; others that truncate or summarize them; and others that record everything.

log/field.go
bensigelman: @jmacd yes, just removed it in 797a392 :)

log/field.go
yurishkuro: I'm just thinking that people seeing this for the first time will be like "add what to where?" We can keep a neutral prefix like `Handle` or `On`

log/field.go
```diff
@@ -0,0 +1,202 @@
...
+	AddBool(key string, value bool)
+	AddInt(key string, value int)
+	AddInt32(key string, value int32)
+	AddInt64(key string, value int64)
+	AddUint32(key string, value uint32)
+	AddUint64(key string, value uint64)
+	AddFloat32(key string, value float32)
+	AddFloat64(key string, value float64)
+	AddObject(key string, value interface{})
+	AddLazyLogger(key string, value LazyLogger)
```
yurishkuro: should we drop the `key` here as well?

bensigelman:
I made some naming changes in 64b1971 per some offline suggestions from @jmacd... I think this is clearer when you consider all of the callers.

log/field.go
```diff
@@ -0,0 +1,201 @@
...
+	EmitBool(key string, value bool)
+	EmitInt(key string, value int)
+	EmitInt32(key string, value int32)
+	EmitInt64(key string, value int64)
+	EmitUint32(key string, value uint32)
+	EmitUint64(key string, value uint64)
+	EmitFloat32(key string, value float32)
+	EmitFloat64(key string, value float64)
+	EmitObject(key string, value interface{})
+	EmitLazyLogger(key string, value LazyLogger)
```
yurishkuro: does `key` make sense here?

log/field.go
```diff
@@ -0,0 +1,201 @@
...
+	intType
+	int32Type
+	uint32Type
+	int64Type
+	uint64Type
+	float32Type
+	float64Type
+	errorType
+	objectType
+	lazyLoggerType
```
yurishkuro: one other interesting feature in zap was nested fields, which allow logging hierarchical structured data.

jmacd: I was wondering if we could remove key from the Encoder interface, but I
fear that doing so ultimately requires using an interface type as the Field
type, which will hurt performance.

When you ask about nested fields, I start to wonder exactly how you'll use
it.

Another question: I don't see slices or array types in the list of
supported types. Omission or intentional?

ᐧ

yurishkuro: nested field example:
```go
span.LogFields(
    log.Nested("employee",
        log.String("firstName", ...),
        log.String("lastName", ...),
    ),
)
```

yurishkuro: As for arrays, I would say the main reason not to include them is that it cannot be done while preserving type safety without doubling the number of types already present. Array of `interface{}` is fairly meaningless (might as well use `log.Object()`), while `log.Uint32Array()`, `log.StringArray()`, etc. is (to me) unpalatable.

jmacd: You could also write something like:

LogFields("employee", struct{First, Last string}{first, last})

I experimented with nested logging--I ran into trouble with the semantics. Are labels tags? Can you apply more than one tag to a value? What's the resulting structure look like? In the end I did come up with a satisfying approach, but it was significantly more complicated than anything discussed in this thread, possibly beyond the point of actually being useful. This is why I ask--what the application is for nested logging.

Do you want to find all logs that have a "firstName" property? Is there any sort of namespace to prevent matching logs from unrelated code with "firstName" properties? Can you join logs by these keys?

bensigelman: Personally, I see a strong enough use case for logging structs (as `interface{}` values, via something like the `log.Object(...)` in this PR), but pretty much as just a big (possibly truncated) blob for debugging purposes. Nested key-value fields at this logging layer are not indefensible, but IMO cross the complexity threshold and aren't worth the corner cases they expose.

And there's the perennial "easier to add than to remove" argument, too...

bensigelman:
@yurishkuro @tschottdorf PTAL... I added some stuff that made the basictracer-go update easier (and would help other implementors as well).

log/testutil.go
```diff
@@ -0,0 +1,122 @@
...
+// Usage:
+//
+//     fv := log.NewFieldValidator(t, someLogStructure.Fields)
+//     fv.
+//         ExpectNextFieldEquals("key1", reflect.String, "some string value").
+//         ExpectNextFieldEquals("key2", reflect.Uint8, "255")
+//
+// FieldValidator satisfies the log.Encoder interface and thus is able to
+// marshal log.Field instances (which it takes advantage of internally).
+type FieldValidator struct {
```
bensigelman: See https://github.com/opentracing/basictracer-go/pull/37/files#diff-94fbb48b23e68ffcc13ee7a6aa73773aR99 for an example of how this can be used.

log/testutil.go
```diff
@@ -0,0 +1,122 @@
...
+//
+// FieldValidator SHOULD ONLY BE USED FOR TESTS. It is not intended for
+// production code.
+//
+// Usage:
+//
+//     fv := log.NewFieldValidator(t, someLogStructure.Fields)
+//     fv.
+//         ExpectNextFieldEquals("key1", reflect.String, "some string value").
+//         ExpectNextFieldEquals("key2", reflect.Uint32, "4294967295")
```
bensigelman: I like this approach more than `ExpectNextFieldEquals("key2", uint32(4294967295))` because `reflect.Interface` is an important case to test with object logging.

log/field.go
yurishkuro: still think there should be no key argument in `EmitLazyLogger` method

log/field.go
```diff
@@ -0,0 +1,201 @@
...
+
+// LazyLogger allows for user-defined, late-bound logging of arbitrary data
+type LazyLogger func(fv Encoder)
+
+// Lazy adds a LazyLogger to a Span.LogFields() record; the tracing
+// implementation will call the LazyLogger function at an indefinite time in
+// the future (after Lazy() returns).
+func Lazy(ll LazyLogger) Field {
+	return Field{
+		key:          "Lazy", // will be overwritten
```
yurishkuro: I'd leave off assigning anything

span.go
```diff
@@ -135,15 +133,38 @@ type FinishOptions struct {
...
+
+// LogData is DEPRECATED
+type LogData struct {
+	Timestamp time.Time
+	Event     string
+	Payload   interface{}
+}
+
+// ToLogRecord converts a deprecated LogData to a non-deprecated LogRecord
+func (ld *LogData) ToLogRecord() LogRecord {
```
yurishkuro: looks like `'d` shouldn't be a pointer, because nowhere else are we using pointers to LogData.

log/testutil.go
yurishkuro: not sure I see a strong need for this in the API package. E.g. you didn't need to use it in mock tracer tests. Whatever the internal format of field storage a tracer implementation is using, they should be able to do assertions directly against that struct, like https://github.com/opentracing/opentracing-go/pull/108/files#diff-6d19a913e5779231470a18e52e75389fR180

yurishkuro: for instance, [this is the only example](https://github.com/opentracing/basictracer-go/pull/37/files#diff-712c45fed6eb613b9582291e641be233R40) of actually using Field values in basictracer. A real recorder would have to call the visitor at this point and convert the fields into some other structure where it has access to private fields, either its own struct or serialization into some thrift/protobuf, from where the tests can read the values.

log/field.go
bensigelman: Sorry, missed this comment -- removed the param, you're right.

log/testutil.go
bensigelman: I do want to use this is basictracer. I could move it over there (actually, that's where it began life), whatever.

bensigelman:
(changes pushed)

log/testutil.go
bensigelman: (moved)

span.go
yurishkuro: Not convinced without benchmark, but ok, doesn't matter. 

I once benchmarked passing `time.Time` (two uint64s) by value and by ptr, and by-value ns/op was about 60% of the by-pointer ns/op. But here the struct is 2x the size of time.Time, so meh.

yurishkuro:
LGTM

How are we doing with versioning it, 0.11? yarpc guys were asking for a 1.0 milestone plan.

span.go
bensigelman: @yurishkuro I did do an informal benchmark before writing my comment above (just to sanity-check things). In this case it's a wash. Just FYI, here are some more formal numbers comparing a thin (but not empty) struct and a thick (1K) struct...

Here's `bench_test.go`:

```go
package value_ptr

import "testing"

/////////////////////////////////////////
// THIN STRUCT (8 bytes)
/////////////////////////////////////////
type Thin struct {
       	idx int
       	data [4]byte
}

func (t Thin) ByValue() {
       	t.data[t.idx]++
       	t.idx++
       	if t.idx > len(t.data) { t.idx = 0 }
}

func (t *Thin) ByPointer() {
       	t.data[t.idx]++
       	t.idx++
       	if t.idx >= len(t.data) { t.idx = 0 }
}

func BenchmarkThinByValue(b *testing.B) {
       	t := Thin{idx: 0}
       	for i := 0; i < b.N; i++ {
       		t.ByValue()
       	}
}

func BenchmarkThinByPointer(b *testing.B) {
       	t := Thin{idx: 0}
       	for i := 0; i < b.N; i++ {
       		t.ByPointer()
       	}
}

/////////////////////////////////////////
// THICK STRUCT (1K)
/////////////////////////////////////////
type Thick struct {
       	idx int
       	data [1020]byte
}

func (t Thick) ByValue() {
       	t.data[t.idx]++
       	t.idx++
       	if t.idx > len(t.data) { t.idx = 0 }
}

func (t *Thick) ByPointer() {
       	t.data[t.idx]++
       	t.idx++
       	if t.idx >= len(t.data) { t.idx = 0 }
}

func BenchmarkThickByValue(b *testing.B) {
       	t := Thick{idx: 0}
       	for i := 0; i < b.N; i++ {
       		t.ByValue()
       	}
}

func BenchmarkThickByPointer(b *testing.B) {
       	t := Thick{idx: 0}
       	for i := 0; i < b.N; i++ {
       		t.ByPointer()
       	}
}
bhs@bhs-ls:~/tmp/go/src/value_ptr$
```

... and here are the results:

```
BenchmarkThinByValue-8         	1000000000     	         2.64 ns/op
BenchmarkThinByPointer-8       	500000000      	         3.12 ns/op
BenchmarkThickByValue-8        	50000000       	        30.0 ns/op
BenchmarkThickByPointer-8      	500000000      	         3.40 ns/op
```

bensigelman:
@yurishkuro 

> How are we doing with versioning it, 0.11? yarpc guys were asking for a 1.0 milestone plan.

I think the ideal 1.0 milestone would be the thing that removes the deprecated logging methods. The only remaining item for the OT-wide "1.0" is this KV logging thing: https://github.com/opentracing/opentracing.github.io/milestone/2

bensigelman:
(so yes, `0.11.0` after this is merged)

tschottdorf:
Haven't reviewed in detail, but generally +1 for a kv-style logging API. cc @raduberinde who's most active in our tracer usage right now.

log/field.go
RaduBerinde: This opens up an interesting case - the object could have a `String()` method that returns a *dynamic* value. It should be clarified if that is disallowed or not (and what the implications are).  If we require any implementation to only convert to string once, then there is no issue (but that requirement should be documented).

log/util.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	fields := make([]Field, len(keyValues)/2)
+	for i := 0; i*2 < len(keyValues); i++ {
+		key, ok := keyValues[i*2].(string)
+		if !ok {
+			return nil, fmt.Errorf(
+				"Non-string key (pair=%v): %v",
+				i, reflect.TypeOf(keyValues[i*2]))
+		}
+		switch typedVal := keyValues[i*2+1].(type) {
```
RaduBerinde: What is the value of this `switch` vs. just using `Object()` to store the value as an `interface{}`? As far as I understand, the field stuff is useful to avoid the allocations involved in boxing things into `interface{}`, but that has already happened here.

log/field.go
bensigelman: I think this case must be left up to the implementations for now. There are many viable but mutually-incompatible ways to interpret an `interface{}` value in the context of logging and I'd feel better having specific OT guidance once the pros and cons of the possible interpretations are clearer (in production).

log/util.go
bensigelman: The approach taken here allows `Span` implementations to handle the "simple" (POD, string) cases via the `EmitInt`, `EmitString`, etc functions of the `Encoder` interface. If we just called everything an `Object`/`interface{}` that wouldn't work.

bensigelman:
@basvanbeek regarding https://github.com/openzipkin/zipkin-go-opentracing: I don't want to break your stuff, but this is going to add two new methods to `Span`. One is trivial to implement thanks to a helper in `opentracing-go`, and the other one is probably (?) straightforward but will require some actual effort.

The basictracer changes are in PR here: https://github.com/opentracing/basictracer-go/pull/37/files#diff-2cd9647056f99d73ee03273cbaaa4aa2R92

Do you want help making the change on the zipkin-go-opentracing side?

This will not break any callers (it deprecates rather than removes the old logging methods).

log/util.go
RaduBerinde: Ah, right, that makes sense, I missed the fact that the `Encoder` is part of a specific implementation.

span.go
```diff
@@ -135,15 +133,38 @@ type FinishOptions struct {
...
+	return LogRecord{
+		Timestamp: ld.Timestamp,
+		Fields: []log.Field{
+			log.String("event", ld.Event),
+			log.Object("payload", ld.Payload),
```
yurishkuro: @bensigelman this broke one of my test hat was calling LogEvent, so the payload here was nil. I think we should not be adding the `payload` field in that case. I, of course, could handle it in the `Encoder` implementation, but in theory `log.Object("someKey", nil)` might be a meaningful combination for the end user and the span should preserve it.

yurishkuro: Suggestion:
```go
func (ld *LogData) ToLogRecord() LogRecord {
	lr := LogRecord{
		Timestamp: ld.Timestamp,
	}
	if ld.Payload == nil {
		lr.Fields = []log.Field{
			log.String("event", ld.Event),
		}
	} else {
		lr.Fields = []log.Field{
			log.String("event", ld.Event),
			log.Object("payload", ld.Payload),
		}
	}
	return lr
}
```

RaduBerinde: +1
I agree that `log.Object("someKey", nil)` can be a meaningful combination. We have been using a somewhat similar key-value interface for log messages and log events.; we can tag messages with key-value information like `node`=`1` but some things don't involve a value (e.g. `aborted`). We use a nil object value for those.

bensigelman: @RaduBerinde this particular code location is just handling the deprecated code path... i.e., `log.Object("someKey", nil)` will still work even if I fix the thing Yuri asked for (which I'm going to do... seems closer to the behavior the user would expect than what I have here).

README.md
```diff
@@ -50,7 +50,9 @@ happily rely on it for `Span` propagation. To start a new (blocking child)
         ...
         span, ctx := opentracing.StartSpanFromContext(ctx, "operation_name")
         defer span.Finish()
-        span.LogEvent("xyz_called")
+        span.LogFields(
+            log.String("event", "xyz called"),
+            log.Int("request.size", request.Size()))
```
dominikh: Minor nitpick, but from a user's POV, this specific example (`request.size`) is confusing because it looks more like a tag, not something that should have a timestamp.

log/field.go
```diff
@@ -0,0 +1,200 @@
...
+		fieldType:  uint32Type,
+		numericVal: int64(val),
+	}
+}
+
+// Uint64 adds a uint64-valued key:value pair to a Span.LogFields() record
+func Uint64(key string, val uint64) Field {
+	return Field{
+		key:        key,
+		fieldType:  int64Type,
```
dominikh: This should be `uint64Type`

log/field.go
```diff
@@ -0,0 +1,200 @@
...
+// https://github.com/uber-go/zap
+type Field struct {
+	key          string
+	fieldType    fieldType
+	numericVal   int64
+	stringVal    string
+	interfaceVal interface{}
+}
+
+// String adds a string-valued key:value pair to a Span.LogFields() record
```
dominikh: Idiomatic Go documentation drops the parentheses, as we're talking about the identifier, not the function call. There are a bunch of others in this PR, like `Lazy()`.

log/util.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"fmt"
+	"reflect"
+)
+
+// InterleavedKVToFields converts keyValues a la Span.LogKV() to a Field slice
+// a la Span.LogFields().
+func InterleavedKVToFields(keyValues ...interface{}) ([]Field, error) {
+	if len(keyValues)%2 != 0 {
+		return nil, fmt.Errorf("Non-even keyValues len: %v", len(keyValues))
```
dominikh: Idiomatic errors start with a lowercase letter. Also, there's a word for this: uneven.

dominikh: Use `%d` instead of `%v`.

log/util.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	fields := make([]Field, len(keyValues)/2)
+	for i := 0; i*2 < len(keyValues); i++ {
+		key, ok := keyValues[i*2].(string)
+		if !ok {
+			return nil, fmt.Errorf(
+				"Non-string key (pair=%v): %v",
```
dominikh: Unidiomatic error as per above, use `%d` as per above.

log/util.go
```diff
@@ -0,0 +1,57 @@
...
+				i, reflect.TypeOf(keyValues[i*2]))
```
dominikh: You can just use the `%T` printf verb to print the type, instead of using `reflect.TypeOf`.

span.go
```diff
@@ -49,22 +53,35 @@ type Span interface {
...
 	SetTag(key string, value interface{}) Span
 
-	// LogEvent() is equivalent to
+	// LogFields is an efficient and type-checked way to record key:value
+	// logging data about a Span, though the programming interface is a little
+	// more verbose than LogKV(). Here's an example:
 	//
-	//   Log(LogData{Event: event})
+	//    span.LogFields(
+	//        log.String("request_path", request.Path()),
```
dominikh: Same concern about tags vs logging as per above. Is structured logging actually supposed to replace tags? Because I didn't see that in the proposal.

span.go
```diff
@@ -90,39 +107,20 @@ type Span interface {
 
 	// Provides access to the Tracer that created this Span.
 	Tracer() Tracer
+
+	// DEPRECATED
```
dominikh: godoc actually has an idiomatic way of declaring an identifier as deprecated:

> To signal that an identifier should not be used, add a paragraph to its doc comment that begins with "Deprecated:" followed by some information about the deprecation.

The note being in a separate paragraph also implies that the deprecated identifiers should still carry meaningful (most likely their old, pre-deprecation) documentation.

basvanbeek:
@bensigelman: I'll inspect the change to opentracing-go and PR for basictracer-go and update https://github.com/openzipkin/zipkin-go-opentracing with the needed changes.

span.go
RaduBerinde: Yes, I was agreeing with Yuri. We were both saying that `log.Object("payload", nil)` might not be the no-op that the existing code assumes it is.

README.md
bensigelman: The idea here is that `"xyz called"` happened at some specific moment, and the call had a certain `"request.size"`... whether that "call" should be its own Span is a tradeoff, but I do think there's an implied timestamp for this event.

log/field.go
bensigelman: I think it's helpful as a hint that `LogFields` is, in fact, a `func`. I don't feel strongly but that's my rationale. I wish godoc were more explicit (via something like `@see` in Javadoc) which would allow tools to format/decorate things appropriately.

log/field.go
bensigelman: great catch!

bensigelman: (PS: if I had more time I would write tests for all of these... I did just manually re-check all of the `fieldType`s though.)

log/util.go
bensigelman: "uneven" doesn't refer to even-odd-ness... more like "unlevelness".

image

bensigelman: meh, ok. changed.

log/util.go
bensigelman: good idea...

span.go
bensigelman: It's not supposed to replace tags, no... maybe the "request" example isn't working for you. I see this come up when people have little events during the lifetime of a Span that don't warrant their own child Spans and don't contribute to the critical path but are worth remembering.

I can choose a different sort of example (e.g., recording information about a soft error) if you'd prefer...

span.go
bensigelman: Given that the API we're deprecating was pre-1.0 and has few users, I'd frankly rather just bury it entirely. I want to remove it ASAP but OT has enough users at this point that I can't just delete it. So I will scour the earth for OSS code that references the old API, submit PRs to adopt the new API, then delete all of this.

Inasmuch as removing the docs will help people not-use-it, I would rather remove the docs.

span.go
bensigelman: (I changed the example to a soft error case... open to suggestions here.)

log/util.go
```diff
@@ -0,0 +1,54 @@
+package log
+
+import "fmt"
+
+// InterleavedKVToFields converts keyValues a la Span.LogKV() to a Field slice
+// a la Span.LogFields().
+func InterleavedKVToFields(keyValues ...interface{}) ([]Field, error) {
+	if len(keyValues)%2 != 0 {
+		return nil, fmt.Errorf("non-even keyValues len: %d", len(keyValues))
```
dominikh: I don't know what got into me when I suggested "uneven". *Obviously* I meant to say "odd". :blush:

bensigelman: haha :)

While I see what you are getting at (and while I recognize that this is micro-nit territory for both of us), I do like "non-even" because it is unmistakable. "Odd" has other (primary) connotations in English and would be easier to misinterpret.

span.go
dominikh: SGTM. Just adding a reference to the not deprecated alternative should be good enough then.

span.go
dominikh: Liking the new example a lot more, thanks.

log/field.go
dominikh: I'll give you a free pass because it's wrong in all of the ot-go documentation already, but do change your ways one day ;)

bensigelman:
I want to merge this but don't want to break Zipkin... I sent a PR to help on that front and will wait for a consensus there before finishing this off.

In the meantime, if there are any further concerns about this PR, please raise them now! It'll probably go in within 48h.

log/field.go
yurishkuro: why not have reference field type explicitly? lazyLogger ultimately still ends up with primitive type of `interface{}`, but Reference is something we should support as a first class citizen imo.

bensigelman: @yurishkuro while I agree we should support Reference as a first-class citizen, I was hoping to add more appropriate (for logging) Reference types before exposing that feature.

yurishkuro: one problem with this approach is that adding a new field type breaks the visitor API by requiring an extra method to be implemented.

bensigelman: It's true... I dunno, seems like there are reasonable arguments in either direction. I would have liked to add a comment about supporting `Reference` types in the N-1 other languages (which have been submitted already). How strongly do you feel about this? (I'm on the fence)

yurishkuro: Don't feel very strongly. If we don't do it now, it will require a major version bump to add references, assuming we release the current state as 1.0

bensigelman: Funny... I've been thinking of semver bumps in terms of clients of the API (rather than implementations), but I see what you mean. We'd have to resign the symbolic importance of major version numbers :-/

bensigelman:
@basvanbeek has done https://github.com/openzipkin/zipkin-go-opentracing/pull/26 , so I'm going to merge this.

For anyone that's broken by this: tell Go team to endorse some particular semver-aware packaging system, please :-/

yurishkuro:
@bensigelman 👏  v1.0.0?

bensigelman:
@yurishkuro sure... btw, I thought of a somewhat horrible thing we could do to avoid a major version bump. Right now, we have

```
type Encoder interface {
	EmitString(key, value string)
	EmitBool(key string, value bool)
	EmitInt(key string, value int)
	EmitInt32(key string, value int32)
	EmitInt64(key string, value int64)
	EmitUint32(key string, value uint32)
	EmitUint64(key string, value uint64)
	EmitFloat32(key string, value float32)
	EmitFloat64(key string, value float64)
	EmitObject(key string, value interface{})
	EmitLazyLogger(value LazyLogger)
}
```

We could do something like

```
type ExtendedEncoder interface {
        Encoder
	EmitSpanReference(ref SpanReference)
}
```

... then do a checked type assertion in `Field.Marshal` to see if we have an Encoder or an ExtendedEncoder. If it's a plain encoder, we could treat the SpanReference as an Object.

I also realized that the SpanReference would be a package import cycle. :-/

bensigelman:
@yurishkuro https://github.com/opentracing/opentracing-go/releases/tag/v1.0.0

PR #113, opened 2016-09-06
Remove URL escaping from HTTPHeadersCarrier and add a test for it

yurishkuro:
Support HTTPHeaders format in MockTracer with URL escaping

Addresses https://github.com/opentracing/opentracing.github.io/issues/113

### DIFF
--- a/mocktracer/mocktracer_test.go
+++ b/mocktracer/mocktracer_test.go
@@ -1,6 +1,7 @@
 package mocktracer
 
 import (
+	"net/http"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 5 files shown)


mocktracer/propagation.go
```diff
@@ -52,7 +55,11 @@ func (t *TextMapPropagator) Inject(spanContext MockSpanContext, carrier interfac
 	writer.Set(mockTextMapIdsPrefix+"sampled", fmt.Sprint(spanContext.Sampled))
 	// Baggage:
 	for baggageKey, baggageVal := range spanContext.Baggage {
-		writer.Set(mockTextMapBaggagePrefix+baggageKey, baggageVal)
+		safeVal := baggageVal
+		if t.HTTPHeaders {
+			safeVal = url.QueryEscape(baggageVal)
+		}
+		writer.Set(mockTextMapBaggagePrefix+baggageKey, safeVal)
```
bensigelman: The keys are a problem, too. IMO, the only truly robust thing is to have the Tracer implementation completely control the keyspace. This either becomes `ot-baggage-item-0`, `ot-baggage-item-1`, etc, or we encode the entire lot as `ot-baggage-map` or similar. I prefer the latter, FWIW.

yurishkuro: agreed, but one thing at a time. The fact that `HTTPHeadersCarrier` was applying url-encoding was an actual bug in the API.

yurishkuro:
@bensigelman ok to merge?

yurishkuro:
also, this looks like v0.10.0, rather than v0.9.1 due to a change in the behavior.

mocktracer/mocktracer_test.go
```diff
@@ -111,23 +111,23 @@ func TestMockTracer_Propagation(t *testing.T) {
 	textCarrier := func() interface{} {
 		return opentracing.TextMapCarrier(make(map[string]string))
 	}
-	textLen := func (c interface{}) int {
```
bensigelman: (weird that `go fmt` didn't get applied to this?!)

propagation.go
```diff
@@ -147,21 +146,14 @@ type HTTPHeadersCarrier http.Header
 // Set conforms to the TextMapWriter interface.
 func (c HTTPHeadersCarrier) Set(key, val string) {
```
bensigelman: github won't let me comment on line 144, but I think we ought to include a quick example of how `HTTPHeadersCarrier` should be used in practice; or, alternatively, point to the examples in `Tracer.Inject` and `Tracer.Extract`.

bensigelman:
Sure, it's an improvement, let's :ship: it. Thanks for fixing.

I do want the various Tracer impls of the world to do something smarter about HTTP headers, but that's not a core OpenTracing-Go thing (and this is).

yurishkuro:
> also, this looks like v0.10.0, rather than v0.9.1 due to a change in the behavior.

@bensigelman thoughts on ^^^?

bensigelman:
@yurishkuro I agree about `v0.10.0`, yeah...

PR #114, opened 2016-09-22
Pass options along when creating a span with StartSpanFromContext

gwik:
Options were ignored when creating a span.

### DIFF
--- a/testtracer_test.go
+++ b/testtracer_test.go
@@ -33,25 +33,25 @@ type testSpan struct {
 	Tags          map[string]interface{}
 }
 
-func (s testSpan) Equal(os Span) bool {
+func (n testSpan) Equal(os Span) bool {
 	other, ok := os.(testSpan)
 	if !ok {
 		return false
 	}
-	if s.spanContext != other.spanContext {
+	if n.spanContext != other.spanContext {
 		return false
 	}
-	if s.OperationName != other.OperationName {
+	if n.OperationName != other.OperationName {
 		return false
 	}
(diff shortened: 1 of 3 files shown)


yurishkuro:
Do you mind adding unit tests?

gwik:
here you go.

testtracer_test.go
```diff
@@ -21,6 +21,7 @@ func nextFakeID() int {
 type testSpanContext struct {
 	HasParent bool
 	FakeID    int
+	Options   *StartSpanOptions
```
bensigelman: I'd have a slight preference to just add tags to `testSpan` and test that they get passed through correctly... if you don't feel like implementing this, I can do it :)

testtracer_test.go
```diff
@@ -28,6 +29,35 @@ func (n testSpanContext) ForeachBaggageItem(handler func(k, v string) bool) {}
 type testSpan struct {
 	spanContext   testSpanContext
 	OperationName string
+	StartTime     time.Time
+	Tags          map[string]interface{}
```
gwik: struct equality won't work with a map member, so I implemented an Equal method.

gwik: maybe adding an id and compare it would make more sense.

bensigelman: LGTM, thanks.

bensigelman:
Thanks @gwik... the test LGTM but I will give @yurishkuro a chance to provide feedback if he so desires...

PR #116, opened 2016-09-27
Add convenience methods to log.Field

RaduBerinde:
Currently the only way to print a `Field` is to define an `Encoder`. This is a
lot of code to write, especially if it's needed just for debugging or a quick
test. Adding a `Value()` method that returns the value as an `interface{}`
(which can be passed to `fmt.Print` functions), as well as a `String()` method.

Also exporting the key field.

CC @tschottdorf

### DIFF
--- a/log/field.go
+++ b/log/field.go
@@ -1,6 +1,9 @@
 package log
 
-import "math"
+import (
+	"fmt"
+	"math"
+)
 
 type fieldType int
 
@@ -20,7 +23,7 @@ const (
 )
 
 // Field instances are constructed via LogBool, LogString, and so on.
-// Tracing implementations may then handle them via the Field.Process
+// Tracing implementations may then handle them via the Field.Marshal
 // method.
 //
 // "heavily influenced by" (i.e., partially stolen from)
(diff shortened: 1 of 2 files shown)


log/field.go
```diff
@@ -20,13 +23,13 @@ const (
...
 // Field instances are constructed via LogBool, LogString, and so on.
-// Tracing implementations may then handle them via the Field.Process
+// Tracing implementations may then handle them via the Field.Marshal
 // method.
 //
 // "heavily influenced by" (i.e., partially stolen from)
 // https://github.com/uber-go/zap
 type Field struct {
-	key          string
+	Key          string
```
yurishkuro: would it make more sense to expose `Key()` instead? That would keep the `Field` immutable.

RaduBerinde: Good point!

log/field.go
```diff
@@ -198,3 +201,41 @@ func (lf Field) Marshal(visitor Encoder) {
...
+		return int64(lf.numericVal)
+	case uint32Type:
+		return uint32(lf.numericVal)
+	case uint64Type:
+		return uint64(lf.numericVal)
+	case float32Type:
+		return math.Float32frombits(uint32(lf.numericVal))
+	case float64Type:
+		return math.Float64frombits(uint64(lf.numericVal))
+	case errorType, objectType, lazyLoggerType:
```
bensigelman: I wonder what we should do for `lazyLoggerType`... I guess this is "fine for now", though I can imagine someone expecting to see the thing the lazy log function encodes instead.

log/field.go
```diff
@@ -20,7 +23,7 @@ const (
 )
 
 // Field instances are constructed via LogBool, LogString, and so on.
-// Tracing implementations may then handle them via the Field.Process
+// Tracing implementations may then handle them via the Field.Marshal
```
bensigelman: (thanks)

log/field.go
RaduBerinde: Ah.. Yeah I guess we would need to define an Encoder that captures the emitted key-value, and run the lazy logger. I'll file an issue.

BTW this makes me curious - why does the `Encoder` interface have an `EmitLazyLogger` method? Wouldn't any implementation of that (that follows the intended purpose) always just call the lazyLogger? Seems that instead of calling `visitor.EmitLazyLogger(lf.interfaceVal.(LazyLogger))` we could call `lf.interfaceVal.(LazyLogger)(visitor)` no?

bensigelman: IIRC this was adapted somewhat blindly from the `zap` approach... my only rationale would be that some impls might not want to do lazy logging at all, and the way things are structured now gives them more control (???). But I am just rationalizing.

yurishkuro: @RaduBerinde I think the Encoder serves two purposes. At the top level it allows the implementations to be told that a log field has a lazy encoder, so the respective method is needed. But then the same interface is reused recursively in the lazy encoder signature.

RaduBerinde: I see, thanks. There are a few more things that could use some clarification, I filed #118 so we can move the discussion there.

PR #123, opened 2016-11-09
Address race in nextMockID()

### LINKED ISSUE #107: nextMockID is flawed
As written using atomic add then load, it isn't guaranteed to return unique values. There is a race where multiple adds could happen before load causing the loads to all return the same value. Luckily there is an easy fix, atomic add returns the new value so just get rid of the load and return the value from atomic add.

billf:
Fixes #107

### DIFF
--- a/mocktracer/mockspan.go
+++ b/mocktracer/mockspan.go
@@ -28,8 +28,7 @@ type MockSpanContext struct {
 var mockIDSource = uint32(42)
 
 func nextMockID() int {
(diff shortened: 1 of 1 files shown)


yurishkuro:
@billf can you rebase it off master? Travis build should work then.

nicolaiskogheim:
Ping, because no one is notified when rebasing :)
Travis works now.

bensigelman:
@nicolaiskogheim thank you!

PR #128, opened 2016-11-30
Deprecate InitGlobalTracer in favor of SetGlobalTracer

yurishkuro:
* Clarify the purpose of NoopTracer

### DIFF
--- a/globaltracer.go
+++ b/globaltracer.go
@@ -4,18 +4,18 @@ var (
 	globalTracer Tracer = NoopTracer{}
 )
 
-// InitGlobalTracer sets the [singleton] opentracing.Tracer returned by
+// SetGlobalTracer sets the [singleton] opentracing.Tracer returned by
 // GlobalTracer(). Those who use GlobalTracer (rather than directly manage an
-// opentracing.Tracer instance) should call InitGlobalTracer as early as
+// opentracing.Tracer instance) should call SetGlobalTracer as early as
 // possible in main(), prior to calling the `StartSpan` global func below.
-// Prior to calling `InitGlobalTracer`, any Spans started via the `StartSpan`
+// Prior to calling `SetGlobalTracer`, any Spans started via the `StartSpan`
 // (etc) globals are noops.
-func InitGlobalTracer(tracer Tracer) {
+func SetGlobalTracer(tracer Tracer) {
 	globalTracer = tracer
 }
 
 // GlobalTracer returns the global singleton `Tracer` implementation.
-// Before `InitGlobalTracer()` is called, the `GlobalTracer()` is a noop
+// Before `SetGlobalTracer()` is called, the `GlobalTracer()` is a noop
 // implementation that drops all data handed to it.
 func GlobalTracer() Tracer {
 	return globalTracer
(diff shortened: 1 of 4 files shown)


bensigelman:
My only other thought would be to make the global tracer an exported `var`... unless we think the mutator (or accessor) will eventually do something more interesting?

yurishkuro:
I'm fine with a var. But GlobalTracer isn't backwards compatible with the current method name. DefaultTracer is an option.

bensigelman:
Meh, I thought about it some more and like what you have here. If for some reason we decide to do more than just set the (unexported) global var someday, this will be better.

bensigelman:
(i.e., LGTM)

bensigelman:
@kriskowal it's already aliased.

noop.go
```diff
@@ -2,8 +2,17 @@ package opentracing
...
+// for which all operations are no-ops.
+//
+// The primary use of this implementation is in libraries, such as RPC
+// frameworks, that make tracing an optional feature controlled by the
+// end user. A no-op implementation allows said libraries to use it
+// as the default Tracer and to write instrumentation that does
+// not need to keep checking if the tracer instance is nil.
+//
+// For the same reason, the NoopTracer is the default "global" tracer
+// (see GlobalTracer and SetGlobalTracer functions).
```
kriskowal: Please also note that NoopTracer is not suitable for RPC frameworks because it does not propagate baggage.

.travis.yml
```diff
@@ -1,7 +1,8 @@
 language: go
 
 go:
-  - 1.5
```
breerly: You can't drop support for 1.5 now that you've gone 1.0

globaltracer.go
```diff
@@ -4,18 +4,18 @@ var (
...
+// SetGlobalTracer sets the [singleton] opentracing.Tracer returned by
 // GlobalTracer(). Those who use GlobalTracer (rather than directly manage an
-// opentracing.Tracer instance) should call InitGlobalTracer as early as
+// opentracing.Tracer instance) should call SetGlobalTracer as early as
 // possible in main(), prior to calling the `StartSpan` global func below.
-// Prior to calling `InitGlobalTracer`, any Spans started via the `StartSpan`
+// Prior to calling `SetGlobalTracer`, any Spans started via the `StartSpan`
 // (etc) globals are noops.
-func InitGlobalTracer(tracer Tracer) {
+func SetGlobalTracer(tracer Tracer) {
```
breerly: Renaming this is a backwards breaking change. Suggest creating an alias instead, or incrementing the major version.

globaltracer.go
```diff
@@ -25,3 +25,8 @@ func GlobalTracer() Tracer {
 func StartSpan(operationName string, opts ...StartSpanOption) Span {
 	return globalTracer.StartSpan(operationName, opts...)
 }
+
+// InitGlobalTracer is deprecated. Please use SetGlobalTracer.
+func InitGlobalTracer(tracer Tracer) {
```
kriskowal: found it, thanks @bensigelman

globaltracer.go
breerly: Looks like there *is* an alias.

.travis.yml
yurishkuro: See #130

noop.go
yurishkuro: I added a warning below. But I don't think it's fair to say that it's not suitable for RPC frameworks - it is used in TChannel, for example. Even if we took a reverse approach and built tracing on top of distributed context propagation, there'd still be a case where API for context propagation would have a no-op implementation meaning "no context propagation", and you'd have the same problem.

kriskowal: Ok. It is perhaps too strong to claim that NoopTracer is suitable for RPC, since it would need to bring its own baggage propagation to the table like TChannel, just as it is too strong to claim that it is not suitable for RPC.

PR #131, opened 2016-12-30
log: avoid panic marshaling nil error

voutasaurus:


bensigelman:
Thanks for fixing this. I'll give people 24h to comment and merge if there are no objections.

yurishkuro:
A couple of questions:
  1. Why shouldn't it be the responsibility of the caller to check for `err != nil` before logging it to the span? To me, inserting an error field into the log implicitly says "there *was* an error".
  1.  If we want to redefine the semantics of `field.Error` to be sensitive to nil, then maybe it should skip emitting the value altogether instead of writing a string `"<nil>"` (although this could cause other issues like some backend expecting the list of log fields to be non-empty).

bensigelman:
@yurishkuro I am somewhat sympathetic to `2.` but think we should make a best effort to prevent crashes/panics in production due to the monitoring code.

voutasaurus:
This is an example where the assumption "LogFields won't panic" might hurt a production system:
```
if x, err := getX(); err != nil || invalid(x) {
    sp.LogFields(log.Error(err))
}
```

I'm open to not emitting anything on nil. As long as it doesn't panic I'll be happy.

PR #139, opened 2017-04-01
Add more semantic tags

stamm:
I added more semantic tags as describe on https://github.com/opentracing/specification/blob/master/semantic_conventions.md

### DIFF
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -15,7 +15,7 @@ import opentracing "github.com/opentracing/opentracing-go"
 //
 var (
 	//////////////////////////////////////////////////////////////////////
-	// SpanKind (client/server)
+	// SpanKind (client/server or producer/consumer)
 	//////////////////////////////////////////////////////////////////////
 
 	// SpanKind hints at relationship between spans, e.g. client/server
@@ -30,6 +30,16 @@ var (
 	// or other remote call
 	SpanKindRPCServerEnum = SpanKindEnum("server")
 	SpanKindRPCServer     = opentracing.Tag{Key: string(SpanKind), Value: SpanKindRPCServerEnum}
+
+	// SpanKindProducer marks a span representing the producer-side of a
+	// message bus
+	SpanKindProducerEnum = SpanKindEnum("producer")
+	SpanKindProducer     = opentracing.Tag{Key: string(SpanKind), Value: SpanKindProducerEnum}
+
+	// SpanKindConsumer marks a span representing the consumer-side of a
+	// message bus
+	SpanKindConsumerEnum = SpanKindEnum("consumer")
+	SpanKindConsumer     = opentracing.Tag{Key: string(SpanKind), Value: SpanKindConsumerEnum}
 
 	//////////////////////////////////////////////////////////////////////
 	// Component name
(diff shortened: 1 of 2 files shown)


ext/tags.go
```diff
@@ -55,6 +65,9 @@ var (
 	// PeerService records the service name of the peer
 	PeerService = stringTagName("peer.service")
 
+	// PeerAddress records the address name of the peer
```
bhs: this should have a more detailed comment since it's easy to misinterpret... maybe just adapt the text from the spec?

ext/tags.go
```diff
@@ -83,18 +96,67 @@ var (
...
+	// DB Tags
+	//////////////////////////////////////////////////////////////////////
+
+	// DBInstance is database instance name.
+	DBInstance = stringTagName("db.instance")
+
+	// DBStatement is a database statement for the given database type.
+	DBStatement = stringTagName("db.statement")
+
+	// DBType is a database type.
```
bhs: this one needs a lengthier explanation and perhaps 2-3 examples

ext/tags.go
```diff
@@ -83,18 +96,67 @@ var (
 	HTTPStatusCode = uint16TagName("http.status_code")
 
 	//////////////////////////////////////////////////////////////////////
+	// DB Tags
+	//////////////////////////////////////////////////////////////////////
+
+	// DBInstance is database instance name.
+	DBInstance = stringTagName("db.instance")
+
+	// DBStatement is a database statement for the given database type.
```
bhs: let's clarify that it can be a query or a prepared statement (i.e., before substitution)

ext/tags.go
```diff
@@ -83,18 +96,67 @@ var (
...
 
 	// Error indicates that operation represented by the span resulted in an error.
 	Error = boolTagName("error")
+
+	//////////////////////////////////////////////////////////////////////
+	// Log Tags
+	//////////////////////////////////////////////////////////////////////
+
+	// LogErrorKind is the type or kind of error.
+	LogErrorKind = stringTagName("error.kind")
```
bhs: in OT-Go, we would use `Field` to represent these logging concepts, not tags... let me know if you need more explanation / direction.

ext/tags.go
stamm: What about JDBC? Shall I write here something about database DSN?

ext/tags.go
stamm: Oh, yeap, now I see this `Field`. Thanks for info!

ext/tags.go
stamm: Done.

stamm:
@bhs Could you please look at [comment](https://github.com/opentracing/opentracing-go/pull/139#discussion_r109315715)?

ext/tags.go
bhs: I would be fine with the precise text from the table at https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table with the JDBC bit translated to the database DSN or simply removed... either way.

bhs:
@stamm done, sorry to miss it earlier.

ext/tags.go
stamm: I added description about database DSN.

ext/tags.go
```diff
@@ -95,6 +135,9 @@ var (
 // SpanKindEnum represents common span types
 type SpanKindEnum string
 
+// SpanLogEventEnum represent common span log events
+type SpanLogEventEnum string
```
bhs: this is unused – let's leave it off for now?

stamm: Of course it should be deleted

stamm:
@bhs yeap, my bad. I deleted unnecessary struct and did rebase to squash into single commit.

yurishkuro:
> did rebase to squash into single commit.

fyi, squashing is not necessary, this repo is configured to "squash & merge" automatically.

PR #142, opened 2017-04-21
Switch to http.Header.Set in HTTPHeadersCarrier

### LINKED ISSUE #141: TextMapWriter interface implementation inconsistency
Currently there are two implementations for the [TextMapWriter](https://github.com/opentracing/opentracing-go/blob/master/propagation.go#L95) interface: TextMapCarrier and HTTPHeadersCarrier. The behaviour of the Set operation however is very different. In the case of TextMapCarrier, we would overwrite a key:

```
// Set implements Set() of opentracing.TextMapWriter
func (c TextMapCarrier) Set(key, val string) {
	c[key] = val
}
```

With HTTPHeaderCarrier, we append to the value if a key ( header ) exists: 

```
// Set conforms to the TextMapWriter interface.
func (c HTTPHeadersCarrier) Set(key, val string) {
       h := http.Header(c)
       h.Add(key, val)
}
```

To me this is confusing and it also makes it harder to implement a consistent trace propagation behaviour.

Wouldn't it make sense to use header.Set instead of header.Add for the HTTPHeaderCarrier implementation?

boyand:
Fixes #141

### DIFF
--- a/propagation.go
+++ b/propagation.go
@@ -160,7 +160,7 @@ type HTTPHeadersCarrier http.Header
 // Set conforms to the TextMapWriter interface.
 func (c HTTPHeadersCarrier) Set(key, val string) {
 	h := http.Header(c)
-	h.Add(key, val)
+	h.Set(key, val)
 }
 
 // ForeachKey conforms to the TextMapReader interface.


yurishkuro:
-1

This is a potentially breaking change. For example, some tracers may implement baggage over HTTP like this:
```
baggage: key1=val1
baggage: key2=val2
```
This change breaks them.

boyand:
Agree but them working today is merely because of a side effect of the implementation? Currently the TextMapCarrier and the HTTPHeadersCarrier behave differently so each tracer implementation based on opentracing-go would have to compensate anyhow. This becomes especially hard if you need to use them both ( as we do since we support both RPC like and HTTP like comms )

My points are that: 

1. The current Set implementation of the TextMapCarrier is misleading
2. If a tracer is using the current logic to enrich an existing HTTP Header Set, this may work for the baggage, but surely won't work for TraceId and SpanId headers. Otherwise you would get behaviour like:

```
ot-tracer-traceid: 1,1
ot-tracer-spanid: 1,2
```

yurishkuro:
I am not sure I follow what is confusing. The text map is a single-value map, thus it has Set but no Add. The HTTP headers, on the other hand, is a multi-value map, so Add() makes more sense. There are many existing tracers that implement that API already, without issues.

We actually ran into a situation similar to your example, where trace headers were duplicated when leaving the service. The problem wasn't with the tracer, but with the service itself having unnecessary logic to copy all headers from inbound request to the outbound request. Even if we had HTTP carrier to use Set instead of Add, the same duplication could've occurred, depending on when tracing headers vs. application headers were added to the outbound request.

boyand:
Thanks for the quick response, @yurishkuro. Given there are already many tracers using this logic, then I am going to decline the PR.

In our case headers get duplicated when they enter a service if a parent span already exists. Since the middleware injects the new headers into the request before it is passed to the handler, I can't really think of a good way to address this issue. We need to preserve all original headers therefore the only remaining option I can think of would be to remove any pre-existing tracing headers from the request prior to injection. 

Did you have to deal with similar cases?

yurishkuro:
@boyand yes, we ran into this one as well, with tchannel-based services which were, for some reason, verifying that the requests contained only known application headers. What we did there is we implemented opentracing middleware in tchannel in such a way that it injects tracing headers just before the request leaves the service, and on the inbound side it extracts them and _removes them from the request_, so that by the time the request reaches the application's handler it never sees the tracing headers.

This was easier to implement because tchannel is its own protocol, and we could use an additional prefix known to tchannel code to identify tracing headers (even though we don't know the actual header names since they are controlled by the OT tracer). The same is a bit harder with HTTP, as it needs to happen in all HTTP instrumentations.

PR #145, opened 2017-05-06
Fix some usage examples

cce:
While working on TraceView's OT tracer and porting the [opentracing-python API test harness](https://github.com/opentracing/opentracing-python/blob/master/opentracing/harness/api_check.py) to Go, I noticed not all the usage examples worked against the current OT API.

I also changed "spanContext" to "clientContext" in the propagation.go HTTP Extract examples to match the name @bhs used in tracer.go — hope that is OK!

### DIFF
--- a/propagation.go
+++ b/propagation.go
@@ -72,18 +72,18 @@ const (
 	//
 	// For Tracer.Extract(): the carrier must be a `TextMapReader`.
 	//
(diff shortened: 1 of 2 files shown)


yurishkuro:
Thanks!

bhs:
(thanks++)

PR #146, opened 2017-05-06
Port of Python opentracing/harness/api_check.py to Go

cce:
This PR aims to provide a suite of helper methods for testing tracers, based on Python's [opentracing/harness/api_check.py](https://github.com/opentracing/opentracing-python/blob/master/opentracing/harness/api_check.py) (including all of @yurishkuro's test strings and cases)

Like in the Python OT library's [test_api.py](https://github.com/opentracing/opentracing-python/blob/master/tests/test_api.py), I tried wiring the tests up to the Go NoopTracer, but the ported assertions failed because the Python no-op tracer behaves differently from Go's.

Specifically: Python's Extract() returns a "[noop_span_context](https://github.com/opentracing/opentracing-python/blob/master/opentracing/tracer.py#L134)" with no error (unless the specified format is not one of text, binary, or HTTP), whereas the Go NoopTracer's Extract() always returns a nil SpanContext and the ErrSpanContextNotFound error. So all the the propagation tests are failing for NoopTracer, as is TestUnknownFormat since NoopTracer always returns ErrSpanContextNotFound regardless of format.

I was thinking to update the tests so that NoopTracer passes, but what I really want to do is figure out how to provide some output about the various features different tracers provide, like whether they propagate baggage or context correctly, handle different formats, handle HTTP header string corner cases, etc.

As for code structure, open to comments on how to organize this better to make it flexible and easy to use for tracer implementors. As you can see there's a fair amount of boilerplate required to plug in a tracer that isn't necessary when using the Python APICompatibilityCheckMixin. It might also be nice to use Go 1.7's [subtests](https://blog.golang.org/subtests) to move more of the test logic into the harness, but I noticed your Travis builds include 1.6.

I could also try and consolidate the options struct with the tracer, but I wanted to ensure that there was no tracer state shared between tests by pushing that off to the caller of the check functions. I could also move the check functions to be methods of a struct containing the opts, and am open to any suggestions on naming, structure, or anything else!


cce:
The other point I wanted to make was that since many tracer behaviors are optional, I don't know if it's fair to have Go tests that fail for those optional behaviors, so I would love to hear ideas for how to capture or output those differences between tracers without having FAIL show up. Perhaps log lines are enough to start e.g. "OPTIONAL: Baggage propagation support not detected."

harness/api_checkers.go
```diff
@@ -0,0 +1,190 @@
...
+		SetTag("unicode_val", "non-ascii: \u200b").
+		SetTag("unicode_key_\u200b", "ascii val")
+	span.Finish()
+}
+
+func CheckSpanLogs(t *testing.T, tracer opentracing.Tracer, opts CheckOpts) {
+	span := tracer.StartSpan("Fry")
+	span.LogKV(
+		"frozen.year", 1999,
+		"frozen.place", "Cryogenics Labs")
```
yurishkuro: given recent semantic specs I would've written this `event=frozen year=1999 place=xxx`

yurishkuro:
I opened a PR to remove 1.6, but since we're using testify already, it has support for test suites.

I think calling the actual test functions should be automated. The concrete implementation needs to provide a function `func() (tracer opentracing.Tracer, closer func())` so that the harness can create and dispose new instance of the tracer for each test.

In Python harness we had a "capabilities" descriptor (with just one flag, baggage yes/no, but it could be more elaborate if necessary), and the harness adjusts its behavior depending on the capabilities, e.g. by skipping some tests (and yes, logging something like "XXX not supported").

Baggage propagation can be tested with sampling on and off, as much as it can be affected by the `sampling.priority` tag.

The capturing of tags and logs can be tested by an optional Probe in the Capabilities struct that would allow reading the value of a tag and/or log entries. It's hard to make portable since the spec does not dictate any specific behavior for tags & logs, but the harness could define _support levels_ for those and if the Tracer declares a certain level in its Capabilities, then the harness can test for it.

cce:
Cool, I hadn't used testify suites before, thanks for the tip! I've refactor the API checkers around a testify suite and used a callback like you described for creating and disposing of tracers for each test. This relies on the [BeforeTest/AfterTest interfaces added to testify in December 2016](https://github.com/stretchr/testify/pull/318) though, which I didn't have in my environment until pulling the latest, and I was on the fence about whether to use them or write my own.

I saw the Python `check_baggage_values` flag and ported it to `APICheckOpts.CheckBaggageValues` in this PR, and also ported the conditional checks from Python. It will take another option or two to get the Go NoopTracer to pass, though, since its `Extract()` always returns `nil, ErrSpanContextNotFound`. I could rename it to Capabilities and think a little more about how to add a Probe function/interface for reading span tags and logs from tests.

Another nice thing to be able to ask the Probe would be "are these two spans in the same trace?" for testing context propagation.

cce:
To get the tests to pass for `NoopTracer`, I added CheckExtract and CheckInject flags to the Capabilities struct. I also added a Probe interface with "is this the same trace" methods and added it to the `TestStartSpanWithParent` test as well as the context propagation tests. The tests all pass for `NoopTracer` and I'm trying them out with my own tracer as well as basictracer next.

cce:
I gave a try wiring this up to basictracer, which is now passing all the API checks in https://github.com/opentracing/basictracer-go/compare/master...cce:api_check

For the Probe interface you suggested, I think it will need to be added as another return value of the function `func() (tracer opentracing.Tracer, closer func())` that makes a Tracer for each test if it needs to examine span data recorded by basictracer's in-memory recorder like some of the other basictracer do.

cce:
I added some more checks to improve coverage around cases like Inject/Extract returning `ot.ErrInvalidSpanContext` or `ot.ErrInvalidCarrier` when bad carriers are provided; a test for `ForeachBaggageItem` returning early if the handler returns false; and calls to FinishWithOptions with LogRecords as well as the deprecated log methods.

One could imagine using the (still unimplemented in this PR; coming soon) Probe interface and adding extra capability flags to check for support for recording log data, as well as to flag whether to check for support for the deprecated log methods.

harness/api_checkers.go
```diff
@@ -0,0 +1,400 @@
...
+	"time"
+
+	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/log"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+// APICheckCapabilities describes options used by APICheckSuite when testing a Tracer.
+type APICheckCapabilities struct {
```
bhs: nit / thought: a "functional options" pattern would make it easy to add helpers like `harness.CheckEverything()`

harness/api_checkers.go
```diff
@@ -0,0 +1,400 @@
...
+	CheckBaggageValues bool          // whether to check for propagation of baggage values
+	CheckExtract       bool          // whether to check if extracting contexts from carriers works
+	CheckInject        bool          // whether to check if injecting contexts works
+	Probe              APICheckProbe // optional interface providing methods to check recorded data
+}
+
+// APICheckProbe exposes methods for testing data recorded by a Tracer.
+type APICheckProbe interface {
+	SameTrace(first, second opentracing.Span) bool // whether two spans are from the same trace
+	SameTraceContext(opentracing.Span, opentracing.SpanContext) bool
```
bhs: you mean `SameSpanContext`? please provide a one-line comment about what this is checking. I assume that the `Span`'s `SpanContext` has the same carrier representation as the second (`SpanContext`) argument, even if it's a different actual object/address.

harness/api_checkers.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+	s.tracer, s.closer = nil, nil
+}
+
+// NewAPICheckSuite returns a testify suite for checking a Tracer against the OpenTracing API.
+// It is provided a function that will be executed to create and destroy a tracer for each test
+// in the suite, and API test options described by APICheckCapabilities.
+func NewAPICheckSuite(
+	newTracer func() (tracer opentracing.Tracer, closer func()),
+	opts APICheckCapabilities) *APICheckSuite {
```
tedsuo: please put the argument `opts APICheckCapabilities` on it's own line, it's hard to read in this form.

harness/api_checkers.go
tedsuo: Or at least a constructor that turned everything on by default, I would like to opt out of tests rather than accidentally miss new test types.

tedsuo:
In general, I think this is great. 👍

harness/api_checkers.go
cce: OK I gave a try at implementing functional options based on Tracer.StartSpan — let me know what you think.

harness/api_checkers.go
cce: Originally I thought this would assert a span and context were simply part of the same trace, but looking at my test cases it is actually being used to assert an extracted context matches the span it came from after being serialized & deserialized, and comments.

harness/api_checkers.go
```diff
@@ -0,0 +1,437 @@
...
+}
+
+// TestSpanBaggage tests calls to set and get span baggage, and if the CheckBaggageValues option
+// is set, asserts that baggage values were successfully retrieved.
+func (s *APICheckSuite) TestSpanBaggage() {
+	span := s.tracer.StartSpan("Fry")
+	assertEmptyBaggage(s.T(), span.Context())
+
+	spanRef := span.SetBaggageItem("Kiff-loves", "Amy")
+	assert.Exactly(s.T(), spanRef, span)
```
yurishkuro: I thought Suite had all those assert methods on itself, i.e. instead of passing `s.T()` you can just do `s.Equals()`

harness/api_checkers.go
```diff
@@ -0,0 +1,437 @@
...
+	span.Finish()
+}
+
+// TestHTTPPropagation tests if the Tracer can Inject a span into HTTP headers, and later Extract it.
+// If CheckExtract is set, it will check if Extract was successful (returned no error). If a Probe is set,
+// it will check if the extracted context is in the same trace as the original span.
+func (s *APICheckSuite) TestHTTPPropagation() {
+	span := s.tracer.StartSpan("Bender")
+	textCarrier := opentracing.HTTPHeadersCarrier{}
+	// XXX add same test cases around valid HTTP header characters, casing
```
yurishkuro: `s/XXX/TODO/` ?  XXX means "hack, ugly code"

http://legacy.python.org/dev/peps/pep-0350/#mnemonics

yurishkuro: btw, the comment is especially relevant to baggage, because it had different requirements in the past (allowed more limitations), and not all tracers support it correctly. I would make that "strict" check controllable via capabilities flag.

harness/api_checkers.go
```diff
@@ -0,0 +1,437 @@
...
+	assert.NoError(s.T(), err)
+
+	extractedContext, err := s.tracer.Extract(opentracing.HTTPHeaders, textCarrier)
+	if s.opts.CheckExtract {
+		assert.NoError(s.T(), err)
+		assertEmptyBaggage(s.T(), extractedContext)
+	} else {
+		s.T().Log("Tracer.Extract not supported, skipping")
+	}
+	if s.opts.Probe != nil {
```
yurishkuro: Do you think it's possible to refactor the tests in such a way that when we run into a capabilities situation, we can actually `t.Skip()` instead of ignoring the check? It would provide a much nicer summary when the harness is run - you could clearly see which tests were not performed because of capabilities settings.

yurishkuro: PS: doesn't need to be in this PR, just a nice to have

yurishkuro: PPS: if refactoring this way is too hard, then I think these should be `else { t.Log(.. blah skipped blah ..) }`

harness/api_checkers.go
cce: oh right that XXX was because I was thinking of adding some tests around adding baggage with differently-cased key names to see if they overwrite each other... hrmm.

harness/api_checkers.go
```diff
@@ -0,0 +1,437 @@
...
+	s := &APICheckSuite{newTracer: newTracer}
+	for _, o := range opts {
+		o.Apply(s)
+	}
+	return s
+}
+
+// APICheckOption instances may be passed to NewAPICheckSuite.
+type APICheckOption interface {
+	Apply(*APICheckSuite)
```
cce: that is simpler ... I was following the example from OT-Go here https://github.com/opentracing/opentracing-go/blob/master/tracer.go#L154-L156

harness/api_checkers.go
cce: wow cool! OK!

harness/api_checkers.go
cce: Hrmmm I can definitely add else Log calls everywhere I missed ... but to do what you're describing I think it would be much easier to use Go 1.7 subtests to call `t.Skip` from a subtest rather than refactoring / ripping apart the test cases into separate test functions depending on capabilities.

cce: The TestInvalidInject and TestInvalidExtract tests were straightforward enough to switch to using t.Skip so I switched them over, and updated the other log messages to be clear about which capabilities were causing checks to be skipped.

harness/api_checkers.go
cce: also XXX can also mean TODO or FIXME or just "needs attention"(https://stackoverflow.com/questions/1452934/what-is-the-meaning-of-xxx) ... I think the culture of C hacking that I learned from uses XXX to flag things or questions that should be addressed later. Didn't mean to commit the XXX in the final version of this PR though.

harness/api_checkers.go
cce: Hrmm I started to switch the options pattern to use a global "options factory" variable like jaeger's `var TracerOptions tracerOptions` example but then stopped because I got worried it would be confusing to have two different options patterns in the OT-Go repo — what do you think?

harness/api_checkers.go
cce: The OT Spec doesn't really talk about serializing baggage names into HTTP headers, but I added a new test (`TestHTTPBaggage`) to ensure that differently-cased versions of the same case-insensitive baggage key name get combined into a single item after inject/extract.

harness/api_checkers.go
yurishkuro: the "options factory" is just a namespacing trick, you don't need it if the namespace is small enough like that of `harness`. I was simply referring to defining Option as a function, rather than interface, it's more Go-ish (I guess we dropped the ball when doing `StartSpanOption` :-) ). Not a big deal either way.

harness/api_checkers.go
```diff
@@ -0,0 +1,477 @@
...
+		"Leela",
+		opentracing.ChildOf(parentSpan.Context()))
+	span.Finish()
+
+	span = s.tracer.StartSpan(
+		"Leela",
+		opentracing.FollowsFrom(parentSpan.Context()),
+		opentracing.Tag{Key: "birthplace", Value: "sewers"})
+	if s.opts.Probe != nil {
+		s.True(s.opts.Probe.SameTrace(parentSpan, span))
```
yurishkuro: this only runs on the FollowsFrom span, not the ChildOf span, intentional?

cce: Nope, thanks for catching, I added another Probe.SameTrace check for the ChildOf span.

harness/api_checkers.go
```diff
@@ -0,0 +1,482 @@
...
+func (s *APICheckSuite) TestInvalidExtract() {
+	if !s.opts.CheckExtract {
+		s.T().Skip("CheckExtract capability not set, skipping")
+	}
+	span := s.tracer.StartSpan("op")
+
+	// binary extract
+	ctx, err := span.Tracer().Extract(opentracing.Binary, NotACarrier{})
+	s.Equal(opentracing.ErrInvalidCarrier, err, "Carrier that's not io.Reader should return error")
+	s.Nil(ctx)
```
yurishkuro: these null checks failed when I ran the test against Jaeger implementation. I don't think this is a requirement that the returned context is nil.

harness/api_checkers.go
```diff
@@ -0,0 +1,482 @@
...
+	Apply(*APICheckSuite)
+}
+
+// Apply satisfies the APICheckOption interface.
+func (c APICheckCapabilities) Apply(s *APICheckSuite) {
+	s.opts = c
+}
+
+// CheckEverything enables all API checks.
+type CheckEverything struct{}
```
yurishkuro: I think using structs exposes unnecessary complexity. If you define the Option as a function rather than interface, then you only need public methods like `func WithProbe(probe Probe)`

harness/api_checkers.go
```diff
@@ -0,0 +1,482 @@
...
+	closer    func()
+}
+
+// NewAPICheckSuite returns a testify suite for checking a Tracer against the OpenTracing API.
+// It is provided a function that will be executed to create and destroy a tracer for each test
+// in the suite, and API test options described by APICheckCapabilities.
+func NewAPICheckSuite(
+	newTracer func() (tracer opentracing.Tracer, closer func()),
+	opts ...APICheckOption,
+) *APICheckSuite {
```
yurishkuro: it's not necessary to expose the users of this package to the fact that it uses testify suite under the cover. This function can be changed to `RunAPIChecks(t *testing.T, newTracer func..., options...)`

cce: Will fix..

harness/api_checkers.go
cce: Interesting — I was basing this off of the Tracer interface description [here](https://github.com/opentracing/opentracing-go/blob/master/tracer.go#L100-L109):

> Return values:
> - A successful Extract returns a SpanContext instance and a nil error
> - If there was simply no SpanContext to extract in `carrier`, Extract() returns (nil, opentracing.ErrSpanContextNotFound)
> - If `format` is unsupported or unrecognized, Extract() returns (nil, opentracing.ErrUnsupportedFormat)
> - If there are more fundamental problems with the `carrier` object, Extract() may return opentracing.ErrInvalidCarrier, opentracing.ErrSpanContextCorrupted, or implementation-specific errors.

cce: Hrmm I guess the third bullet does not specify (nil, opentracing.ErrInvalidCarrier).

harness/api_checkers.go
cce: Cool, updated to just use functions.

harness/api_checkers.go
cce: Cool I tried the options factory thing, then trimmed it down to just define some functions without the global factory.

harness/api_checkers.go
cce: I would be happy to remove this assertion on the nil return for ErrInvalidCarrier/ErrSpanContextCorrupted since the language in the comment doesn't directly address it, but I would also recommend updating the Tracer interface comment to mention ctx could be non-nil.

cce: Here's that change, WDYT? https://github.com/cce/opentracing-go/compare/api_check...api_check_notnil-invalidcarrier

yurishkuro: On second thought, I think what you have is good, and we should update the comments in the API. It is indeed a convention in Go to return nil when there's an error, and it's not difficult for us to change Jaeger to support that.

cce:
@yurishkuro WDYT about the latest? I think I addressed all your comments so far.

harness/api_checkers.go
```diff
@@ -0,0 +1,493 @@
...
+	return func(s *APICheckSuite) {
+		s.opts.Probe = probe
+	}
+}
+
+// BeforeTest creates a tracer for this specific test invocation.
+func (s *APICheckSuite) BeforeTest(suiteName, testName string) {
+	s.tracer, s.closer = s.newTracer()
+	if s.tracer == nil {
+		panic(fmt.Sprintf("newTracer returned nil Tracer before running %s, %s", suiteName, testName))
```
yurishkuro: `s.T.Fatalf(...)` might be more appropriate

harness/api_checkers.go
```diff
@@ -0,0 +1,493 @@
...
+		opentracing.Tag{Key: "birthday", Value: "August 14 1974"})
+	span.LogFields(
+		log.String("hospital", "Brooklyn Pre-Med Hospital"),
+		log.String("city", "Old New York"))
+	span.Finish()
+}
+
+// TestStartSpanWithParent checks if a Tracer can start a span with a specified parent.
+func (s *APICheckSuite) TestStartSpanWithParent() {
+	parentSpan := s.tracer.StartSpan("parent")
```
yurishkuro: s/"parent"/"Turanga Morris"/  (or Turanga Munda). 

Damn, we need a ref type for >1 parent! :-)

harness/api_checkers.go
```diff
@@ -0,0 +1,493 @@
...
+
+	span := s.tracer.StartSpan(
+		"Leela",
+		opentracing.ChildOf(parentSpan.Context()))
+	span.Finish()
+	if s.opts.Probe != nil {
+		s.True(s.opts.Probe.SameTrace(parentSpan, span))
+	} else {
+		s.T().Log("harness.Probe not specified, skipping")
+	}
```
yurishkuro: could DRY this a bit
```
childFns := []func(opentracing.SpanContext) opentracing.Reference { 
    opentracing.ChildOf, 
    opentracing.FollowsFrom,
}
for _, childFn := childFns {
   //...
}
```

harness/api_checkers.go
```diff
@@ -0,0 +1,493 @@
...
+	// check baggage items
+	count := 0
+	items := make(map[string]string)
+	extractedContext.ForeachBaggageItem(func(k, v string) bool {
+		count++
+		items[k] = v
+		return true
+	})
+	s.Len(items, 2, "HTTP Baggage propagation seems to depend on case-sensitive HTTP header field names")
+	s.Equal(count, 2, "HTTP Baggage propagation seems to depend on case-sensitive HTTP header field names")
```
yurishkuro: hm, I am not sure about this one. It is indeed how Jaeger's baggage implemented, but case-insensitivity was in the spec some time ago and has since been removed, baggage keys are now any Unicode strings.

harness/api_checkers.go
cce: Sure, I agree!

harness/api_checkers.go
cce: Oh cool, I wasn't aware of that — I was probably reading basictracer and jaeger — how does that work with HTTP headers? I could get rid of the test but I don't know that it's good practice for a tracer to try and send two different HTTP headers with different casing because field names are supposed to be case-insensitive.

yurishkuro: If you are asking how it works in basic or Jaeger, the answer is it doesn't. Unicode keys are nice in theory, but we don't recommend then to users. I'm not sure if Lightstep supports those. We also recommend lower-snake-case keys only.

It's not too difficult to support arbitrary keys, but they would have to be encoded, which makes them unreadable when inspecting the traffic/headers. 

For tracers that use a binary blob the arbitrary keys are easy. We choose to keep baggage in clear text and defer Unicode support until someone really complains. 

Of course if we make that as part of the harness we'd need to fix that in Jaeger :-)

cce: OK, I'll remove this, thanks for the info. I don't see any mention of Unicode in https://github.com/opentracing/specification/blob/master/specification.md#set-a-baggage-item but if we want to require tracers to live up to that spec we could bring back a test like this that exercises weird mixed-case and Unicode baggage key names requiring encoding for correct propagation.

cce:
@yurishkuro OK I think this is finally ready for merge..

cce:
Aha I was worried about users not running the latest testify Before/After features but didn't realize they didn't tag a release with those features. The Setup/Teardown methods from `^1.1.3` provide the same functionality (but without suite/test names as inputs) and I'll switch over to those instead.

cce:
For documentation I'll update the README to mention it, and have it link to godoc of the harness package where I'll add a usage example for RunAPIChecks.

yurishkuro:
I was thinking of using both before/after and setup/teardown so that both master and 1.1.3 of testify would work.

I just put a PR #157 that performs clean install of the dependencies, not sure if your change will work with that

cce:
Well, I finally wrote a package comment and added a mention in the README for this package — I think it's ready to go.

yurishkuro:
:shipit:

PR #147, opened 2017-05-06
Stop testing with Go 1.6

yurishkuro:
Mainly because Go 1.6 does not allow nested tests. Go 1.8 has been released already, do we need to be testing with more than two recent versions?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 
 go:
-  - 1.6
   - 1.7
   - 1.8
   - tip


gchaincl:
:+1:

bhs:
Light googling failed to reveal the approximate distribution of Go versions in production... would be great to understand that before making this change. (I also wish it was possible to support different Go versions for tests and non-tests respectively)

tedsuo:
In general, if we support a language version we should test with it. But in this particular case, I agree that the only thing we really need to test is if it still compiles in 1.6. 

If we're going to drop support for 1.6 that would help with a number of things, such as the `context` package mismatch it inflicts on us.

tonyghita:
This should be merged before #176 can pass. @yurishkuro are you able to resolve the merge conflicts on this PR so we can move forward with the change?

PR #150, opened 2017-06-17
Add a Skip log field type to log to allow for optional fields

savaki:
prevents cluttering of the field list with optional fields that would otherwise be encoded as opt1:nil, opt2:nil, etc

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,21 @@ reference.
     }
 ```
 
+#### Conditionally capture a field using `log.Noop`
+
+In some situations, you may want to dynamically decide whether or not
+to log a field.  For example, you may want to capture additional data,
+such as a customer ID, in non-production environments:
+
+```go
+    func Customer(order *Order) log.Field {
+        if os.Getenv("ENVIRONMENT") == "dev" {
+            return log.String("customer", order.Customer.ID)
+        }
+        return log.Noop()
+    }
+```
+
 #### Goroutine-safety
 
 The entire public API is goroutine-safe and does not require external
--- a/log/field.go
+++ b/log/field.go
@@ -20,7 +20,7 @@ const (
 	errorType
 	objectType
 	lazyLoggerType
-	skipType
+	noopType
 )
 
 // Field instances are constructed via LogBool, LogString, and so on.
@@ -153,7 +153,7 @@ func Lazy(ll LazyLogger) Field {
 	}
 }
 
-// Skip creates a no-op log field that should be ignored by the tracer.
+// Noop creates a no-op log field that should be ignored by the tracer.
 // It can be used to capture optional fields, for example those that should
 // only be logged in non-production environment:
 //
@@ -161,14 +161,14 @@ func Lazy(ll LazyLogger) Field {
 //          if os.Getenv("ENVIRONMENT") == "dev" {
 //              return log.String("customer", order.Customer.ID)
 //          }
-//          return log.Skip()
+//          return log.Noop()
 //     }
 //
 //     span.LogFields(log.String("event", "purchase"), customerField(order))
 //
-func Skip() Field {
+func Noop() Field {
 	return Field{
-		fieldType: skipType,
+		fieldType: noopType,
 	}
 }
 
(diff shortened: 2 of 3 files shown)


yurishkuro:
I'm not sure I follow why this feature is useful. Why bother passing anything to the logging call if it's not meant to be used?

savaki:
Consider the Error use case.  Error is defined as a higher type that converts into log.String("error", "")  resulting in an err="" in the trace.  Skip allows those higher order types to decide whether or not to actually place something in the trace dynamically.

Imagine the ```my``` package contains two higher order field functions, X and Y which I would only like to generate trace info under certain conditions.  I would like to be able to write the following:

```
span.LogFields(my.X(x), my.Y(y)
```

Instead of

```
fields := []log.Field{}
if x condition is met {
  fields = append(fields, my.X(x))
}
if y condition is met {
  fields = append(fields, my.Y(y))
}
span.LogFields(fields...)
```

yurishkuro:
OK, that makes sense. I would suggesting adding a more detailed comment to the Skip() function.

It looks like the change is actually backwards compatible.

log/field.go
```diff
@@ -152,6 +153,22 @@ func Lazy(ll LazyLogger) Field {
 	}
 }
 
+// Skip allows a field to be optionally be included in a Field list.
```
yurishkuro: I find the `error` example to be misleading, because it's not idiomatic to Go to do something with the error object without checking that it's not actually nil.  How about the following description:

```
// Skip creates a no-op log field that should be ignored by the tracer.
// It can be used to capture optional fields, for example those that should
// only be logged in non-production environment:
//
//     func cunstomerField(order *Order) log.Field {
//          if os.Getenv("ENVIRONMENT") == "dev" {
//              return log.Field("customer", order.Cusomer.ID)
//          }
//          return log.Skip()
//     }
//
//     span.LogFields(log.String("event", "purchase"), customerField(order))
//
```

log/field.go
```diff
@@ -152,6 +153,22 @@ func Lazy(ll LazyLogger) Field {
...
+//
+// Useful for the construction of composed field types like log.Error
+// to enable them to optionally decide where or not to trace content.
+//
+// There are a number of reasons one might wish to dynamically include
+// fields.  For example, if you want to provide additional trace information
+// in non-production environment only.  Another example is if you have
+// fields that may not always be present.
+func Skip(key string) Field {
```
yurishkuro: I don't think it should be accepting a `key`

savaki: That's my preference as well, but I wasn't sure how strong the conventions were for accepting a param.  I'll go ahead and make these changes.

savaki:
How's that?

log/field.go
```diff
@@ -152,6 +153,26 @@ func Lazy(ll LazyLogger) Field {
...
+//              return log.String("customer", order.Customer.ID)
+//          }
+//          return log.Skip()
+//     }
+//
+//     span.LogFields(log.String("event", "purchase"), customerField(order))
+//
+func Skip() Field {
+	return Field{
+		key:       "_",
```
yurishkuro: this can be skipped, zero string value is just as good, we only need to initialize the fieldType

yurishkuro:
@bhs @tedsuo any comments?

tedsuo:
I like the idea. Given that `log.Field` is a non-interface, non-pointer value in our APIs, I don't see how we could use `nil` for this, so an explicit log type seems right. I agree with Ben that `log.Noop` would be more consistent with our naming conventions than `log.Skip`.

savaki:
Let me know if there's anything else you'd like with this pull request

README.md
```diff
@@ -1,4 +1,5 @@
 [![Gitter chat](http://img.shields.io/badge/gitter-join%20chat%20%E2%86%92-brightgreen.svg)](https://gitter.im/opentracing/public) [![Build Status](https://travis-ci.org/opentracing/opentracing-go.svg?branch=master)](https://travis-ci.org/opentracing/opentracing-go) [![GoDoc](https://godoc.org/github.com/opentracing/opentracing-go?status.svg)](http://godoc.org/github.com/opentracing/opentracing-go)
+[![Sourcegraph Badge](https://sourcegraph.com/github.com/opentracing/opentracing-go/-/badge.svg)](https://sourcegraph.com/github.com/opentracing/opentracing-go?badge)
```
yurishkuro: Any idea why this is showing up in the diff? Your fork setup seems a bit unusual, merging master to master. Typically you set up your private master to follow the upstream and do PR from private:branch -> upstream:master

Issue #149, opened 2017-06-06
StartSpanFromContext() to use parentSpan.Tracer() 


Loofort:
Hello

I have a question.
Is there any reason why StartSpanFromContext() func uses GlobalTracer() instead parentSpan.Tracer() ?
Isn't it more logical to use the second option?

yurishkuro:
If you look at it differently, when there is no parent span, then global tracer is the only option. So isn't it more logical to use the same tracer in both scenarios? Global tracer is an implicit dependency of that function.

Loofort:
Well, my goal is to create child span that implements the same type (behavior) as a parent.
Probably you could suggest me a proper way to do that.

Let me describe a use case:
My service imports library that utilizes opentracing. Internally it may spawn goroutines with child spans.

The service should write logs records associated with a  transaction. So I've made log wrapper for the global tracer.
This tracer also produces spans that encapsulate transaction info required for correct logging.

However, it seems there is no way to pass the transaction object to child spans. So the logger is unable to find the transaction object associated with parent span.

As a workaround I have an idea to encapsulate the transaction object to Tracer that could be obtained by span.Tracer() 
This way we also get more flexible way in span creation. That could be suitable for big monolithic app - custom log/metrics workflow for particular routes (code branches)

The other idea is to use SpanContext , but it doesn't have a suitable method, so I'd have to make type assertion. and in general, it doesn't seem like proper way.

yurishkuro:
You lost me a bit with "to pass the transaction object to child spans". If "transaction object" is something specific to your application and you want to propagate it similar to how the spans are propagated, you may want to store it explicitly in the Context.

As for StartSpanFromContext(), I personally never used this function, exactly because it requires a global tracer for at least one of its code paths (when there is no parent). I find it better to build instrumentation functions to expect the tracer to be passed explicitly, and StartSpanFromContext() is _really_ simple to worry about repeating that code manually in the instrumentation.

Loofort:
The "transaction object" indeed is something specific , that my custom tracer-logger wants to use.
Indeed this object is stored in context, however the Tracer doesn't consider the context.Context while creating a span ( tracer.StartSpan() doesn't accept context )

StartSpanFromContext() is simple indeed , but it is part of public API that is used by third party packages.
And when you use those packages and pass a span created by your custom tracer (not global), you could face with unexpected behavior.

[...]

PR #152, opened 2017-06-26
Export StartSpanFromContextWithTracer

devoxel:
Originally discussed in  #149 . This is a fairly simple PR. It just exports a helper function that already exists.

It can be useful for people not using the global tracer. It's only a small thing really, but still, could be nice.

### DIFF
--- a/gocontext.go
+++ b/gocontext.go
@@ -41,11 +41,17 @@ func SpanFromContext(ctx context.Context) Span {
 //        ...
 //    }
 func StartSpanFromContext(ctx context.Context, operationName string, opts ...StartSpanOption) (Span, context.Context) {
-	return startSpanFromContextWithTracer(ctx, GlobalTracer(), operationName, opts...)
+	return StartSpanFromContextWithTracer(ctx, GlobalTracer(), operationName, opts...)
 }
 
-// startSpanFromContextWithTracer is factored out for testing purposes.
-func startSpanFromContextWithTracer(ctx context.Context, tracer Tracer, operationName string, opts ...StartSpanOption) (Span, context.Context) {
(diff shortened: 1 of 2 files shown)


yurishkuro:
@bhs @tedsuo any objections to exposing this method?

We could also simplify the name to just `opentracing.StartSpan(context.Context, Tracer)`

bhs:
@yurishkuro 

> We could also simplify the name to just opentracing.StartSpan(context.Context, Tracer)

Might be a little odd / asymmetric given the presence of `opentracing.StartSpanFromContext`... I don't have strong feelings, though.

devoxel:
The name does seem a bit long. I was wondering, is there any strong reason why Tracer isn't potentially a SpanOption? It could potentially be nice to have the same API across the board.

gocontext.go
```diff
@@ -41,11 +41,18 @@ func SpanFromContext(ctx context.Context) Span {
...
-func startSpanFromContextWithTracer(ctx context.Context, tracer Tracer, operationName string, opts ...StartSpanOption) (Span, context.Context) {
+// StartSpanFromContextWithTracer starts and returns a span with `operationName`
+// using  a span found within the context as a ChildOfRef. If that doesn't exist
+// it creates a root span. It also returns a context.Context object built
+// around the returned span.
+//
+// It's behavior is identical to StartSpanFromContext except that it takes an explicit
+// tracer as opposed to using the global tracer.
+func StartSpanFromContextWithTracer(ctx context.Context, tracer Tracer, operationName string, opts ...StartSpanOption) (Span, context.Context) {
+	var span Span
```
yurishkuro: build fails because this statement is not needed

pritishshah: @devoxel would you please fix it?

PR #155, opened 2017-07-27
add ipv4Tag type for `peer.ipv4` to support add ipv4 string to the `span`

### LINKED ISSUE #154: Shall we change peer.ipv4 to String tag in golang impl?
I noticed that `peer.ipv4` tag defined in `ext` package is uint32 type
```golang
// PeerHostIPv4 records IP v4 host address of the peer
PeerHostIPv4 = uint32TagName("peer.ipv4")
```
but , according to [Standard Span tags and log fields](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table), the `peer.ipv4` tag is described as 

-- | -- | --
:--: | :--: | :-- 
peer.ipv4 | string | Remote IPv4 address as a .-separated tuple. E.g., "127.0.0.1"
-- | -- | --

So, follow the conventions or keep the current code?

NeoCN:
Fixes #154

### DIFF
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -73,11 +73,8 @@ var (
 	// PeerHostname records the host name of the peer
 	PeerHostname = stringTagName("peer.hostname")
 
-	// PeerHostIPv4 records IP v4 host address of the peer as an uint32 value, keep this for backward compatibility
-	PeerHostIPv4 = uint32TagName("peer.ipv4")
-
-	// PeerHostIPv4String records IP v4 host address of the peer as a .-separated tuple. E.g., "127.0.0.1"
-	PeerHostIPv4String = stringTagName("peer.ipv4")
+	// PeerHostIPv4 records IP v4 host address of the peer
+	PeerHostIPv4 = ipv4Tag("peer.ipv4")
 
 	// PeerHostIPv6 records IP v6 host address of the peer
 	PeerHostIPv6 = stringTagName("peer.ipv6")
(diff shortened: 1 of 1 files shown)


ext/tags.go
```diff
@@ -73,9 +73,12 @@ var (
 	// PeerHostname records the host name of the peer
 	PeerHostname = stringTagName("peer.hostname")
 
-	// PeerHostIPv4 records IP v4 host address of the peer
+	// PeerHostIPv4 records IP v4 host address of the peer as an uint32 value, keep this for backward compatibility
 	PeerHostIPv4 = uint32TagName("peer.ipv4")
 
+	// PeerHostIPv4String records IP v4 host address of the peer as a .-separated tuple. E.g., "127.0.0.1"
+	PeerHostIPv4String = stringTagName("peer.ipv4")
```
yurishkuro: don't know if it's worth it, but it occurred to me that instead of a separate const we could define a new dedicated type for ipv4 that would have both Set(uint32) and SetString(string) methods. This would keep it both backwards compatible and keep the single public constant name.

NeoCN:
```golang
PeerHostIPv4 = ipv4TagName("peer.ipv4")

type ipv4TagName string

func (tag ipv4TagName) Set(span opentracing.Span, value uint32) {
	span.SetTag(string(tag), value)
}

func (tag ipv4TagName) SetString(span opentracing.Span, value string) {
	span.SetTag(string(tag), value)
}
```
Just like this?

yurishkuro: Yes

NeoCN:
@yurishkuro just want to know when will this pr and #152 be merged in master? We need them! Thanks

PR #158, opened 2017-09-21
Reduce unnecessary complexity

DiSiqueira:


DiSiqueira:
Yeap.
https://github.com/opentracing/opentracing-go/blob/master/gocontext_test.go#L32

PR #165, opened 2017-10-20
propagation: Add AMQP header based carrier support

keinohguchi:
Adding `AMQPHeadersCarrier` carrier to inject/extract tracing info into the AMQP header through `github.com/streadway/amqp` AMQP/RabbitMQ go package.

### DIFF
--- a/propagation.go
+++ b/propagation.go
@@ -174,3 +174,41 @@ func (c HTTPHeadersCarrier) ForeachKey(handler func(key, val string) error) erro
 	}
 	return nil
 }
+
+// AMQPHeadersCarrier satisfies both TextMapWriter and TextMapReader.
+//
+// Example usage for server side:
+//
+//     carrier := opentracing.AMQPHeadersCarrier(amqp.Table)
+//     clientContext, err := tracer.Extract(
+//         opentracing.TextMap,
+//         carrier)
+//
+// Example usage for client side:
+//
+//     carrier := opentracing.AMQPHeadersCarrier(amqp.Table)
+//     err := tracer.Inject(
+//         span.Context(),
+//         opentracing.TextMap,
+//         carrier)
+//
+type AMQPHeadersCarrier map[string]interface{}
+
+// ForeachKey conforms to the TextMapReader interface.
+func (c AMQPHeadersCarrier) ForeachKey(handler func(key, val string) error) error {
+	for k, val := range c {
+		v, ok := val.(string)
+		if !ok {
+			continue
+		}
+		if err := handler(k, v); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Set implements Set() of opentracing.TextMapWriter.
+func (c AMQPHeadersCarrier) Set(key, val string) {
+	c[key] = val
+}
--- a/propagation_test.go
+++ b/propagation_test.go
@@ -91,3 +91,46 @@ func TestHTTPHeaderExtract(t *testing.T) {
 		t.Errorf("Failed to read testprefix-fakeid correctly")
 	}
 }
+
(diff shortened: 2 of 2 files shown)


keinohguchi:
Yeah, github.com/streadway/amqp doesn't support 1.6/1.7 anymore.  I'll create PR against `github.com/streadway/amqp` unless they change their code.

Based on their reply, I'll close this PR, or do something about it.

yurishkuro:
Sorry, we cannot accept this change. The opentracing API library cannot have dependencies on some random vendor-specific libs. A better place for this is a separate repo under opentracing-contrib, where you may also provide the actual instrumentation for AMQP, not just the carrier implementation.

keinohguchi:
Thank you, @yurishkuro!  Yeah, that causes a circuler dependencies.  I've changed the carrier type to `map[string]interface{}` which is used by the `github.com/streadway/amqp`.  Is this works or still should be outside of this `opentracing-go` repo?

```
type AMQPHeadersCarrier map[string]interface{}
```

Thank you for your guidance, Yuri!

keinohguchi:
Here is a sample code to exercise `AMQPHeadersCarrier` just FYI:

```
import (
        "context"

        "github.com/opentracing/opentracing-go"
        "github.com/streadway/amqp"
)

// InjectContext injects the context info into the header.
func InjectContext(hdrs amqp.Table, span opentracing.Span) error {
        c := AMQPHeadersCarrier(hdrs)
        return span.Tracer().Inject(span.Context(), opentracing.TextMap, c)
}

// ExtractContext extracts the context info out of the header.
func ExtractContext(ctx context.Context, hdrs amqp.Table) context.Context {
        var sp opentracing.Span

        // Extract the tracer context from the request message.
        const opName = "ExtractContext"
        c := AMQPHeadersCarrier(hdrs)
        spCtx, err := opentracing.GlobalTracer().Extract(opentracing.TextMap, c)
        if err == nil {
                sp = opentracing.StartSpan(opName, opentracing.ChildOf(spCtx))
        } else {
                sp = opentracing.StartSpan(opName)
        }
        return opentracing.ContextWithSpan(ctx, sp)
}
```

yurishkuro:
Transport and/or framework specific implementations of the carriers to not belong here, they belong in the instrumentation code that actually uses them (and they can be private structs at that).

keinohguchi:
Thanks @yurishkuro for the suggestion!  I'll close this PR and created a simple repo, https://github.com/keinohguchi/go-amqp-opentracing, wishing to incorporate into the `opentracing-contrib`.
 
Based on https://github.com/opentracing-contrib/meta, seems sending email to hello@opentracing.com is the way to go but, if that's not the case, it would be super cool to point me the right direction.

Thank you in advance, Yuri!

yurishkuro:
that was a weird email address, I corrected the page to refer to https://groups.google.com/forum/#!forum/opentracing

yurishkuro:
@keinohguchi I created a repo and made you an admin https://github.com/opentracing-contrib/go-amqp/invitations

keinohguchi:
Thanks @yurishkuro!  I've the code to https://github.com/opentracing-contrib/go-amqp.

Can I enable the travis, or did I mis configured somewhere?

https://travis-ci.org/opentracing-contrib/go-amqp

yurishkuro:
As admin you can enable Travis for the repo. Just make sure to re sync the repos so that Travis knows about this one.

keinohguchi:
Ah, thanks!  Enabled and passed!

https://travis-ci.org/opentracing-contrib/go-amqp

Thank you for your support, Yuri!

yurishkuro:
No problem. I booked a couple small issues in the new repo.

PR #170, opened 2018-01-29
Fix race condition in MockSpan.

bradbl:
The comments in the MockSpan struct indicate that SpanContext is guarded
by the embedded mutex, but no mutex lock is done in the Context()
implementation. This allows for a race condition, for example between
calling Context() and calling SetBaggageItem().

### DIFF
--- a/mocktracer/mocktracer_test.go
+++ b/mocktracer/mocktracer_test.go
@@ -266,3 +266,14 @@ func TestMockTracer_Propagation(t *testing.T) {
 		assert.Equal(t, "y:z", extractedContext.(MockSpanContext).Baggage["x"])
 	}
 }
+
+func TestMockSpan_Races(t *testing.T) {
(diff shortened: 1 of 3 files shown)


dray92:
cc @yurishkuro

mocktracer/mocktracer_test.go
```diff
@@ -266,3 +266,14 @@ func TestMockTracer_Propagation(t *testing.T) {
...
+
+func TestMockSpan_Races(t *testing.T) {
+	tracer := New()
+	span := tracer.StartSpan("x")
+	go func() {
+		span.SetBaggageItem("test_key", "test_value")
+	}()
+	go func() {
+		span.Context()
+	}()
```
yurishkuro: a cleaner way would be to start a waitgroup of 2 and wait for each of the goroutines to finish before exiting the test.

Issue #153, opened 2017-07-11
Why not use "context"


cloudaice:
Why not use "context" instead of "golang.org/x/net/context" ?

yurishkuro:
We are still keeping 1.6 compatibility

tonyghita:
Now that 1.9 is released, are we still targeting go 1.6 compat?

yurishkuro:
My preference is to drop 1.6 support (#147)

tonyghita:
I've opened #176 to resolve this issue.

PR #176, opened 2018-02-11
Replace 'golang.org/x/net/context' with 'context'

tonyghita:
The `context` package was introduced in Go 1.7 as a replacement to the `golang.org/x/net/context` package.

Now that we're nearing the release of Go 1.10, we should replace the usage of `golang.org/x/net/context` with the standard library `context`.

This change requires that we drop support for Go 1.6 (#147) and resolves issue #153.

### DIFF
--- a/gocontext.go
+++ b/gocontext.go
@@ -1,6 +1,6 @@
 package opentracing
 
-import "golang.org/x/net/context"
+import "context"
 
 type contextKey struct{}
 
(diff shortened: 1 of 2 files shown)


tonyghita:
Not sure who to reach out to... can this be merged?

yurishkuro:
I just tested it with Jaeger's HotRod, no issues, so I think it's safe to merge.

PR #177, opened 2018-02-11
Use dep to manage dependencies

tonyghita:
[`dep`](https://golang.github.io/dep) is the "official experiment dependency management tool for Go", and is now safe for use in production.

This change adds `Gopkg.toml` and `Gopkg.lock` files generated by `dep init`, and adds `/vendor` to `.gitignore` (since libraries shouldn't check their `/vendor` directory in to VCS).


yurishkuro:
+1, but I would also suggest using `dep` to install dependencies in Travis builds to test that the Gopkg files are correct. Right now we install dependencies manually into global GOPATH.

Gopkg.toml
```diff
@@ -0,0 +1,15 @@
+# Refer to https://golang.github.io/dep/docs/Gopkg.toml.html
+# for detailed Gopkg.toml documentation.
+
+
+[[constraint]]
+  name = "github.com/stretchr/testify"
+  version = "1.2.1"
```
yurishkuro: why not `^1` ? Pinning to exact version might create conflicts when opentracing is used in other projects.

Also, I am not sure if `dep` supports this, but is there a way to specify the scope of the dependency, test vs. runtime?

tonyghita: > Why not `^1`? Pinning to exact version might create conflicts when opentracing is used in other projects.

The exact version was added as a result of running `dep init`. 

I don't think these files are obeyed when this library is used in other projects (although the authors of `dep` hinted that the `Glide.lock` may be used as a hint for conflict resolution in the future). 

As I understand it, the use of `dep` in libraries is more as a manifest for dependencies, and a convenient way of getting up to speed when contributing (i.e. `dep ensure` vs. `go get` * all dependencies).

That said, it's probably fine to specify `^1.2.1` so we catch all future releases in the major version. 

> Also, I am not sure if `dep` supports this, but is there a way to specify the scope of the dependency, test vs. runtime?

`dep` doesn't have the concept of test vs. runtime dependencies, although there is the concept of tooling dependencies (i.e. libraries you don't want in your `/vendor` directory) vs other dependencies (`[required]` vs `[constraint]`).

More on how `Gopkg.toml` works: https://golang.github.io/dep/docs/Gopkg.toml.html

tonyghita:
@yurishkuro I _think_ I addressed your comments. I'm not extremely familiar with the Travis CI configuration though.

.travis.yml
```diff
@@ -8,7 +8,7 @@ go:
 
 install:
   - go get -d -t github.com/opentracing/opentracing-go/...
-  - go get -u github.com/golang/lint/...
```
yurishkuro: don't remove golint, it's a tool, not a dependency

yurishkuro:
oh, it looks like `dep` doesn't work with Go 1.7
```
$ go get -u github.com/golang/dep/cmd/dep
# github.com/golang/dep/gps
../../golang/dep/gps/constraint.go:392: undefined: sort.SliceStable
../../golang/dep/gps/constraint.go:411: undefined: sort.SliceStable
../../golang/dep/gps/lock.go:55: undefined: sort.SliceIsSorted
```
1.8 build is failing on `go fmt ./...` because it descends into /vendor, we may need to do `go fmt $(go list ./...)`

1.9 fails because it can't find golint

Gopkg.toml
```diff
@@ -0,0 +1,16 @@
+# Refer to https://golang.github.io/dep/docs/Gopkg.toml.html
+# for detailed Gopkg.toml documentation.
+
+required = ["github.com/golang/lint"]
```
yurishkuro: I don't think we need this in deps, it's a tool, not a dependency

Makefile
```diff
@@ -22,6 +22,10 @@ cover-pkg:
 	go test -coverprofile cover.out $(PKG)
 	@grep -v mode: cover.out >> cover-all.out
 
+.PHONY: deps
+deps:
+	dep ensure --vendor-only
```
yurishkuro: To work around 1.7 issue we may need to create a travis matrix (like this https://github.com/jaegertracing/jaeger-client-go/blob/master/.travis.yml#L8) and set an env var for 1.7 like MANUAL_DEPS=true so that we only run `dep ensure` when it's not set

.travis.yml
tonyghita: Hmmm are we sure? From reading the [dep documentation](https://golang.github.io/dep/docs/Gopkg.toml.html#required) it seemed like this is what `required` is for:

> Use this for: linters, generators, and other development tools that
> - Are needed by your project
> - Aren't imported by your project, directly or transitively
> - You don't want to put them in your GOPATH, and/or you want to lock the version

yurishkuro:
I am going to close this, since dep is no longer the recommended solution.

Issue #175, opened 2018-02-09
Support for span.LogFields with Zap Fields


bvandewalle:
Hi,

After using `opentracing` together with `zap` (a logger service also in go) , I find myself continuously logging fields both to my Span (which uses `LogFields(fields ...log.Field)` as signature ) and to my Zap logger ( which uses `Debug(log string, fields ...zapcore.Field)`  as signature).

Essentialy allocating twice the amount of `fields` required.

Would it be envisioned that we could also pass Fields of type `...zapcore.Field`  (which got essentialy the same underlying data ) to opentracing ? 

Of course the ask could be reversed and the burden could be on `zap` to get `opentracing` logs implemented as input.

bvandewalle:
As for the use-case and the reason why I need to log and trace those fields in both location, let's just say that in some cases I need to have the possibility to fall backs to logs if my Opentracing backend is not available (in cases where I deploy in a non-controlled environment).

bvandewalle:
This was added in opentracing-contrib. See https://github.com/opentracing-contrib/go-zap

Closing this one

PR #178, opened 2018-02-13
Add support for creating Opentracing fields from Zap fields

bvandewalle:
PR to partially implement #175  

This PR adds two funcs:

`func FieldFromZap(zapField zapcore.Field) Field`

and

`func FieldsFromZap(zapFields ...zapcore.Field) []Field`

That can be used by a user that wishes to use Zap fields as input for Opentracing spans.

Essentially, this means that on a project that uses Zap, the user can now use the same set of `...zapcore.Field` to both log to zap and send to Opentracing.

Since `zap` fields are almost the same as `opentracing` fields, the conversion is pretty much straightforward.

### DIFF
--- a/log/field_test.go
+++ b/log/field_test.go
@@ -102,10 +102,6 @@ func TestFieldsFromZap(t *testing.T) {
 			zap.Bool("namespace", false),
 			Bool("namespace", false),
 		},
-		{
-			zap.Bool("namespace", true),
-			Bool("namespace", true),
-		},
 	}
 
 	for i, data := range testCases {
--- a/log/field_test.go
+++ b/log/field_test.go
@@ -3,6 +3,9 @@ package log
 import (
 	"fmt"
 	"testing"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 func TestFieldString(t *testing.T) {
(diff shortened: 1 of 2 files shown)


yurishkuro:
We don't make a practice of adding dependencies on specific technologies to the API library . Can this extension be done as an opentracing-contrib library?  It doesn't look like it's using any package-private symbols.

bvandewalle:
Thanks for the feedback. Indeed, this is probably better fitted as an external library on opentracing-contrib indeed. I checked quickly over there, is the idea to create a new repo to host this ? Or can you point me to an existing repo on which I could do a pull request?

yurishkuro:
Yes, this would be a new repo, like `go-zap`. I can create it for you if you want.

bvandewalle:
sounds good. I can move the PR over there. Thanks

bvandewalle:
Let me know when you created that repo. Didn't see it on the opentracing-contrib org.

bvandewalle:
@yurishkuro Any news ? :-) Thanks!

yurishkuro:
Sorry, Github didn't email your comments.

Created https://github.com/opentracing-contrib/go-zap and made you an admin.

bvandewalle:
Thanks @yurishkuro

PR #179, opened 2018-02-19
OpenTracing Semantic Conventions uses error.object

### LINKED ISSUE #164: error.object or just error?
https://github.com/opentracing/specification/blob/master/semantic_conventions.md says to use `"error.object"` as the log key for an error object, but https://github.com/opentracing/opentracing-go/blob/1361b9cd60be79c4c3a7fa9841b3c132e40066a7/log/field.go#L126-L132 uses `"error"`. Which is right?

rahmansyed:
The OpenTracing Specification suggests "error.object" as the log
key for an error object.

Fixes #164

### DIFF
--- a/log/field.go
+++ b/log/field.go
@@ -122,10 +122,10 @@ func Float64(key string, val float64) Field {
 	}
 }
 
-// Error adds an error with the key "error" to a Span.LogFields() record
+// Error adds an error with the key "error.object" to a Span.LogFields() record
 func Error(err error) Field {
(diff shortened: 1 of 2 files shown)


miry:
It seems the valid implementation of semantic. Why it is not merged?

andrewslotin:
Hi there! This turned out to be a breaking change to the libraries that depend on `opentracing-go` and rely on the name of the tag that carries an error.

PR #181, opened 2018-03-27
update license to Apache 2.0

eirinikos:
This commit replaces the MIT license with the Apache 2.0 license.

(All CNCF projects should use the Apache License 2.0 - please see opentracing/specification#61 for context.)

Note:
- none of the existing files has a license header, so I haven't added any headers in this commit.
- I have preserved the copyright as `Copyright 2016 The OpenTracing Authors`. Should the year be listed as 2018 instead?

Please air any objections by April 3rd, thanks!

CCing @dankohn

CCing previous contributors: @bhs @tschottdorf @yurishkuro @bg451 @jmacd @slimsag @basvanbeek @cce @stamm @RaduBerinde @gwik @sebastianvera @IncSW @bradbl @kyleconroy @bmizerany @RichardScothern @tamird @PikBot @ror6ax @dvrkps @NeoCN @savaki @rohanpai @tedsuo @billf @voutasaurus @kriskowal

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,201 @@
-The MIT License (MIT)
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Copyright (c) 2016 The OpenTracing Authors
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
+   1. Definitions.
 
-The above copyright notice and this permission notice shall be included in all
-copies or substantial portions of the Software.
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
 
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


bhs:
@eirinikos looks good to me, though we should def wait to see if others object!

Re the copyright year, I think the convention would be `Copyright 2016-2018 The OpenTracing Authors`, but I am definitely not an expert (that's just what I've seen).

dankohn:
We would recommend leaving off the year -- `Copyright The OpenTracing Authors` -- though it's not essential.

slimsag:
I am fine with this change, but I am a bit worried with how it will be carried out. As far as I know (IANAL), you cannot simply "ask for objections" before relicensing a project even if the licenses are similar in nature.

Any prior contributor to this project released their contribution under the MIT license terms, one of which is:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

So it stands to reason that we would need the approval of every prior contributor to make this change, or we must continue to distribute the MIT license "in all copies or substantial portions of the Software."

Likewise, we can't simply say all prior contributions are under the Apache 2.0 license unless we establish those contributor's intent / approval to relicense their contributions under the new license.

So, I'm totally fine with this change, and I'm a big fan of the Apache 2.0 license -- but doing this without some form of explicit permission from each contributor (PR approval here, etc) is a legal minefield IMHO.

bhs:
@slimsag:

> but doing this without some form of explicit permission from each contributor (PR approval here, etc) is a legal minefield IMHO.

IAAlsoNAL, but my understanding from conversations with a variety of people who know about these matters suggests that it is, in fact, ok... you can read a little more at https://github.com/opentracing/specification/issues/61.

eirinikos:
@bhs 👋 

For whatever reason, Licensee is still not identifying this repo's license type as Apache 2.0. (Licensee being the tool that GitHub uses to tag repositories with their license types.)

Could we hold off on merging this PR until I push a new commit to fix the problem? I hope to have a solution in the next day or so.

(Right now, this problem affects OT-Go, OT-Objc, OT-PHP, and OT-Ruby.)

yurishkuro:
> For whatever reason, Licensee is still not identifying this repo's license type as Apache 2.0

it wouldn't until the PR is merge.

eirinikos:
> it wouldn't until the PR is merge.

I'm testing changes locally.

eirinikos:
For some reason, in some situations, using a `# License` header in a repo's README will cause Licensee to scan the README as a license file.

Replacing `# License` with `## Licensing` allows us to bypass this problem.

If there are no objections to switching to Apache 2.0, then this PR is ready for a merge tomorrow!

Issue #159, opened 2017-09-25
Should use Set() instead of Add() in  HTTPHeadersCarrier


vetinari:
When re-using an incoming http.Request, Add() will create a double header (i.e. result is a comma separated value), see also https://github.com/instana/golang-sensor/pull/40

jeremyxu2010:
+1

tobias-bahls:
Such headers might cause OpenTracing implementations on the JVM throw an unchecked `IllegalArgumentException`, which is not handled properly in e.g. https://github.com/opentracing-contrib/java-web-servlet-filter.

PR #191, opened 2018-08-22
Should use Set() instead of Add() in HTTPHeadersCarrier

jeremyxu2010:
Should use Set() instead of Add() in HTTPHeadersCarrier. see https://github.com/opentracing/opentracing-go/issues/159


yurishkuro:
I agree that this is probably the right change, but it might break users who already depend on the existing behavior (e.g. by writing the same header multiple times to the carrier).

@tedsuo

Issue #194, opened 2018-09-23
Using mocktracer with log.Lazy to log multiple fields


alexsn:
## Use Case
Trying to log multiple fields from a single log.Lazy field:
```
span.LogFields(log.Lazy(func(enc log.Encoder) {
    enc.EmitString("level", "info")
    enc.EmitString("message", "hello world")
}))
```

## Problem
MockKeyValue::EmitLazyLogger holds only the latest emitted field meaning you can't verify all your emitted fields in test.
```
func (m *MockKeyValue) EmitLazyLogger(value log.LazyLogger) {
	var meta MockKeyValue
	value(&meta)
	m.Key = meta.Key
	m.ValueKind = meta.ValueKind
	m.ValueString = meta.ValueString
}
```

## Questions to address (if any)
Am I not suppose to log multiple fields from a single log.Lazy or is this a bug / limitation?
Note that I've tested the above code with jaeger and all the emitted fields are visible.

PR #195, opened 2018-09-23
mocktracer: move log.Encoder interface implementation to MockLogRecord

alexsn:
log.Lazy field can emit multiple encoded fields however since MockKeyValue was
the one implementing log.Encoder a one to one relation was forced between logged
and encoded fields.

see: https://github.com/opentracing/opentracing-go/issues/194#issue-362915564


alexsn:
Project looks dead, abandoning.

PR #202, opened 2018-11-29
Fix lint and coverage

yurishkuro:
## Problem
* The old `go get -u github.com/golang/lint/...` no longer works
* The new `go get -u golang.org/x/lint/golint/...` does not work in older Go versions (it uses type alias)
* Coverage was never uploaded

## Fixes
* Use Travis build matrix with env vars to control LINT and COVERAGE
* Only run lint and coverage on tip, not on older versions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,1 @@
-# IntelliJ project files
-.idea/
-opentracing-go.iml
-opentracing-go.ipr
-opentracing-go.iws
-
-# Test results
-*.cov
-*.html
-test.log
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,13 +1 @@
-# IntelliJ project files
-.idea/
-opentracing-go.iml
-opentracing-go.ipr
-opentracing-go.iws
```
yurishkuro: these should go to user's global gitignore

PR #206, opened 2019-02-15
Fix redundant import alias

huynq0911:
Because these package names are short and they are unique, it can not conflict. So the import aliases that can be omitted and we should remove them.


yurishkuro:
Thx

PR #208, opened 2019-02-18
Simplify RPCServerOption

funny-falcon:


funny-falcon:
Ops, misread original code

PR #210, opened 2019-03-01
Fix function comments based on best practices from Effective Go

CodeLingoTeam:
Hi, we updated some exported function comments based on best practices from [Effective Go](https://golang.org/doc/effective_go.html#commentary). It’s admittedly a relatively minor fix up. Does this help you?

### DIFF
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -192,7 +192,7 @@ func (tag uint16TagName) Set(span opentracing.Span, value uint16) {
 
 type boolTagName string
 
-// Add adds a bool tag to the `span`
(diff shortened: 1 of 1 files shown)


ext/tags.go
```diff
@@ -192,7 +192,7 @@ func (tag uint16TagName) Set(span opentracing.Span, value uint16) {
 
 type boolTagName string
 
-// Add adds a bool tag to the `span`
+// Set adds adds a bool tag to the `span`
```
yurishkuro: adds twice

yurishkuro:
no longer needed

PR #215, opened 2019-06-17
Add Go module support

Zaba505:


cristaloleg:
Bump. Any plans on this?
Fails is only on tip, regarding getting golint (which is kinda strange)

Zaba505:
@cristaloleg Fixed it! Honestly, not sure what the problem was? It may have been me forgetting to run `go mod tidy` before committing `go.mod`.

EronWright:
Bump.

PR #226, opened 2019-11-27
ext: Add LogFields helpers for keys from specification

dmonakhov:
Opentracing spec specify prefered names for log messages [1]
This patch adds declaration for fields which are meaningful for golan

I've added only minimal subset of features as a first step,
For example this patch has no stacktrace logging for errors because  stacktrace generation is implementation depended, so caller should do it explicitly.
Footnotes:
[1] https://github.com/opentracing/specification/blob/master/semantic_conventions.md

### DIFF
--- a/ext/field.go
+++ b/ext/field.go
@@ -1,0 +1,17 @@
+package ext
+
+import (
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/log"
+)
+
+// LogError sets the error=true tag on the Span and logs err as an "error" event.
+func LogError(span opentracing.Span, err error, fields ...log.Field) {
+	Error.Set(span, true)
+	ef := []log.Field{
+		log.Event("error"),
+		log.Error(err),
+	}
+	ef = append(ef, fields...)
+	span.LogFields(ef...)
+}
--- a/log/field.go
+++ b/log/field.go
@@ -143,6 +143,16 @@ func Object(key string, obj interface{}) Field {
 	}
 }
 
+// Event creates a string-valued Field for span logs with key="event" and value=val.
+func Event(val string) Field {
+	return String("event", val)
+}
+
+// Message creates a string-valued Field for span logs with key="message" and value=val.
+func Message(val string) Field {
+	return String("message", val)
+}
+
 // LazyLogger allows for user-defined, late-bound logging of arbitrary data
 type LazyLogger func(fv Encoder)
 
--- a/ext/field_test.go
+++ b/ext/field_test.go
@@ -1,0 +1,50 @@
+package ext_test
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/opentracing/opentracing-go/ext"
+	"github.com/opentracing/opentracing-go/log"
+	"github.com/opentracing/opentracing-go/mocktracer"
+)
+
+func TestLogError(t *testing.T) {
+	tracer := mocktracer.New()
+	span := tracer.StartSpan("my-trace")
+	ext.Component.Set(span, "my-awesome-library")
+	ext.SamplingPriority.Set(span, 1)
+	err := fmt.Errorf("my error")
+	ext.LogError(span, err, log.Message("my optional msg text"))
+
+	span.Finish()
+
+	rawSpan := tracer.FinishedSpans()[0]
+	assert.Equal(t, map[string]interface{}{
+		"component": "my-awesome-library",
+		"error":     true,
+	}, rawSpan.Tags())
+
+	assert.Equal(t, len(rawSpan.Logs()), 1)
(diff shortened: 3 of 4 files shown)


ext/fields.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/opentracing/opentracing-go/log"
+)
+
+// These constants define common logFields names recommended for better portability across
+// tracing systems and languages/platforms
+// More info: https://github.com/opentracing/specification/blob/master/semantic_conventions.md
+var (
+	LogEvent   = stringLogName("event")
+	LogMessage = stringLogName("message")
+	LogStack   = stringLogName("stack")
```
yurishkuro: I think these would make more sense in the `log` pkg

ext/fields.go
```diff
@@ -0,0 +1,31 @@
...
+	Error.Set(span, true)
+	span.LogFields(ErrorEvent, log.String(string(LogMessage), err.Error()))
+}
+
+// Add log string to to the `span`
+func (logName stringLogName) Log(span opentracing.Span, value string) {
+	span.LogFields(log.String(string(logName), value))
+}
+
+type stringLogName string
```
yurishkuro: declaration should precede the functions

ext/fields.go
yurishkuro: also, only the first of these is self-contained Field, the others don't make sense as fields, they are just keys. They would make more sense as functions taking the value string and returning a Field.

dmonakhov: > I think these would make more sense in the `log` pkg
Sound reasonable, but LogError() depends on opentrace.Span so this result in cyclic dependency between opentrace and log packages.

ext/fields.go
dmonakhov: Ok, fixed in new version

log/field.go
```diff
@@ -143,6 +143,21 @@ func Object(key string, obj interface{}) Field {
...
+
+// Message adds a string-valued field with name "event" to a Span.LogFields() record
+func Message(val string) Field {
+	return String("message", val)
+}
+
+// Stack adds a string-valued field with name "stack" to a Span.LogFields() record
+func Stack(val string) Field {
+	return String("stack", val)
+}
```
yurishkuro: why is "stack" a string? Is the intention to represent a stack trace?

dmonakhov: It was declared  that way in  [semantic conventions doc](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#log-fields-table) 
`| "stack" | string | A stack trace in platform-conventional format; may or may not pertain to an error. E.g., "File \"example.py\", line 7, in \<module\>\ncaller()\nFile \"example.py\", line 5, in caller\ncallee()\nFile \"example.py\", line 2, in callee\nraise Exception(\"Yikes\")\n"`

I tend to agree that this may be not suitable for some implementations, so I can drop that function.

yurishkuro: I think the convention is a bit misleading here. Once the tag is set and accepted by the span (e.g. when the trace is sampled), then yes, it has to be a string, as there's no other data structure available (NB: OpenTelemetry will have a proper structure for stack frames). But that doesn't mean that the tag must be set as a string, because doing so would be quite inefficient if the trace is not sampled. I always prefer the API to take native objects (e.g. https://golang.org/pkg/runtime/debug/#Stack, or a Throwable in Java) and the tracer implementation can decide to convert them to strings.

So yes, I recommend removing this, as it's not required by the rest of the PR.

dmonakhov: Agree. Fixed in new version.

log/field.go
```diff
@@ -143,6 +143,16 @@ func Object(key string, obj interface{}) Field {
 	}
 }
 
+// Event adds a string-valued field with name "value" to a Span.LogFields() record
```
yurishkuro: description is not accurate. Suggest:

`Event creates a string-valued Field for span logs with key="event" and value=val.`

log/field.go
```diff
@@ -143,6 +143,16 @@ func Object(key string, obj interface{}) Field {
...
+func Event(val string) Field {
+	return String("event", val)
+}
+
+// Message adds a string-valued field with name "event" to a Span.LogFields() record
```
yurishkuro: `Message creates a string-valued Field for span logs with key="message" and value=val.`

ext/field.go
```diff
@@ -0,0 +1,17 @@
+package ext
+
+import (
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/log"
+)
+
+// LogError add error event record for the Span
```
yurishkuro: `LogError sets the error=true tag on the Span and logs err as an "error" event.`

log/field.go
dmonakhov: Ok, fixed in fresh version. Thank you for your patience.

yurishkuro:
Thanks for the PR.

PR #230, opened 2020-01-23
Set nil pointer to string nil when converting key values.

cyriltovena:
This causes a panic that will be recovered by the fmt package.
Althought on darwin this makes any projects using open tracing impossible to debug with dlv

see https://github.com/cortexproject/cortex/issues/1712

Signed-off-by: Cyril Tovena <cyril.tovena@gmail.com>

### DIFF
--- a/log/util_test.go
+++ b/log/util_test.go
@@ -1,0 +1,86 @@
+package log
+
+import (
+	"errors"
+	"io"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var nilInterface io.Reader
+
+func TestInterleavedKVToFields(t *testing.T) {
+
+	tests := []struct {
+		name      string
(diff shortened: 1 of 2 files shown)


yurishkuro:
please add a test

cyriltovena:
This function wasn't covered by any tests, unless I'm mistaken, anyways I add a test to cover it entirely.

log/util_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"errors"
+	"io"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var nilInterface io.Reader
```
yurishkuro: this can be scoped inside the test function

cyriltovena: Sure ! Will do

PR #236, opened 2020-11-06
Adding ppc64le architecture support on travis-ci

kishorkunal-raj:
Hi,
I had added ppc64le(Linux on Power) architecture support on Travis-CI in the PR and looks like its been successfully added. I believe it is ready for the final review and merge. The Travis-CI build logs can be verified from the link below.
https://travis-ci.com/github/kishorkunal-raj/opentracing-go/builds/198782358

Reason behind running tests on ppc64le: This package is included in the ppc64le versions of RHEL and Ubuntu - this allows the top of tree to be tested continuously as it is for Intel, making it easier to catch any possible regressions on ppc64le before the distros begin their clones and builds. This reduces the work in integrating this package into future versions of RHEL/Ubuntu.

Please have a look.

Regards,
Kishor Kunal Raj

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ matrix:
     env:
     - LINT=true
     - COVERAGE=true
+  # Adding ppc64le jobs
+  - go: "tip"
+    arch: ppc64le
 
 install:
   - if [ "$LINT" == true ]; then go get -u golang.org/x/lint/golint/... ; else echo 'skipping lint'; fi


.travis.yml
```diff
@@ -8,6 +8,16 @@ matrix:
...
+  # Adding ppc64le jobs
+  - go: "1.13.x"
+    arch: ppc64le
+  - go: "1.14.x"
+    arch: ppc64le
+  - go: "tip"
+    arch: ppc64le
+    env:
+    - LINT=true
+    - COVERAGE=true
```
yurishkuro, line 20: 1. please remove LINT/COVERAGE, they should be in one job only
1. is it necessary to test ppc64le on anything other than tip?

kishorkunal-raj: Hi @yurishkuro 
I had removed coverage from the file. Also It will be a good idea to test ppc64le on go all go versions along with tip.

kishorkunal-raj (commented):
Based on the suggestion, made the required changes. Please have a look

.travis.yml
yurishkuro: `COVERAGE=true` must bet set only on the main Go Tip step.

> Also It will be a good idea to test ppc64le on go all go versions along with tip.

Nobody tests with "all go versions", standard practice is only the latest 2 releases. And we already do that for the default architecture, I don't see what it adds to test on another architecture. ppc64le only needs tip.

kishorkunal-raj: Hi @yurishkuro ,
As suggested, I had made the changes to the file and it appears fine to me. Will you please have a look and provide suggestion. Also the changes are not adding anything extra apart from testing the existing tests on ppc64le platform.

.travis.yml
```diff
@@ -8,6 +8,11 @@ matrix:
     env:
     - LINT=true
     - COVERAGE=true
+  # Adding ppc64le jobs
+  - go: "tip"
+    arch: ppc64le
+    env:
+    - COVERAGE=true
```
yurishkuro, line 15: We cannot use COVERAGE in this step, because it will double-report it to codecov.

You can introduce another var TEST=true and add this under `script:`
```
if [ "$TEST" == true ]; then make test ; fi
```

yurishkuro: Actually, don't need new variable at all, tests are run unconditionally already. So just remove the env var.

kishorkunal-raj: Thank you for suggesting the changes. I had removed the COVERAGE env while running go tip  on ppc64le

yurishkuro (approved):

PR #239, opened 2021-02-05
Add started spans to mock tracer

jordanbangia:
My team has started using opentracing a bunch, but one problem we often run into is someone will forget to finish their started spans.  We noticed that in some cases that would cause the entire span to be dropped.  

I'm hoping to catch that in testing by providing a way to assert that every started span has a finished span, and to provide a nice error to point to which span was not finished.

### DIFF
--- a/mocktracer/mocktracer.go
+++ b/mocktracer/mocktracer.go
@@ -1,6 +1,7 @@
 package mocktracer
 
 import (
+	"fmt"
 	"sync"
 
 	"github.com/opentracing/opentracing-go"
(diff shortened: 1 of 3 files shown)


mocktracer/mocktracer.go
```diff
@@ -48,12 +60,37 @@ func (t *MockTracer) FinishedSpans() []*MockSpan {
 	return spans
 }
 
+func (t *MockTracer) AssertStartedSpansAreFinished() error {
+	startedSpans := t.StartedSpans()
```
yurishkuro, line 64: instead of doing comparison, why not just remove the span from `started` when it is finished?

jordanbangia: That works too, will update

mocktracer/mocktracer.go
```diff
@@ -98,8 +115,21 @@ func (t *MockTracer) Extract(format interface{}, carrier interface{}) (opentraci
...
+	t.startedSpans = append(t.startedSpans, span)
+}
+
+func (t *MockTracer) recordFinishedSpan(span *MockSpan) {
 	t.Lock()
 	defer t.Unlock()
 	t.finishedSpans = append(t.finishedSpans, span)
+
+	for i := range t.startedSpans {
+		if t.startedSpans[i].SpanContext.SpanID == span.SpanContext.SpanID {
```
yurishkuro, line 130: I would compare trace ID as well, otherwise you might get false positives here

mocktracer/mocktracer.go
```diff
@@ -98,8 +115,21 @@ func (t *MockTracer) Extract(format interface{}, carrier interface{}) (opentraci
...
+			t.startedSpans = append(t.startedSpans[:i], t.startedSpans[i+1:]...)
```
yurishkuro, line 131: `t.startedSpans[i+1:]` - I don't remember exactly if this could cause an overflow in Go, but you're potentially using index out of bounds

jordanbangia: I think its safe, tried it out in a playground here https://play.golang.org/p/TTKWlI51fTF

yurishkuro: interesting, that same example fails with `[7:]` but ok with `[6:]` even though both are out of bounds, strictly speaking.

yurishkuro (approved):

jordanbangia:
Thanks for the fast reviews @yurishkuro 😄

PR #240, opened 2021-04-25
Added http response size as external tag

ekrucio:
Http response size is a useful metric and I think it's good to have it defined as external tag to keep it consistent.

Related to: https://github.com/opentracing-contrib/go-stdlib/pull/60

### DIFF
--- a/ext/tags.go
+++ b/ext/tags.go
@@ -97,6 +97,10 @@ var (
 	// HTTP response.
 	HTTPStatusCode = Uint16TagName("http.status_code")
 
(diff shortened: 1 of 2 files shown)


yurishkuro:
Tags are first defined in the spec. But the spec is frozen due to migration to opentelemetry.

ekrucio:
> Tags are first defined in the spec. But the spec is frozen due to migration to opentelemetry.

My bad, sorry about that. Should I close the issuse for now since the specifications are frozen?

PR #242, opened 2021-07-12
Set up Dependabot

mmorel-35:


yurishkuro (approved):
Thanks!  Would you be interested in adding a github action for the build? We should've migrated from Travis long time ago.

mmorel-35:
I can make a PR for this, if you want me to?

yurishkuro:
it would be good, because travis runs take forever (if they ever finish), so dependabot just created a PR that we can't merge because there's no green build.

PR #244, opened 2021-07-12
Set up Go workflow and update testify

mmorel-35:


.github/workflows/go.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    - name: Build
+      run: go build -v ./...
+
+    - name: Test with coverage 
+      run: go test -race -v -count=1 -coverprofile=coverage.out ./...
+    - name: Upload Test Coverage
+      uses: codecov/codecov-action@v1
+      with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
mmorel-35, line 47: You'll need to provide the codecov_token in the project's secret

.github/dependabot.yml
```diff
@@ -4,3 +4,8 @@ updates:
   directory: /
   schedule:
     interval: daily
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: daily
+
```
mmorel-35, line 11:
```suggestion
```

.github/workflows/go.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        continue-on-error: true
```
mmorel-35, line 13: This can be removed after linting warning has been fixed

.github/workflows/go.yml
yurishkuro: added

yurishkuro (approved):

yurishkuro:
Thanks, @mmorel-35 , the workflow ran successfully on master. Is there a way to suppress the deprecation warnings so that the lint can be set to fail the build?

mmorel-35:
The warning can be ignored, but can't they be fixed instead ?

yurishkuro:
The usage of the deprecated functions is in the tests and api_checker, they cannot be removed. #245

mmorel-35:
Alright :)

PR #245, opened 2021-07-12
Fix lint warnings and re-enable failing CI on lint errors

yurishkuro:
- disable deprecated function warnings since they are used in tests & api_checker and must remain
- re-enable failing CI on lint errors


mmorel-35:
You can also extend the list of linters you want to use with .golangci.yml
They are listed here : https://golangci-lint.run/usage/linters/

